PR #19, opened 2015-07-14
the vmbackup extension fix and the support for the set public key without convert.

andyliuliming:
the vmbackup extension fix and the support for the set public key without convert.


andyliuliming:
will seperate the two items.

PR #20, opened 2015-07-14
1. fix in vmbackup. 2. add the plain text setting for the ssh authorized publickey.

andyliuliming:
1. fix the vm hangs when the loop backup device exists.
2. add the plain text setting for the ssh authorized publickey.
3. fix one issue in the  case that the format not begins with BeginCertificateTag


yuezh:
Looks good to me.

PR #23, opened 2015-07-31
Adding Check and Repair disk to VMAccessExtension

Snesha:
Yue, could you please review? This is new functionality that has been requested.

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -33,18 +33,19 @@ import traceback
 from Utils.WAAgentUtil import waagent
 import Utils.HandlerUtil as Util
 
-#Define global variables
+# Define global variables
 ExtensionShortName = 'VMAccess'
 BeginCertificateTag = '-----BEGIN CERTIFICATE-----'
 EndCertificateTag = '-----END CERTIFICATE-----'
 OutputSplitter = ';'
 
+
 def main():
-    waagent.LoggerInit('/var/log/waagent.log','/dev/stdout')
-    waagent.Log("%s started to handle." %(ExtensionShortName)) 
+    waagent.LoggerInit('/var/log/waagent.log', '/dev/stdout')
+    waagent.Log("%s started to handle." % (ExtensionShortName))
     waagent.MyDistro = waagent.GetMyDistro()
 
-    for a in sys.argv[1:]:        
+    for a in sys.argv[1:]:
         if re.match("^([-/]*)(disable)", a):
             disable()
         elif re.match("^([-/]*)(uninstall)", a):
@@ -56,21 +57,25 @@ def main():
         elif re.match("^([-/]*)(update)", a):
             update()
 
+
 def install():
     hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName)
     hutil.do_parse_context('Uninstall')
-    hutil.do_exit(0,'Install','Installed','0', 'Install Succeeded')
+    hutil.do_exit(0, 'Install', 'Installed', '0', 'Install Succeeded')
+
 
 def enable():
     hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName)
     hutil.do_parse_context('Enable')
     try:
-        public_settings = hutil._context._config['runtimeSettings'][0]['handlerSettings'].get('publicSettings')
-        protect_settings = hutil._context._config['runtimeSettings'][0]['handlerSettings'].get('protectedSettings')
+        public_settings = hutil._context._config['runtimeSettings'][
+            0]['handlerSettings'].get('publicSettings')
+        protect_settings = hutil._context._config['runtimeSettings'][
+            0]['handlerSettings'].get('protectedSettings')
         reset_ssh = protect_settings.get('reset_ssh')
         check_disk = public_settings.get('check_disk')
         repair_disk = public_settings.get('repair_disk')
-	# check port each time the VM boots up
+    # check port each time the VM boots up
         if reset_ssh:
             _open_ssh_port()
             hutil.log("Succeeded in check and open ssh port.")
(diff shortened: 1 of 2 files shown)


bingosummer:
Hi Snesha, you can follow https://github.com/Azure/azure-linux-extensions/blob/master/README.md to test your code and make sure it works in your test environment. Any questions, feel free to leave a comment.

VMAccess/vmaccess.py
```diff
@@ -65,9 +65,12 @@ def enable():
     hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName)
     hutil.do_parse_context('Enable')
     try:
+	public_settings = hutil._context._config['runtimeSettings'][0]['handlerSettings'].get('publicSettings')
         protect_settings = hutil._context._config['runtimeSettings'][0]['handlerSettings'].get('protectedSettings')
         reset_ssh = protect_settings.get('reset_ssh')
-        # check port each time the VM boots up
+        check_disk = public_settings.get('check_disk')
+	repair_disk = public_settings.get('repair_disk')
```
yuezh: Hi, @Snehsa

Would you please expand "tab" to 4 spaces at the beginning of the line? Using "tab" could introduce issue since different text editors would interpreter it differently.

VMAccess/vmaccess.py
```diff
@@ -65,9 +65,12 @@ def enable():
     hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName)
     hutil.do_parse_context('Enable')
     try:
+	public_settings = hutil._context._config['runtimeSettings'][0]['handlerSettings'].get('publicSettings')
         protect_settings = hutil._context._config['runtimeSettings'][0]['handlerSettings'].get('protectedSettings')
```
paulmey: it looks like an extra space snuck in here. I would be nice to clean that up to avoid unintentional consequences in later changes.

VMAccess/vmaccess.py
```diff
@@ -79,6 +82,14 @@ def enable():
...
             _remove_user_account(remove_user, hutil)
         _set_user_account_pub_key(protect_settings, hutil)
+	if check_disk:
+            _fsck_check()
+            hutil.log("Successfully checked disk")
+            hutil.exit_if_enabled()
+        if repair_disk:
+            _fsck_repair()
+            hutil.log("Repaired and remounted disk")
+            hutil.exit_if_enabled()
```
paulmey: I think this line and the line 88 above are noops, since hutil.exit_if_enabled() would have exited on line 77 if this was not a new config according to the sequence number (https://github.com/Azure/azure-linux-extensions/blob/master/Utils/HandlerUtil.py#L252-L260).
So if you want to exit, then use hutil.do_exit...

VMAccess/vmaccess.py
```diff
@@ -321,6 +332,41 @@ def _insert_rule_if_not_exists(rule_string):
...
+	cmd_result = waagent.Run("umount -dflr")
+	if cmd_result!=0:
+	#Fail fast
+		raise Exception("Failed to unmount disks")
+	else:
+	  # run repair
+	  retcode = waagent.Run("fsck -AR -y")
+	    if retcode >0:
+		raise Exception("Failed to repair disk")
+		else:
```
paulmey: it looks like this else should be at the same indent as the if on line 357?
Otherwise, this else will only be hit if retcode>0 (which doesn't make sense and will probably result in a crash) and only the next line will be the 'else block' (since it's not indented more)

VMAccess/vmaccess.py
```diff
@@ -321,6 +332,41 @@ def _insert_rule_if_not_exists(rule_string):
...
+		#remount all post repair and return
+		retcode = waagent.Run("mount -a")
+		if retcode ==0:
+		return retcode
+		else:
+		raise Exception("Failed to mount disks")
+	except Exception, e:
+		hutil.error("%s" %(str(e), traceback.format_exc()))
+		hutil.do_exit(1, 'Repair','error','0', 'Repair failed.')
+
```
paulmey: it's tabs instead of spaces till here

VMAccess/vmaccess.py
Snesha: Fixed the file

VMAccess/vmaccess.py
Snesha: Fixed

VMAccess/vmaccess.py
Snesha: Fixed

VMAccess/vmaccess.py
Snesha: Fixed

VMAccess/vmaccess.py
```diff
@@ -56,49 +57,70 @@ def main():
...
+        check_disk = public_settings.get('check_disk')
+        repair_disk = public_settings.get('repair_disk')
+    # check port each time the VM boots up
         if reset_ssh:
             _open_ssh_port()
             hutil.log("Succeeded in check and open ssh port.")
-        hutil.exit_if_enabled()
-        remove_user = protect_settings.get('remove_user')
+            hutil.exit_if_enabled()
+            remove_user = protect_settings.get('remove_user')
```
paulmey: this fix breaks... :-)
`remove_user` is used below in the if statement. I think these two lines were intended to be at the original indent level.

VMAccess/vmaccess.py
paulmey: See my comment at line --/83

bingosummer: @Snesha 
Agree with @paulmey .
If you want to exit at line 88 or line 92, you should call hutil.do_exit with right status.

VMAccess/vmaccess.py
bingosummer: @Snesha 
Agree with @paulmey

VMAccess/vmaccess.py
bingosummer: @Snesha 
We should not perform _fsck_check() and _fsck_repair() at the same time, right?

VMAccess/vmaccess.py
```diff
@@ -308,18 +359,58 @@ def _open_ssh_port():
...
+      else:
+        # remount all post repair and return
+        retcode = waagent.Run("mount -a")
+      if retcode == 0:
+        return retcode
+      else:
+        raise Exception("Failed to mount disks")
+    except Exception, e:
+        hutil.error("%s" %(str(e), traceback.format_exc()))
+        hutil.do_exit(1, 'Repair','error','0', 'Repair failed.')
```
bingosummer: @Snesha 
The indent should be 4 spaces.
Please fix it.

VMAccess/vmaccess.py
```diff
@@ -308,18 +359,58 @@ def _open_ssh_port():
...
+    try:
+        retcode = waagent.Run("fsck -As -y")
+        if retcode > 0:
+            raise Exception("Disk check was not successful")
+        else:
+            return retcode
+    except Exception, e:
+        hutil.error(
+    "Failed to run disk check with error: %s" %
+     (str(e), traceback.format_exc()))
```
bingosummer: @Snesha 
This will break because there should be two "%s".

VMAccess/vmaccess.py
```diff
@@ -308,18 +359,58 @@ def _open_ssh_port():
...
+        raise Exception("Failed to repair disk")
+      else:
+        # remount all post repair and return
+        retcode = waagent.Run("mount -a")
+      if retcode == 0:
+        return retcode
+      else:
+        raise Exception("Failed to mount disks")
+    except Exception, e:
+        hutil.error("%s" %(str(e), traceback.format_exc()))
```
bingosummer: @Snesha 
This will break because there should be two "%s".

PR #24, opened 2015-08-07
Remove SteppingStone Extension

bingosummer:
SteppingStone Extension is an internal experimental extension.
We should remove it.


thomas1206:
Make sense.

PR #38, opened 2015-09-22
Enhance retry logic and add internal DNS check

bingosummer:


CustomScript/customscript.py
```diff
@@ -103,37 +103,81 @@ def enable(hutil):
...
+    But if download_files fails, we will wait and retry.
     """
     hutil.exit_if_enabled()
     prepare_download_dir(hutil.get_seq_no())
-    maxRetry = 2
-    for retry in range(0, maxRetry + 1):
+    public_settings = hutil.get_public_settings()
+    if public_settings is None:
+        raise ValueError("Public configuration couldn't be None.")
+    tries = public_settings.get('tries')
```
thomas1206: Shall we use retrycount?

CustomScript/customscript.py
```diff
@@ -103,37 +103,81 @@ def enable(hutil):
...
+    is_idns_ready = check_idns()
+    while (not is_idns_ready):
+        if retry == tries:
+            error_msg = "Internal DNS is not ready, ignore it."
+            hutil.error(error_msg)
+            waagent.AddExtensionEvent(name=ExtensionShortName,
+                                      op="CheckIDNS",
+                                      isSuccess=False,
+                                      version=Version,
+                                      message="(01306)"+error_msg)
```
thomas1206: We need not only add telemetry data for failed case, but for all cases. The reason for this is we want see when can we safely remove the iDNS check.

CustomScript/customscript.py
bingosummer: Here I followed the options of the command "wget".
```
Download:
  -t,  --tries=NUMBER            set number of retries to NUMBER (0 unlimits).
  -w,  --wait=SECONDS            wait SECONDS between retrievals.
```
If we used `retrycount`, then `retrycount` should be `tries -1`.

CustomScript/customscript.py
```diff
@@ -103,37 +103,81 @@ def enable(hutil):
...
+        is_idns_ready = check_idns()
+        retry = retry + 1
+    else:
+        msg = "Internal DNS is ready"
+        hutil.log(msg)
+        waagent.AddExtensionEvent(name=ExtensionShortName,
+                                  op="CheckIDNS",
+                                  isSuccess=True,
+                                  version=Version,
+                                  message="(01306)"+msg)
```
bingosummer: This is the successful case for checking iDNS.
We can determine when to safely remove the iDNS check by it.

CustomScript/customscript.py
thomas1206: OK. Looks good.

CustomScript/customscript.py
thomas1206: Like what we synced offline, please add telemetry for both downloading and iDNS retry count.

bingosummer:
Close it and I will create a new one.

PR #42, opened 2015-10-12
Refine CustomScript Extension

bingosummer:
1. Enhance retry logic and add internal DNS check in case of network failure
2. Change permission of extension log path to enhance security
3. Remove syslog handler from Utils

### DIFF
--- a/CustomScript/HandlerManifest.json
+++ b/CustomScript/HandlerManifest.json
@@ -1,6 +1,6 @@
 [{
   "name": "CustomScript",
-  "version": "1.3.0.0",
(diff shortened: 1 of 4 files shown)


AbelHu:
You can define a retry method and pass the operation as a parameter.

bingosummer:
@AbelHu You mean that the customer can define their own retry method?

thomas1206:
The retry logic here is to mitigate a known issue in Azure Network. In some situation, the vm network could not be ready for around serveral sec right after the vm get provisioned.

PR #51, opened 2015-11-17
Support passing sensitive data

bingosummer:
Users can put commandToExecute in the protected settings if they want pass their sensitive data, for e.g. password.
If commandToExecute is specified both in public and protected settings, the extension will fail with an error.

### DIFF
--- a/CustomScript/customscript.py
+++ b/CustomScript/customscript.py
@@ -85,12 +85,12 @@ def main():
             elif re.match("^([-/]*)(update)", a):
                 dummy_command("Update", "success", "Update succeeded")
     except Exception, e:
-        err_msg = ("Failed to enable the extension with error: {0}, "
+        err_msg = ("Failed with error: {0}, "
                    "{1}").format(e, traceback.format_exc())
         waagent.Error(err_msg)
         hutil.error(err_msg)
         hutil.do_exit(1, 'Enable','failed','0',
-                      'Enable failed: {0}'.format(e))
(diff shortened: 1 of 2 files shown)


thomas1206:
Looks good to me.

CustomScript/customscript.py
```diff
@@ -594,5 +592,20 @@ def get_formatted_log(summary, stdout, stderr):
     return msg_format.format(summary, stdout, stderr)
 
 
+def get_command_to_execute(hutil):
+    public_settings = hutil.get_public_settings()
+    protected_settings = hutil.get_protected_settings()
+    cmd_public = public_settings.get('commandToExecute')
```
andyliuliming: can we make 'commandToExecute' a constant string instead of hard code several times.

bingosummer: Yes. However, we can postpone this change to next version since it is not a blocking issue.

CustomScript/customscript.py
```diff
@@ -594,5 +592,20 @@ def get_formatted_log(summary, stdout, stderr):
...
+    cmd_protected = protected_settings.get('commandToExecute')
```
yuezh: I thought the parameter name should be different in public and protected settings. Correct me if I'm wrong.

bingosummer: The design is:
1. The protected settings field will have the same commandToExecute naming
2. If customer provide both then we will fail and not process the config and properly message the failure in status channel

PR #57, opened 2015-11-30
Updating fsck_repair 

Snesha:
To only unmount disk that failed to mount


Snesha:
cc @paulmey could you please CR

VMAccess/vmaccess.py
```diff
@@ -397,29 +399,26 @@ def _fsck_check(hutil):
...
-def _fsck_repair(hutil):
+def _fsck_repair(disk_name, hutil):
     # first unmount disks and loop devices lazy + forced
     try:
-        cmd_result = waagent.Run("umount -adfr")
+        cmd_result = waagent.Run("umount -f /%s" % disk_name)
         if cmd_result!=0:
             # Fail fast
-            raise Exception("Failed to unmount disks")
+            hutil.log("Failed to unmount disk")
```
bingosummer: I think, this line maybe needs to be removed because the error is logged in line 420.

VMAccess/vmaccess.py
```diff
@@ -397,29 +399,26 @@ def _fsck_check(hutil):
...
             retcode = waagent.Run("fsck -AR -y")
-        if retcode > 0:
-            raise Exception("Failed to repair disk")
-        else:
-            # remount all post repair and return
-            retcode = waagent.Run("mount -a")
+            hutil.log("Ran fsck")
         if retcode == 0:
-            return retcode
+            hutil.log(waagent.Run("mount"))
```
bingosummer: `waagent.Run` returns 0 or 1.
Should log some describable message instead of only the return code.

VMAccess/vmaccess.py
paulmey: maybe log diskname here as well?

VMAccess/vmaccess.py
```diff
@@ -397,29 +399,26 @@ def _fsck_check(hutil):
...
+            raise Exception("Failed to unmount disk %s" % disk_name)
         else:
             # run repair
             retcode = waagent.Run("fsck -AR -y")
-        if retcode > 0:
-            raise Exception("Failed to repair disk")
-        else:
-            # remount all post repair and return
-            retcode = waagent.Run("mount -a")
+            hutil.log("Ran fsck")
```
paulmey: log retcode

VMAccess/vmaccess.py
paulmey: waagent.Run apparently returns retcode, not the stdout. @bingosummer knows more

VMAccess/vmaccess.py
```diff
@@ -397,29 +399,26 @@ def _fsck_check(hutil):
...
+def _fsck_repair(disk_name, hutil):
     # first unmount disks and loop devices lazy + forced
     try:
-        cmd_result = waagent.Run("umount -adfr")
+        cmd_result = waagent.Run("umount -f /%s" % disk_name)
         if cmd_result!=0:
             # Fail fast
-            raise Exception("Failed to unmount disks")
+            hutil.log("Failed to unmount disk")
+            raise Exception("Failed to unmount disk %s" % disk_name)
```
paulmey: maybe soft fail here? since the disk may not have mounted at all if it was 'sufficiently' broken...
Then the 'else' on the next line should not be there and you should always fsck.

VMAccess/vmaccess.py
bingosummer: @paulmey is right.
waagent.run is executed twice in line 415 and 416. I think, it's not expected.
And we should log some describable message here instead of retcode only.

paulmey: Yeah, from @Snesha I got that the intent here was to show stdout of the mount command.

bingosummer: waagent.RunGetOutput will return retcode and output.
https://github.com/Azure/WALinuxAgent/blob/2.0/waagent#L2256

bingosummer:
@Snesha Please update VMAccess/README.md, add your scenario and give the sample configuration about how to use it. Thanks.

VMAccess/vmaccess.py
Snesha: Thanks! found it.

Snesha:
@bingosummer Just updated the PR and testing. I'll comment in a couple of days but could you also run integration tests?

VMAccess/vmaccess.py
```diff
@@ -397,29 +399,24 @@ def _fsck_check(hutil):
...
-        if retcode > 0:
-            raise Exception("Failed to repair disk")
-        else:
-            # remount all post repair and return
-            retcode = waagent.Run("mount -a")
+            hutil.log("Ran fsck with return code: %d" % retcode)
         if retcode == 0:
-            return retcode
+            hutil.log(waagent.RunGetOutput("mount"))
+            return waagent.RunGetOutput("mount")
```
bingosummer: Sample:
```
retcode, output = waagent.RunGetOutput("mount")
```
And it should not be executed twice.

bingosummer:
Will run the integration test next week.

VMAccess/vmaccess.py
Snesha: @bingosummer why can't it not be executed twice?

Snesha:
@bingosummer  Updated based on feedback and tested. please let me know when you can deploy.

Snesha:
@bingosummer LGTM. When can you run the test pass on your end?

bingosummer:
@Snesha Could you squash all your commits into a single one? Thanks.
If it doesn't break the regression test, I'll merge it and release a public 1.4 version EOW.

bingosummer:
@Snesha 
You don't need to open a new PR. In the branch `Snesha:sneshaf-dev`, you just run `git rebase -i <the-last-commit-before-all-your-commits>` to rebase your history.

bingosummer:
Create a new PR #81

PR #79, opened 2015-12-18
Adding in Check and Repair disk

Snesha:
fsck commands for users to be able to Check and Repair disk


Snesha:
@bingosummer

bingosummer:
See #81

PR #80, opened 2015-12-18
Fixing parsing error at mounts.py

chan-alex:
Hi, 

This patch fixes a bug in mounts.py.

Currently Mounts.__init__(self,logger)  parses output from lsblk incorrectly if there are space characters present in any of its output. This can occur if linux LVM is being used.

An example is this output from "lsblk -l -n -o NAME,TYPE,FSTYPE,MOUNTPOINT"

       backups_vg-hk_static_backup_lv-real (dm-0)             lvm   ext4 

Will result in a "Mount" object with the following fields:

        self.name = backups_vg-hk_static_backup_lv-real
        self.type = (dm-0) <---- this is wrong
        self.fstype = lvm   <---- this is wrong
        self.mount_point = ext4   <---- this is wrong

This inaccurate parsing will result in the following (detected) behavior:

  1) All filesystem will be freezed during the backup (because self.mount_point is never null)
  2) The waagent will hang if there are many of this misparsed lines (which happens if there are many LVM snapshots)

The patch fixes this by using the "-p" option from lsblk and parsing the output more accurately.

### DIFF
--- a/VMBackup/main/mounts.py
+++ b/VMBackup/main/mounts.py
@@ -48,17 +48,17 @@ class Mounts:
         out_lsblk_output = str(out_lsblk_output)
         self.logger.log(msg="out_lsblk_output:\n" + str(out_lsblk_output),local=True)
         lines = out_lsblk_output.splitlines()
-        line_number = len(lines)
-        for i in range(0,line_number):
-            item_value = lines[i].strip().split()
-            print("item_value==" + str(item_value))
-            name = item_value[0]
-            type = item_value[1]
-            fstype = ""
-            mountpoint = ""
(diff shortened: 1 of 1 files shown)


chan-alex:
The bug is real. It hangs the waagent for my VMs.

andyliuliming:
@chan-alex  thanks reporting this issue, we are fixing it now. thanks.

ahmetalpbalkan:
@soumyanishan can you please take this over?

soumyanishan:
@andyliuliming looks like the fix is already done. Can you please confirm

soumyanishan:
the fix is already there

PR #84, opened 2016-01-06
Update README.md

squillace:
Added the timeout information; not clear whether that information is ARM only, or ASM as well.


squillace:
@thomas1206 :-) @Snesha

thomas1206:
Thanks @squillace

thomas1206:
It's for both ARM and ASM

PR #98, opened 2016-02-19
Fix for Internal DNS Check

libby2cm:
I needed the ability to disable the internal DNS check, and while the public_setting is exposed, there is a bug where it was not being handled correctly and didn't work.  Setting a var to a string value makes it a 'true' boolean value when used for a boolean operation regardless of what the string value actually is.

This change uses the strtobool function to convert the string value that is optionally defined in public_settings ('enableInternalDNSCheck') into a true boolean value for the 'enable_idns_check' variable, so that the idns check is executes as expected.  Omitting 'enableInternalDNSCheck' from the public_settings results in a default of 'true' (as it was) while allowing a string value of 'y, yes, t, true, on or 1' for boolean true and a string value of 'n, no, f, false, off and 0' as a boolean false.

### DIFF
--- a/CustomScript/customscript.py
+++ b/CustomScript/customscript.py
@@ -39,6 +39,7 @@ import json
 from codecs import *
 from azure.storage import BlobService
 from Utils.WAAgentUtil import waagent
+from distutils.util import strtobool
 import Utils.HandlerUtil as Util
 import Utils.ScriptUtil as ScriptUtil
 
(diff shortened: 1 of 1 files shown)


bingosummer:
@libby2cm 
Could you please set `enableInternalDNSCheck` to a bool value `true` or `false` in your public settings? It should work to disable the internal DNS check.

libby2cm:
..I tried that, it doesn't work for the reason I mention in comments, hence
the fix I submitted..

bingosummer:
This is the example. https://github.com/Azure/azure-quickstart-templates/blob/master/bosh-setup/azuredeploy.json#L261

Could you please share your public settings? Using Azure CLI or Powershell?

ahmetalpbalkan:
@boumenot would be great if you can follow up on this PR.

boumenot:
Thanks for the ping.  I like the change.

PR #114, opened 2016-03-30
Atomically write the status file.

boumenot:


boumenot:
I'm not sure if this will correct the issue or not, and it is unclear from the documentation how to verify this fix.  But, I figured I would start off with code to better illustrate what we were thinking.

hosungsmsft:
Thanks much @boumenot . I've been seeing failures related to the status file unavailability, and I'd be interested in knowing more about this fix. Will sync up with you offline when I get some time from the current urgent coding work.

PR #117, opened 2016-04-05
Atomically write the status file.

boumenot:
Diagnostics has their own copy of Utils, and this is the copy used by CustomScript.


boumenot:
This fixes #114.

PR #130, opened 2016-05-03
[LAD] OMI install/uninstall changes

hosungsmsft:
- OMI install is retried (up to 3 times) if it fails to install.
- OMI is uninstalled when LAD is uninstalled.

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -87,7 +87,6 @@ RedhatConfig =  {"installomi":"bash "+omi_universal_pkg_name+" --upgrade --force
                  'mdsd_env_vars': {"SSL_CERT_DIR": "/etc/pki/tls/certs", "SSL_CERT_FILE": "/etc/pki/tls/cert.pem"}
                  }
 
-
 UbuntuConfig1510OrHigher = dict(DebianConfig.items()+
                     {'installrequiredpackages':'[ $(dpkg -l PACKAGES |grep ^ii |wc -l) -eq \'COUNT\' ] '
                         '||apt-get install -y PACKAGES',
@@ -165,21 +164,21 @@ def parse_context(operation):
     return
 
 
-def setup(local_only):
+def setup(should_install_required_package):
     global EnableSyslog
 
-    if not local_only:
-        install_package_error=""
+    if should_install_required_package:
+        install_package_error = ""
         retry = 3
-        while retry >0:
-            error,msg = install_required_package()
+        while retry > 0:
+            error, msg = install_required_package()
             hutil.log(msg)
-            if error ==0:
+            if error == 0:
                 break
             else:
-                retry-=1
+                retry -= 1
                 hutil.log("Sleep 60 retry "+str(retry))
-                install_package_error=msg
+                install_package_error = msg
                 time.sleep(60)
         if install_package_error:
             if len(install_package_error) > 1024:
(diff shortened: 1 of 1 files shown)


hosungsmsft:
@LiweiPeng & @jasonzio : Please review this PR. Thanks.

Diagnostic/diagnostic.py
```diff
@@ -750,7 +758,17 @@ def install_omi():
...
-        RunGetOutput(distConfig["installomi"])
+        isOmiInstalledSuccessfully = False
+        for trialNum in range(1, 4):   # Retry up to 3 times to install OMI
+            isOmiInstalledSuccessfully = RunGetOutput(distConfig["installomi"])[0] is 0
+            if isOmiInstalledSuccessfully:
+                break;
+            hutil.error("OMI install failed (trial #" + str(trialNum) + "). Retrying in 30 seconds...")
+            time.sleep(30)
+        if not isOmiInstalledSuccessfully:
+            hutil.error("OMI install failed 3 times. Giving up...")
```
jasonzio: Failed *4* times if you get here.

jasonzio:
I'm a bit uncomfortable with the LAD uninstall forcing an uninstall of OMI; there may be other users of OMI also on the system, and a unilateral uninstall would break them.

LiweiPeng:
comment about the time.sleep(30) at line 767: you don't need to sleep another 30-second when it fails in the last loop. when it fails in the last loop, it should just giving up instead of wasting another 30s.

Diagnostic/diagnostic.py
hosungsmsft: @jasonzio : range(1,4) in python is [1,2,3], so I believe it's going to try 3 times...?

hosungsmsft:
Thanks @jasonzio and @LiweiPeng . Jason, I'm OK with not uninstalling OMI. Will revert that change. Liwei, thanks for pointing that out. Yes, the last wait shouldn't happen. I'll update the code and commit another change.

Diagnostic/diagnostic.py
```diff
@@ -750,7 +754,20 @@ def install_omi():
...
         hutil.log("Begin omi installation.")
-        RunGetOutput(distConfig["installomi"])
+        isOmiInstalledSuccessfully = False
+        maxRetries = 3      # Retry up to 3 times to install OMI
+        for trialNum in range(1, maxRetries+1):
+            isOmiInstalledSuccessfully = RunGetOutput(distConfig["installomi"])[0] is 0
+            if isOmiInstalledSuccessfully:
+                break;
+            hutil.error("OMI install failed (trial #" + str(trialNum) + ").")
+            if trialNum < maxRetries:
```
jasonzio: Needs to be <= maxRetries, otherwise you won't sleep when trialNum = maxRetries (the 3rd install, which is the 2nd retry).

jasonzio:
Still didn't quite get the retry-counting logic correct with respect to the 30 second delay.

Diagnostic/diagnostic.py
hosungsmsft: Oh... My bad... Sorry about the confusion. I meant "maxTries", not "maxRetries". Does that make sense? I'll change the naming and push another commit. Thx.

jasonzio:
Nevermind. I had a fundamental misunderstanding of python's range() function. (It's half-open above, not closed.)

jasonzio:
LGTM

hosungsmsft:
Thanks much!

Issue #143, opened 2016-05-25
Wrong version reflected in the logs


ahmetalpbalkan:

ahmetalpbalkan:
[oops hit return too early]

as shown in #142 , I think CustomScriptForLinux  relies on HandlerEnvironment.json to find its own version (same mistake LAD does: #140) and therefore logs version `1.0` even though it is `1.5.2.0`.

hosungsmsft:
Yes. I'm looking at a fix, but it doesn't seem straightforward due to some reason. We need to change the _get_log_prefix() implementation (see the link below, which uses "name" and "version" from HandlerEnvironment.json, where "name" is the extension name, but "version" is the handler version that's fixed to 1.0, adding some confusion).

I know it's not good, but since LAD already forked the Utils directory, I'll try a fix on LAD's Utils code first, and share it out later.

https://github.com/Azure/azure-linux-extensions/blob/master/Utils/HandlerUtil.py#L82

ahmetalpbalkan:
@hosungsmsft any updates?

hosungsmsft:
@ahmetalpbalkan - We already committed a fix for this on LAD ( #148 ), but as I mentioned, LAD's Utils directory was forked to its own copy, before we took it over. So the fix is not on other extensions. We may need to discuss if we want to unify the two Utils forks back into one, and if so, how to achieve that. For the time being, I'd be more than happy to apply the changes in #148 to the top-level Utils code, but I wouldn't be able to test it on other extensions.

boumenot:
@hosungsmsft, please apply the change in a branch, and I will validate them with CustomScript.

boumenot:
@ahmetalpbalkan, based on Hosung's comments I don't think this is specific to CustomScript.  If an extension uses the common utils, they will also have this issue.  I'm dropping the CustomScript prefix to denote that more clearly.

hosungsmsft:
@boumenot - Created a branch ('ext-ver-log') and applied the change on Diagnostic/Utils/HandlerUtility.py to Utils/HandlerUtility.py. Please note that every HandlerUtility construction code should be updated with additional parameters, like https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/diagnostic.py#L171 .

Issue #140, opened 2016-05-24
LAD should report proper extension version


LiweiPeng:
In a VM I deployed on 5/23/2016 Ubuntu 14.04 (5/17/2016 release), with waagent 2.0.16 and LinuxDiagnostics (LAD) 2.3.9, I found misleading version information in the LAD extension.log file. see details below.

- What's in the log: the version is "1.0" in HandlerEnvironment.json, and HanderManifest.json.
- What's expected from a user point of view: as a user of LAD, when looking at extension.log, I am expecting to see something like "Microsoft.OSTCExtensions.LinuxDiagnostic-2.3" or "Microsoft.OSTCExtensions.LinuxDiagnostic-2.3.9" because that's the version I am using.

Log file:
 2016/05/23 17:31:28 [Microsoft.OSTCExtensions.LinuxDiagnostic-1.0]sequence number is 0
 2016/05/23 17:31:28 [Microsoft.OSTCExtensions.LinuxDiagnostic-1.0]setting file path is/var/lib/waagent/Microsoft.OSTCExtensions.LinuxDiagnos
 tic-2.3.9/config/0.settings2

Related files:
  HandlerEnvironment.json
 [{ "name": "Microsoft.OSTCExtensions.LinuxDiagnostic", "seqNo": "0", "version": 1.0, …

 HandlerManifest.json
 [{
 "name": "LinuxAzureDiagnostic",
 "version": 1.0,

> azure vm extension list -n LinuxDiagnostic
...
Version : 2.3

hosungsmsft:
Looked at this again, and confirmed that this is a waagent issue. The HandlerEnvironment.json is created by waagent, and the version number (1.0) is hard-coded there. LAD just reads HandlerEnvironment.json and use the retrieved version number in logging prefix.

HandlerManifest.json is read by waagent, but apparently, the version in HandlerManifest.json is not the version written to the HandlerEnvironment.json. Will still try to update in the future.

Closing this issue as such, and will reactivate the original issue on waagent repo. I also have a fix, and start a PR there as well. Thanks.

hosungsmsft:
Same as #143. Will try a fix on LAD's Utils code first.

hosungsmsft:
#148 fixes this, and it's merged, so closing this.

PR #148, opened 2016-05-25
[LAD] Misc updates for 2.3.9001

hosungsmsft:
- Remove unused XML attribute to avoid unnecessary mdsd warning
- Remove unnecessary dependency installation
- Report proper extension version in logs (addressing Issue #140)

@LiweiPeng : Please review this for Issue #140 and other issues you raised (the mdsd warning).

@ahmetalpbalkan : This applies to LAD only (because LAD's Utils directory is forked before we took over), so this change doesn't directly address #143, but I suppose a similar approach could be applied there. Note that each extension's HandlerUtility instantiation calls will need to be updated, which I suppose we should be very careful about... Also I still don't like the fact that the extension's version is hard-coded in the script, but I couldn't find the version string from anywhere else. Let me know if you know about this. Thanks.

### DIFF
--- a/Diagnostic/Utils/HandlerUtil.py
+++ b/Diagnostic/Utils/HandlerUtil.py
@@ -67,13 +67,15 @@ class HandlerContext:
         return
 
 class HandlerUtility:
-    def __init__(self, log, error, short_name):
+    def __init__(self, log, error, short_name, ext_name_in_log_prefix, ext_version):
         self._log = log
         self._error = error
         self._short_name = short_name
+        self._ext_name_in_log_prefix = ext_name_in_log_prefix
+        self._ext_version = ext_version
 
     def _get_log_prefix(self):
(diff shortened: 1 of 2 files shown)


LiweiPeng:
the code change looks good to me. Thanks.

ahmetalpbalkan:
when is this going to be published? I'd love to see if the ExtensionsConfig specifies the package from an https location.

hosungsmsft:
@ahmetalpbalkan : Hmm... actually it's not planned for a certain deadline yet, because issues we are currently addressing are mostly pretty old. There's one fix on mdsd-side (binary) that might be a potential fix for ongoing memory management-related mdsd crashes, and if that fix is well tested and approved, we might want to publish the new version sooner, like maybe next week. Let's see how it goes, and will let you know when the new version is published. Thanks.

PR #149, opened 2016-05-26
fix a converting bug

bingosummer:


CustomScript/customscript.py
```diff
@@ -322,7 +322,13 @@ def download_blob(storage_account_name, storage_account_key,
                                         blob_uri,
                                         download_dir)
         blob_name, _, _, download_path = result
-        preprocess_files(download_path, hutil)
+        enable_dos_converting = True
+        public_settings = hutil.get_public_settings()
+        if public_settings:
+            if 'enableDOSConverting' in public_settings:
```
ahmetalpbalkan: is `enableDOSConverting` being newly introduced? if so we need a docs update as well

CustomScript/customscript.py
```diff
@@ -405,6 +417,8 @@ def to_process(file_path, extensions=['.sh', ".py"]):
     for extension in extensions:
         if file_path.endswith(extension):
             return True
+        else:
```
ahmetalpbalkan: this now makes anything below this line (and the rest of the elements in the `extensions` array) dead code.

CustomScript/customscript.py
```diff
@@ -405,6 +417,8 @@ def to_process(file_path, extensions=['.sh', ".py"]):
...
+            return False
```
ahmetalpbalkan: this now makes anything below this line (and the rest of the elements in the `extensions` array) dead code.
 Add a line note

CustomScript/customscript.py
```diff
@@ -368,7 +374,13 @@ def download_external_file(uri, command, hutil):
     file_path = os.path.join(download_dir, file_name)
     try:
         download_and_save_file(uri, file_path)
-        preprocess_files(file_path, hutil)
+        enable_dos_converting = True
+        public_settings = hutil.get_public_settings()
+        if public_settings:
```
ahmetalpbalkan: lines 378-383 are repeated exactly the same in both methods, please extract them into a single place and pass it as a flag to these methods.

CustomScript/customscript.py
```diff
@@ -405,6 +417,8 @@ def to_process(file_path, extensions=['.sh', ".py"]):
     for extension in extensions:
         if file_path.endswith(extension):
             return True
+        else:
+            return False
     with open(file_path, 'rb') as f:
         contents = f.read(64)
```
jasonzio: Why are you searching the first 64 characters? Near as I can tell, there are only two cases:
- File begins with a BOM. Strip the BOM, convert from UTF-16 (in the byte gender indicated by the BOM) to UTF-8, and convert /r/n to /n
- File begins with #!. Just convert \r\n to \n.

When I say "begins with", I really mean that - the absolute first characters of the file.

I'm still not a fan of this autoconversion; it will bite you eventually. But tightly restricting the scope in which you apply the conversion is your best defense.

CustomScript/customscript.py
alyukhan: please consider to update preprocess_files doc string

CustomScript/customscript.py
alyukhan: would suggest to delete the rest of the code in this function

theunrepentantgeek: Would be better to delete the else branch and fix the logic for #! below.

alyukhan: I don't like this idea. For example, I may have binary data after __DATA__ literal in perl script.

CustomScript/customscript.py
bingosummer: Sure. Will update the codes according to your comments. Many thanks.

ahmetalpbalkan:
any updates on this PR?

bingosummer:
Sorry, just came back from a business trip. Will update this PR soon.

bingosummer:
1. Since most of us don't like the idea of autoconversion, so I'd like to set the default value of `enableDOSConverting` to `False`.
2. As @alyukhan suggested, I removed the logic for `#!`. It's much more simple and straightforward.

WDYT?

PS.
1. I prefer to rebase this PR before you merge it.
2. Please verify the code changes (I didn't run the tests) and update the changelog before publishing it.

ahmetalpbalkan:
@boumenot please review

PR #159, opened 2016-06-09
[LAD] Fix broken syslog collection on Redhat

hosungsmsft:
@jasonzio @LiweiPeng - Please review this.

Not knowing semanage is not installed by default on Redhat, I removed the policycoreutils-python install step, which caused syslog collection to stop working on Redhat-based distros. It's a good thing that this caught the potential issue of broken syslog collection when a randomly available port was assigned. That potential issue is also fixed here as well.


jasonzio:
lgtm

PR #155, opened 2016-06-06
[LAD] Add OMI diag dump code, enable core dump

hosungsmsft:
@jasonzio @LiweiPeng - Please review this PR. I'm aiming at releasing LAD 2.3.9001 after this PR.

PR #152, opened 2016-05-27
[LAD] Support dynamic mdsd port

hosungsmsft:
@jasonzio, @LiweiPeng - Please review. Thanks.

PR #151, opened 2016-05-26
[LAD] Shouldn't always kill mdsd process by 'kill -9'

hosungsmsft:
@LiweiPeng : Please review. This is a fix for #139 . Thanks.

PR #138, opened 2016-05-23
[LAD] Fix 'enableSyslog' config property name

hosungsmsft:
@LiweiPeng, please review. Thanks.

PR #160, opened 2016-06-09
[LAD] Merge LAD-v.2.3.9001 commits to master for official release

hosungsmsft:
Major commits were already reviewed by @jasonzio and @LiweiPeng. This is a formality PR for merging to master. Confirmed that all existing E2E tests pass on all tested distros/versions. We are ready to release LAD v.2.3.9001.


ahmetalpbalkan:
👍

hosungsmsft:
Just mentioning the actual sub-PRs with actual review comments. Should have used GitHub merges to show them automatically, I guess. Will try that in the future.
#159, #155, #152, #151, #148, #138, #136, #134, #133, #131 #130, #129, #126

ahmetalpbalkan:
@hosungsmsft perhaps we should evaluate moving LAD to its own separate GitHub repository. Problem with all these extensions in this repo is, we can't really tag versions. So it's hard to identify which PR made it to which release.

Good thing is, we now have a single merge commit, merging this version of LAD to master. So people can `git checkout` to that merge commit and retrieve the code tree of the version a lot more clearly now.

hosungsmsft:
@ahmetalpbalkan - Agreed, and it has been a question since we first started to take over extensions, starting with LAD. The sticky thing would be the Utils directory, and as I mentioned, LAD already forked out, and we may need to consider whether to merge back the forked Utils or not. I guess other extensions moving out would be based on those extension owners' choices.

ahmetalpbalkan:
@hosungsmsft the healthiest option is to develop a common extensions utilities in a separate repository and periodically sync with that. We've seen extensions rely on `waagent` even for various [simple functionality](https://github.com/Azure/WALinuxAgent/issues/223#issuecomment-224713170) as well as complicated things like wireserver telemetry (which they probably shouldn't be doing).

Docker Extension ships with a [VM extension utils package for Go](https://github.com/Azure/azure-docker-extension/tree/master/pkg/vmextension) which are used by LinuxAsm as well as DockerExtension.

Probably decoupling waagent + utils here from all the extensions into a separate module (perhaps even on PyPi?) and moving all extensions to their own repos is easier for releasing, tagging and issue tracking. If you folks could find time to make this a reality, it would set a nice example.

PR #162, opened 2016-06-10
[VMBackup] Freeze issue

soumyanishan:
final commits done for freeze issue

### DIFF
--- a/VMBackup/main/handle.py
+++ b/VMBackup/main/handle.py
@@ -165,7 +165,7 @@ def snapshot():
         backup_logger.log(errMsg, False, 'Error')
     snapshot_done = True
 def daemon():
-    global MyPatching,backup_logger,hutil,run_result,run_status,error_msg,freezer,para_parser
+    global MyPatching,backup_logger,hutil,run_result,run_status,error_msg,freezer,para_parser,snapshot_done
     #this is using the most recent file timestamp.
     hutil.do_parse_context('Executing')
     freezer = FsFreezer(patching= MyPatching, logger = backup_logger)
@@ -173,16 +173,31 @@ def daemon():
     # precheck
     freeze_called = False
 
-    configfile=os.path.curdir+'/config/config.ini'
+    configfile='/var/lib/waagent/extension_config/config.ini'
+    thread_timeout=str(60)
     if not os.path.exists(configfile):
-        fopen=open(configfile,'w')
-        fopen.write('[SnapshotThread]\ntimeout=60')
-        fopen.close()
-
-    config = ConfigParser.ConfigParser()
-    config.read(configfile)
-    thread_timeout= config.get('SnapshotThread','timeout')
-    backup_logger.log(thread_timeout,local=True)
+        directory_present= True
+        try:
+            os.makedirs(os.path.dirname(configfile))
+        except OSError as e:
+            if e.errno != err.EEXIST:
+                directory_present=False
+        if(directory_present):
+            try:
+                fopen=open(configfile,'w')
+                fopen.write('[SnapshotThread]\ntimeout=60')
+                fopen.close()
+            except Exception as e:
+                errMsg='Config file not present and cannot be created'
+                backup_logger.log(errMsg, False, 'Warning')
+    try:
+        config = ConfigParser.ConfigParser()
+        config.read(configfile)
+        thread_timeout= config.get('SnapshotThread','timeout')
+        backup_logger.log(thread_timeout,local=True)
+    except Exception as e:
+        errMsg='cannot read config file or file not present'
+        backup_logger.log(errMsg, False, 'Warning')
 
     try:
         # we need to freeze the file system first
@@ -250,11 +265,11 @@ def daemon():
                 """
                 make sure the log is not doing when the file system is freezed.
                 """
+                temp_status= 'transitioning'
+                temp_result=CommonVariables.success
+                temp_msg='Transitioning state in extension'
+                trans_report_msg = None
                 if(para_parser is not None and para_parser.statusBlobUri is not None and para_parser.statusBlobUri != ""):
-                    temp_status= 'transitioning'
-                    temp_result=CommonVariables.success
-                    temp_msg='Transitioning state in extension'
-                    trans_report_msg = None
                     trans_report_msg = do_backup_status_report(operation='Enable',status=temp_status,\
                                     status_code=str(temp_result),\
                                     message=temp_msg,\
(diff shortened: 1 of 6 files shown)


soumyanishan:
The freeze issue code fix is provided to customers as privates and it is working. pushing the code to master to release it

VMBackup/main/handle.py
```diff
@@ -165,33 +164,41 @@ def snapshot():
...
-    global MyPatching,backup_logger,hutil,run_result,run_status,error_msg,freezer,para_parser
+    global MyPatching,backup_logger,hutil,run_result,run_status,error_msg,freezer,para_parser,snapshot_done
     #this is using the most recent file timestamp.
     hutil.do_parse_context('Executing')
     freezer = FsFreezer(patching= MyPatching, logger = backup_logger)
     global_error_result = None
     # precheck
     freeze_called = False
+
+    configfile='/var/lib/waagent/extension_config/config.ini'
```
ahmetalpbalkan: @soumyanishan sorry for catching this late but I think you shouldn't create a config file here

ahmetalpbalkan:
@soumyanishan any reason why you closed?

soumyanishan:
@ahmetalpbalkan my bad. reopening it

VMBackup/main/handle.py
```diff
@@ -165,33 +164,41 @@ def snapshot():
...
+        directory_present= True
+        try:
+            os.makedirs(os.path.dirname(configfile))
+        except Exception as e:
+            if e.errno != err.EEXIST:
+                directory_present=False
+        if(directory_present):
+            try:
+                fopen=open(configfile,'w')
+                fopen.write('[SnapshotThread]\ntimeout=60')
```
ahmetalpbalkan: can't you instead have a default value in `config.get` instead of writing a configuration file for the user?

I suggest you should make this default config file as part of your `"install"` step in your extension for the future users.

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan what is the problem in this path? Basically I need a path which will be independent of new extension installation

ahmetalpbalkan: @soumyanishan no other program writes configuration here. even waagent's own config is at `/etc/waagent.conf`. `/var/lib` is usually used for app data, which can be nuked at anytime. For instance you want to nuke your `/var/lib/mysql` and your config lives at `/etc/my.cnf` or something like that. 

In this case my suggestion was bad, too. `/etc/..` is better. I don't think we can sign off on polluting under `/var/lib/waagent` directory with a brand new convention like this. We are not even planning to keep `/var/lib/waagent` path around.

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan  do you want me to move the config file creation code to install method?

ahmetalpbalkan: ideally yes, this way new installations will have a single place where the file is created and another separate place where it is read.

waagent-2.0 had the same mistake, it was editing the configuration file as they parse, which pretty much made the code undebuggable. I see you have a 200-line `daemon()` method as well and I suggest splitting this into smaller and more testable/readable pieces, such as:

```py
create_config_file_if_not_exists(path)
cfg = parse_config_file(path)
# use cfg.thread_timeout
```

instead of doing everything inline. Unfortunately existing extensions did not particularly set a good precedent in terms of coding practices.

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan I agree with you. so lets keep the path as /etc/extension_config/config.ini. does that look fine with you?

ahmetalpbalkan: @soumyanishan I don't think `extension_config` usually won't mean anything to anyone, and you're already in /etc, meaning it's already a config. How about: `/etc/azure/vmbackup.conf` (or `.cfg`).

`.ini` is a Windows convention, let's not bring it to Linux. 😉

soumyanishan: @ahmetalpbalkan /etc/azure/vmbackup.conf  looks good

VMBackup/main/handle.py
ahmetalpbalkan: Also, if you move this to install method, existing users of the agent will not be getting this file, so you might want to have a `create_config_file_if_not_exists` method and call it both here and in `install()` for a little longer.

VMBackup/main/handle.py
```diff
@@ -165,33 +164,41 @@ def snapshot():
...
+                fopen.close()
+            except Exception as e:
+                errMsg='Config file not present and cannot be created'
+                backup_logger.log(errMsg, False, 'Warning')
+    try:
+        config = ConfigParser.ConfigParser()
+        config.read(configfile)
+        thread_timeout= config.get('SnapshotThread','timeout')
```
ahmetalpbalkan: What if users remove this value from the config file? You should probably have a default value for it (also documented in your README).

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan there are lots of refactoring to be done to the extension code. We will take it up in the next iteration. As of now we are planning to release this code as soon as possible to fix the freeze issue.

VMBackup/main/handle.py
ahmetalpbalkan: This comment also goes further, you can probably operate totally without a config file too. Most Linux applications would work just fine if you delete their configuration files, because they have good defaults. Those don't usually need a mandatory configuration key from the config files, but in your case, you just have one key.

Did any of your customers ask to make `SnapshotThread` configurable?

soumyanishan: @ahmetalpbalkan will check with PM how to document it. if there is any exception while fetching the value from config then the default value will be 60.

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan wont the install be called when a new version of extension is published? If it is called then our objective is fulfilled.

ahmetalpbalkan: @soumyanishan depends on the version. if it is a hotfix release (major.minor.**hotfix**), then I believe the agent will call its `"upgrade"` entrypoint first, then `"enable"`.

If it's a major/minor release, the existing VMs will not get it. Only removing/adding the extension and the new VMs will be getting the new version and they will call `"install"` only once.

That's why I suggest having a little strong logic here like calling `create_config_file_if_not_exists` twice (here and in 'install').

Or better, as I've said before, avoid the file altogether and just hardcode a default for now. If you want to make it configurable, you can still look for a file, but perhaps your users should just create the file themselves only if they need it?

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan there is no request from customer. Earlier the timeout was 5 mins. we reduced it to 60 seconds by bringing parallelization in snapshotting. But the fsfreeze is out of our control, it could take more time if there are lots of I/O going on with the filesystem. In that case customer can increase the timeout to have the backup successful. By doing this we can avoid the regular timeout scenario for some customers

VMBackup/main/handle.py
soumyanishan: @ahmetalpbalkan this looks a better idea. Let the customer create themselves if they need it.

VMBackup/main/handle.py
ahmetalpbalkan: ok fair point, there's nothing wrong with the existing code in terms of functionality, it is just less maintainable.I think we can merge if you just commit the config file path change.

soumyanishan: @ahmetalpbalkan Monday I will do the changes. I think its better to remove the file creation step and let the customer create it as on when required.

ahmetalpbalkan: Fantastic.

soumyanishan:
will send another pull request with some other changes

Issue #145, opened 2016-05-25
README latest versions go out of date easily


ahmetalpbalkan:
It's probably not a good idea to list a latest version in the README:

1. the information goes out-of-date very easily
2. a version number like `1.5.2.0` cannot be used in ARM templates or CLI. API allows only `major.minor` precision. so it's almost useless.

therefore I suggest listing versions  only in Changelog for information purposes.

> ![image](https://cloud.githubusercontent.com/assets/159209/15558340/283d0bce-228e-11e6-8056-04deabb721d7.png)

boumenot:
This issue appears to override #147.  What would you like?

ahmetalpbalkan:
@boumenot this is more broad (for extensions other than CSE)

PR #164, opened 2016-06-13
Support for CI of extensions.

boumenot:
1. Specify the namespace for all extensions.
 2. Set the version number to the latest released public version.
 3. Add manifest.xml for all extensions.
 4. Re-format JSON documents according to jq's will.


boumenot:
@hosungsmsft, @mbearup, @ahmetalpbalkan: please review.

Diagnostic/HandlerManifest.json
```diff
@@ -1,15 +1,16 @@
...
-   "installCommand": "diagnostic.py -install",
-    "uninstallCommand": "diagnostic.py -uninstall",
-    "updateCommand": "diagnostic.py -update",
-    "enableCommand": "diagnostic.py -enable",
-    "disableCommand": "diagnostic.py -disable",
-    "rebootAfterInstall": false,
-    "reportHeartbeat": false,
+    "namespace": "Microsoft.OSTCExtensions",
+    "name": "LinuxDiagnostic",
+    "version": "2.3.9001",
```
ahmetalpbalkan: This `"version"` is the version for the schema of this file; not extension handler. And it should be a `float` with value `1.0` per spec.

Please read https://github.com/Azure/azure-marketplace/wiki/User%20Guide%20for%20writing%20extension%20handlers.pdf or see https://github.com/Azure/azure-docker-extension/blob/master/metadata/HandlerManifest.json as a reference.

hosungsmsft: I've heard from @ahmetalpbalkan that this version is not the extension version, but the handler version, so it should be all 1.0. Ahmet should confirm.

VMAccess/HandlerManifest.json
```diff
@@ -1,13 +1,14 @@
 [{
```
ahmetalpbalkan: these formattings don't look like `jq` to me. I get `[` and `{` on separate lines?

VMBackup/HandlerManifest.json
```diff
@@ -1,15 +1,14 @@
...
+[{
+    "namespace": "Microsoft.OSTCExtensions", 
+    "name": "VMBackupForLinuxExtension", 
+    "version": "0.1.0.993",
+    "handlerManifest": {
+        "disableCommand": "main/handle.py -disable", 
+        "enableCommand": "main/handle.py -enable", 
+        "installCommand": "main/handle.py -install", 
+        "rebootAfterInstall": false, 
+        "reportHeartbeat": false,
```
ahmetalpbalkan: maybe standardize order of the fields as well? this one particularly seems a bit off

VMAccess/HandlerManifest.json
boumenot: Fixed.  jq converts 1.0 to 1.  I would rather keep the experience pure so we can just run the tool in the future without having to hand edit it afterwards.

Diagnostic/HandlerManifest.json
boumenot: Fixed.

CustomScript/HandlerManifest.json
```diff
@@ -1,13 +1,17 @@
...
-    "updateCommand": "customscript.py -update",
-    "enableCommand": "customscript.py -enable",
-    "disableCommand": "customscript.py -disable",
-    "rebootAfterInstall": false,
-    "reportHeartbeat": false
+[
+  {
+    "namespace": "Microsoft.OSTCExtensions",
+    "name": "CustomScriptForLinux",
+    "extension_version": "1.5.2.0",
```
ahmetalpbalkan: whyy?

VMBackup/HandlerManifest.json
boumenot: I alphabetized the the invocations of the script, followed by the boolean fields, followed by the enum fields (if any).

CustomScript/HandlerManifest.json
```diff
@@ -1,13 +1,17 @@
...
+    "version": 1,
```
ahmetalpbalkan: please make it a `1.0` explicitly, you may lose 2 bytes but you may end up saving backwards compatibility or something like that when waagent changes one day and decides to look at this field. everything is python so this is prolly gonna deserialize into an int rather than float.

CustomScript/HandlerManifest.json
boumenot: The CI tool needs to know the version that you're deploying, and I need to store that metadata somewhere.  I can store it here, manifest.xml, or some other document.

CustomScript/HandlerManifest.json
boumenot: Sure.  We lose the ability to just run jq, which is a big negative IMHO.

CustomScript/HandlerManifest.json
ahmetalpbalkan: @boumenot another place would be great. let's not introduce random fields in the schematized object and add one more place to maintain the versions. as you can see in #145 we're not particularly doing a good job maintaining version values in multiple places

CustomScript/HandlerManifest.json
ahmetalpbalkan: @boumenot unless you dynamically need to update this document (you should not) I see you already ran jq once so it should be fine. I'd rather rely on the value covered by the spec than losing the ability to use jq.

OSPatching/manifest.xml
```diff
@@ -0,0 +1,18 @@
...
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
+  <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
+  <Type>OSPatchingForLinux</Type>
+  <Version></Version>
+  <Label>Microsoft Azure OS Patching Extension for Linux IaaS</Label>
+  <HostingResources>VmRole</HostingResources>
+  <MediaLink></MediaLink>
+  <Description>Microsoft Azure OS Patching Extension for Linux IaaS</Description>
+  <IsInternalExtension>true</IsInternalExtension>
+  <Eula>https://github.com/Azure/azure-linux-extensions/blob/1.0/LICENSE-2_0.txt</Eula>
```
ahmetalpbalkan: these are 404, please point it to master.

CustomScript/manifest.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version='1.0' encoding='utf-8' ?>
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
+  <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
+  <Type>CustomScriptForLinux</Type>
+  <Version></Version>
+  <Label>Microsoft Azure Custom Script Extension for Linux IaaS</Label>
```
ahmetalpbalkan: We don't publicly use word `IaaS` in our branding. I suggest using `Virtual Machines`

CustomScript/HandlerManifest.json
boumenot: @ahmetalpbalkan in that case I'll remove namespace, name, and extension_version has they are not part of the schema.  Any objections to moving them to manifest.xml?  They have to be there anyway.

ahmetalpbalkan: > Any objections to moving them to manifest.xml? They have to be there anyway.

They are already in manifest.xml as you said, so it would be better to not to have redundancy here. You can refer to the pdf I pointed out earlier for the precise schema.

boumenot:
@ahmetalpbalkan, any more comments?

ahmetalpbalkan:
All good for now. In the future we probably should update `PrivacyUrl` to http://www.microsoft.com/privacystatement/en-us/OnlineServices/Default.aspx

boumenot:
I'll make the change the now, what's one more. :grinning:

PR #168, opened 2016-06-14
Simple Makefile to build extensions.

boumenot:
Use the version and type (name) from an extension's manifest.xml to
create an extension's zip archive.

Support building with directories that contain whitespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,22 @@
+default: clean build
+
+EXTENSIONS = \
+ 	CustomScript \
+ 	DSC \
+ 	OSPatching \
+	VMBackup
+
+clean:
+	rm -f build/*.zip
+
+build: $(EXTENSIONS)
+
+$(EXTENSIONS):
+	$(eval NAME    = $(shell grep -Pom1 "(?<=<Type>)[^<]+" $@/manifest.xml))
+	$(eval VERSION = $(shell grep -Pom1 "(?<=<Version>)[^<]+" $@/manifest.xml))
+
+	@echo "Building '$(NAME)-$(VERSION).zip' ..."
+	@cd $@ && find . -type f | grep -v "/test/" | grep -v "./references" | zip -9 -@ ../build/$(NAME)-$(VERSION).zip > /dev/null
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
while you have the chance, let's switch to a little bit more strict bash mode http://redsymbol.net/articles/unofficial-bash-strict-mode/

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+clean:
+	@rm -f build/*.zip
+
+build: $(EXTENSIONS)
+
+$(EXTENSIONS):
+	$(eval NAME    = $(shell grep -Pom1 "(?<=<Type>)[^<]+" $@/manifest.xml))
+	$(eval VERSION = $(shell grep -Pom1 "(?<=<Version>)[^<]+" $@/manifest.xml))
+	@sh script/create_zip.sh $@ $(NAME) $(VERSION)
```
ahmetalpbalkan: hm you shouldn't need `sh`?

script/create_zip.sh
```diff
@@ -37,7 +37,7 @@ ext_name=$2
 ext_version=$3
 cd $ext_dir
 ext_dir=`pwd`
-cd $curr_dir
+cd "$curr_dir"
```
ahmetalpbalkan: I find this `pwd` thing pretty sloppy. we can use relative paths to the build script by using `readonly SCRIPT_DIR=$(dirname $0)` and find our way from there

Makefile
boumenot: Because the file isn't executable.  I can make that change instead.

boumenot:
Slight change of plans.  There is considerably more work here than I plan on doing for this iteration.  The goal is to get a build that produces a deployable version of CustomScript.  To that end, I reverted create_zip.sh work because it is unnecessary.  I have changed the Makefile to do the work necessary to create a CustomScript deployable artifact.

There's more debt to pay off, but I'm deferring that for another day.

ahmetalpbalkan:
gotta invite someone with makefile literacy here

boumenot:
@jasonzio - would you please take a look?

jasonzio:
lgtm

PR #171, opened 2016-06-15
Ensure correct extension version is logged

### LINKED ISSUE #143: Wrong version reflected in the logs

boumenot:
Closes #143 

Consolidate more of the various copies of Utils to the one true global copy.  The major extensions have now all been moved over.  Merge the differences between Diagnostic's version and the global version.

When HandlerUtil is constructed attempt to determine the extension's long name, short name, and verison by parsing the manifest or by examining the current working directory.

Remove any calls to determine the name by checking HandlerManifest.json.  Name is not a recognized schema element.

Remove any calls to determine the version by checking HandlerManifest.json.  Version is a recognized element, but it determines the schema version, not the extension version.


Makefile
```diff
@@ -1,9 +1,10 @@
 default: clean build
 
 EXTENSIONS = \
- 	CustomScript \
- 	DSC \
- 	OSPatching \
```
ahmetalpbalkan: is this diff intentional?

Utils/HandlerUtil.py
```diff
@@ -61,26 +61,62 @@
...
+
 class HandlerContext:
     def __init__(self,name):
         self._name = name
         self._version = '0.0'
         return
 
 class HandlerUtility:
-    def __init__(self, log, error, short_name):
+    def __init__(self, log, error, s_name=None, l_name=None, version=None):
```
ahmetalpbalkan: naming nit: `version` is stored in `_extension_version`. it would be good to name this `extension_version` instead. not important.

Utils/HandlerUtil.py
```diff
@@ -61,26 +61,62 @@
...
+            self._log_prefix = '[%s-%s] ' % (l_name, version)
+
+    def get_extension_version(self):
+        return self._extension_version
+    
     def _get_log_prefix(self):
-        return '[%s-%s]' %(self._context._name, self._context._version)
+        return self._log_prefix if self._log_prefix else '[%s-%s] ' % (self._context._name, self._context._version)
+
+    def _get_extension_info(self):
```
ahmetalpbalkan: is this used?

AzureEnhancedMonitor/ext/handler.py
```diff
@@ -30,6 +30,8 @@
 import Utils.HandlerUtil as util
 
 ExtensionShortName = 'AzureEnhancedMonitor'
+ExtensionFullName  = 'Microsoft.OSTCExtensions.AzureEnhancedMonitor'
+ExtensionVersion   = 'AzureEnhancedMonitor'
```
ahmetalpbalkan: is this value by design?

VMAccess/vmaccess.py
```diff
@@ -105,19 +105,19 @@ def enable():
...
 def disable():
-    hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName)
+    hutil = Util.HandlerUtility(waagent.Log, waagent.Error)
     hutil.do_parse_context('Disable')
     hutil.do_exit(0, 'Disable', 'success', '0', 'Disable Succeeded')
 
 
 def update():
-    hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName)
+    hutil = Util.HandlerUtility(waagent.Log, waagent.Error)
```
ahmetalpbalkan: are we actually going to pass `version` to these?

Utils/WAAgentUtil.py
```diff
@@ -67,10 +68,49 @@ class _WALAEventOperation:
         Update = "Update"           
     waagent.WALAEventOperation = _WALAEventOperation
 
-__ExtensionName__=None
+
+# begin: waagent http proxy config stuff
```
ahmetalpbalkan: unnecessary comment?

Utils/HandlerUtil.py
```diff
@@ -61,26 +61,62 @@
...
+    def __init__(self, log, error, s_name=None, l_name=None, version=None):
         self._log = log
         self._error = error
-        self._short_name = short_name
 
+        if s_name is None or l_name is None or version is None:
+            (long_name, short_name, ext_version) = self._get_extension_info()
+            self._short_name = short_name
+            self._extension_version = ext_version
+            self._log_prefix = '[%s-%s] ' % (long_name, ext_version)
```
ahmetalpbalkan: currently implementation has no version passed so will it have trailing dash, such as:

    [Microsoft.Azure.Extensions.CustomScriptExtension-] LOG MESSAGE

Also, I personally prefer seeing short name here. I doubt we'll ever have namespace collision here.

Utils/HandlerUtil.py
```diff
@@ -61,26 +61,62 @@
...
+        else:
+            self._short_name = s_name
+            self._extension_version = version
+            self._log_prefix = '[%s-%s] ' % (l_name, version)
```
ahmetalpbalkan: `self._log_prefix = ` stmts can be moved outside `if`

Utils/HandlerUtil.py
```diff
@@ -61,26 +61,62 @@
...
+
+    def get_extension_version(self):
+        return self._extension_version
+    
     def _get_log_prefix(self):
-        return '[%s-%s]' %(self._context._name, self._context._version)
+        return self._log_prefix if self._log_prefix else '[%s-%s] ' % (self._context._name, self._context._version)
```
ahmetalpbalkan: I miss point of this one. is there way of skipping the constructor and having `self._log_prefix == None`?

Utils/HandlerUtil.py
```diff
@@ -61,26 +61,62 @@
...
+
+    def _get_extension_info(self):
+        if os.path.isfile(MANIFEST_XML):
+            return self._get_extension_info_manifest()
+
+        ext_dir = os.path.basename(os.getcwd())
+        (long_name, version) = ext_dir.split('-')
```
ahmetalpbalkan: Nice trick. However I'm not sure this can be relied upon. I don't think it is covered by the spec as all these `settings/` `status/` paths of the extension are provided to us via `HandlerEnvironment.xml` and extensions are supposed to read it from there. Not sure if dirname can have this format forever.

Makefile
boumenot: EMACS flagged my spacing in the Makefile, and I **always** listen to EMACS. :grinning:

Utils/HandlerUtil.py
boumenot: I'm not very happy with my solution.  I would prefer that the code just know the correct values for long name, short name, and version, but I'm allowing it to be overridden.  (Actually, I would prefer a lot of things about the code, but ...) The user of of *... is None or ... is None or ... is None* is goofy.  The only caller of this is a test that I question the value of.  I'm thinking about making the parameters list just (self, log, error), and dropping the rest.

Thoughts?

Utils/HandlerUtil.py
boumenot: Yes.  Extensions were hard coding the their version number in the code, which made yet another location where the version to be maintained.  This is my **first** attempt at consolidating that, by making the source of the extension version the manifest.xml file.  (There's more work to do here too.)

I also found cases were people were using the version field in the HandlerManifest.json for the extension version.  You corrected me on this issue, so it's fresh in my mind.  The version in HandlerManifest.json is for the schema version of the file.

VMAccess/vmaccess.py
boumenot: No.  The intent is that Util.HandlerUtility will determine the version by examining the environment.  It will read the manifest.xml to determine the version, or the directory structure.

Utils/WAAgentUtil.py
boumenot: Left over from the merging of Diagnostics/Utils and Utils/.

Utils/HandlerUtil.py
boumenot: How so?  ext_version is set on line 86.

This is the way that @hosungsmsft formatted the logs, so maybe he has the motivation as to why?

Utils/HandlerUtil.py
boumenot: I fail to see how.

Utils/HandlerUtil.py
boumenot: The file is manifest.xml not HandlerEnvironment.xml, which is what I committed previously.  I'm asserting that every extension **we** ship from this repository will have manifest.xml included.

The directory structure is one that I'm less sure about, but there are existing code comments that state this is defacto.  It's our agent, so I assume the convention is what we own.  I'm not even sure I need the fallback case because we control all everything in the first place.

Utils/HandlerUtil.py
boumenot: Good point!

Utils/HandlerUtil.py
hosungsmsft: I've had no other motivation than fixing the wrong extension version in logs (1.0) to the correct one (e.g., 2.3.9001), and retaining all other existing behaviors. The long name has been used for LAD for so long, and current LAD E2E test suite depends on the long name, which can be changed, of course, so I'm OK with changing it to short name.

Utils/HandlerUtil.py
ahmetalpbalkan:
```
if foo:
    l = ...
    v = ...
else:
   l = ...
   v = ...
self._log_prefix = '[%s-%s] ' % (l,v)
```

VMAccess/vmaccess.py
ahmetalpbalkan: @boumenot then it makes sense to remove the `version` argument, but you need to be careful the path & name guarantees (see my comment pertinent to this).

Utils/HandlerUtil.py
hosungsmsft: I don't quite understand the original comment. Again the motivation for this is to retain the existing behavior as is, when the desired extension name/version are not specified at the HandlerUtility construction time, as I wouldn't be too comfortable to change the existing behavior for unknown cases.

ahmetalpbalkan: @hosungsmsft `self._log_prefix` will never be empty as it is set in all code paths in the ctor

Utils/HandlerUtil.py
ahmetalpbalkan: @boumenot no I'm saying, I dont see any code invoking this method here?

Utils/HandlerUtil.py
ahmetalpbalkan: @boumenot I'll leave the risk assessment to you. The hack seems nice enough. You can talk to the AzureRT folks and make it even a part of the spec.

Utils/HandlerUtil.py
hosungsmsft: Looks like it's invoked from line 86?

Utils/HandlerUtil.py
hosungsmsft: That's true now, but wasn't earlier. If the new code change guarantees self._log_prefix to be non-None all the time, we can change this code accordingly as well. I still don't understand "skipping the constructor and having self._log_prefix == None".

Again, the motivation for this original workaround was to retain the existing behavior as much as possible. I'm not saying the existing behavior is correct. The thing with the original implementation is that the self._context can change anytime, resulting in the varying extension name/ver in logs. And I didn't want to compose the string everytime, when it's fixed with the explicitly passed extension name and version at the construction time.

Utils/HandlerUtil.py
jasonzio: If changing the name would break your tests, then it'll break other tests as well. Do we really want to do that?

ahmetalpbalkan: Let's not change it then.

hosungsmsft: That's the whole point of my intention to try to retain the existing behavior as much as possible. I simply don't know how all other extensions were using (or not using) the existing log format, and I just wanted to make sure that this fix becomes available only if each extension owner explicitly specifies the desired extension name & version.

Utils/HandlerUtil.py
jasonzio:
```
if s_name is None or l_name is None or version is None:  
    (l_name, s_name, version) = self._get_extension_info()  
self._short_name = s_name  
self._extension_version = version  
self._log_prefix = '[%s-%s] ' % (l_name, version) 
```

Utils/HandlerUtil.py
boumenot: This is a simple decision, long name it is.

As an aside, I do not want to live in fear of the code.  We need to make changes to improve its quality, and we are going to break things.  Continued abuses of if *this* elif *those* elif *his* is a mess, and at some point we need to deal with the associated pains to have a better outcome.  We control the universe of extensions in this repository, so it is bound and known.

Utils/HandlerUtil.py
boumenot: I think we are in agreement.  The check is needless, and can be removed.

Utils/HandlerUtil.py
boumenot: I saw a solution that involved using the constructor's parameters, but this makes me feel dirty. :stuck_out_tongue_winking_eye:

boumenot:
I think this addresses all of the comments.

I am taking on the risk the @ahmetalpbalkan mentions concerning environment and path values.

PR #172, opened 2016-06-16
Ensure build/ exists.

boumenot:


Makefile
```diff
@@ -7,7 +7,10 @@ EXTENSIONS = \
 	VMBackup
 
 clean:
-	rm -f build/*.zip
+	rm -rf build
+
+init:
+	@mkdir -p build
```
jasonzio: Why not just make this the first step in the "build" recipe? I don't see the need for a separate target.

boumenot: Why? Because I got too fancy. 😀

boumenot:
@jasonzio - sorry I deleted your comment when I meant to delete mine.  Per your feedback I made init part of the build dependency list.

PR #173, opened 2016-06-17
Remove WAAgent 2.0.14 as it has a security issue and is replaced with…

mbearup:
Remove waagent 2.014 due to a security issue, replaced with 2.0.16

Fix a security issue where mysql password is logged by diagnostic extension


ahmetalpbalkan:
LGTM

Diagnostic/diagnostic.py
```diff
@@ -890,7 +890,7 @@ def install_omi():
         mysqladdress=readPrivateConfig("mysqladdress")
         mysqlusername=readPrivateConfig("mysqlusername")
         mysqlpassword=readPrivateConfig("mysqlpassword")
-        RunGetOutput("/opt/microsoft/mysql-cimprov/bin/mycimprovauth default "+mysqladdress+" "+mysqlusername+" '"+mysqlpassword+"'")
+        RunGetOutput("/opt/microsoft/mysql-cimprov/bin/mycimprovauth default "+mysqladdress+" "+mysqlusername+" '"+mysqlpassword+"'", log_cmd=False)
```
hosungsmsft: Unfortunately, this doesn't look right on LAD, as LAD defined a new function LogRunGetOutput, and assign LogRunGetOutput to RunGetOutput.

It looks like we need to change the param name here to 'should_log', and also in LogRunGetOutput(), the actual waagent.RunGetOutput() call should be 'waagent.RunGetOutput(log_cmd=should_log)'... Please take a look at lines 63-72.

Thanks much for doing this!

hosungsmsft:
@mbearup - I just remembered this change on LAD logging issue, and realized that it's not still merged. So I kind of applied the LAD logging change myself directly on a commit I'm adding to next LAD release. Just FYI. Thx.

ahmetalpbalkan:
Should we close or this needs rebasing?

hosungsmsft:
I encountered an issue with this change (not logging MySQL password). It turns out that the log_cmd param to RunGetOutput is not available on waagent 2.0.14, and many old distros still use that (e.g., Ubuntu 12.04), so this change fails extension execution.

Should those old waagent 2.0.14 defaults be changed to 2.0.16? What's the status of waagent 2.0.14 on existing images? If we can't upgrade them to 2.0.16, I don't think this (the MySQL password logging removal change) should be merged. I'll come up with a different solution for not logging the credential.

PR #176, opened 2016-06-22
Merge LAD v.2.3.9003 changes to master for release

hosungsmsft:
@jasonzio @LiweiPeng - Please review.

ChangeLogs:

2016-06-21: LAD-2.3.9003
    - Monitoring agent (mdsd) updates for a memory issue fix, a signal handler fix,
      and a fix to avoid a spin loop under certain circumstances
    - doesn't count non-quick crashes (>30 mins) towards retry limit
    - OMI reconfiguration not to listen to port 1270
    - Use systemd on Ubuntu 16.04 as well
    - Validate mdsd XML config before starting mdsd, fails fast on invalid
      config (with success)
    - Small Python 2.6 bug fix (syslog.openlog())

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -42,7 +42,7 @@ import Utils.ApplicationInsightsUtil as AIUtil
 
 ExtensionShortName = 'LinuxAzureDiagnostic'
 ExtensionFullName = 'Microsoft.OSTCExtensions.LinuxDiagnostic'
-ExtensionVersion = '2.3.9002.2'   # Must be updated on each new release! Improve this!
+ExtensionVersion = '2.3.9003'   # Must be updated on each new release! Improve this!
 WorkDir = os.getcwd()
 MDSDPidFile = os.path.join(WorkDir, 'mdsd.pid')
 MDSDPidPortFile = os.path.join(WorkDir, 'mdsd.pidport')
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -402,9 +402,14 @@ def setEventVolume(mdsdCfg,ladCfg):
     XmlUtil.setXmlValue(mdsdCfg,"Management","eventVolume",eventVolume)
         
 def configSettings():
+    '''
+    Generates XML cfg file for mdsd, from JSON config settings (public & private).
+    Returns (True, '') if config was valid and proper xmlCfg.xml was generated.
+    Returns (False, '...') if config was invalid and the error message.
+    '''
     mdsdCfgstr = readPublicConfig('mdsdCfg')
-    if not mdsdCfgstr :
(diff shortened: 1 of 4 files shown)


Diagnostic/diagnostic.py
```diff
@@ -642,6 +649,19 @@ def start_mdsd():
             copy_env['MDSD_http_proxy'] = proxy_config
 
     xml_file = os.path.join(WorkDir, './xmlCfg.xml')
+
+    # We now validate the config and proceed only when it succeeds.
+    config_validate_cmd = '{0} -v -c {1}'.format(os.path.join(MdsdFolder, "mdsd"), xml_file)
+    config_validate_cmd_status, config_validate_cmd_msg = RunGetOutput(config_validate_cmd)
+    if config_validate_cmd_status is not 0:
+        # Invalid config. Log error and report success.
+        message = "Invalid mdsd config given. Can't enable. Still considered successful. Config validation result: "\
```
LiweiPeng: i am not sure what 'Still considered successful' mean from the error message. does this mean that the invalid configuration is OK although it failed? If so, that's misleading.

Diagnostic/diagnostic.py
```diff
@@ -879,13 +905,19 @@ def install_omi():
             hutil.error("OMI install failed " + str(maxTries) + " times. Giving up...")
             return 1, "OMI install failed " + str(maxTries) + " times"
 
+    shouldRestartOmi = False
+
+    # Check if OMI is configured to listen to 1270 and reconfigure if so.
+    omiListensTo1270 = RunGetOutput("grep '^httpsport=0,1270' /etc/opt/omi/conf/omiserver.conf")[0] is 0
```
LiweiPeng: there are several  issues here: if OMI change port 1270 to 1271, this will fail; .if OMI add extra space like 0, 1270. this will fail too.

One way to do is: 1) search httpsport line. 2) check whether it has only 0 or not, if not, make it to be port 0 only.

Diagnostic/diagnostic.py
hosungsmsft: It means the extension install/enable is considered successful (as it's a user error), not meaning that mdsd is OK with this invalid configuration. This invalid config issue shouldn't penalize the extension install/failure success rate. I'll update the comment so that it's clearer.

Diagnostic/diagnostic.py
hosungsmsft: Thanks. I was worried about those possibilities as well. Will strengthen the checking logic and push an update.

hosungsmsft:
Thanks much @LiweiPeng . Pushed another commit reflecting your feedback and some other minor changes.

jasonzio:
Consider refactoring the "is this crash quick or not" to a separate method, since you make that check twice; if you ever want to change the definition of "quick", your life will be easier.

Otherwise, lgtm.

LiweiPeng:
looks good to me.

PR #178, opened 2016-06-24
IDE is Probably Smarter than Me

boumenot:
There are numerous issues that were flagged by the IDE that I corrected.  There should **not** be any functional changes in this commit.  These are just issues flagged as bad practice, or legacy issues that will bite us in the future.  This commit may already be too big, but there were a lot of issues.

Examples of fixes include:

1. There was code that would fail at runtime because variables did not exist in the scope of the method.
1. Unused variables - deleted
1. Unused imports - deleted (still lots of cases)
1. try/except should be _except Exception as e:_ not _except Exception, e_ to ensure future compatibility.
1. Typos
1. Commented out code - check source control.
1. Print as a function
1. Useless parens, or parens with different semantic information that intended, e.g. ((self))
1. Formatting of octal numbers.

Every module had its own view of the world as to what version of Python it supports.  I saw 2.4+, 2.6+ and 2.7+.  There was common code that claimed versions later than the code that was consuming it.  This is bad. I made a statement in the README as to the supported version, and I latched onto 2.7+.  We should discuss.

Almost (every) file's copyright notice is out of date.  Why is the license included in every file?  I'll remove this on a future PR.


Diagnostic/diagnostic.py
```diff
@@ -9,36 +9,29 @@
...
-import datetime
-from Utils.WAAgentUtil import waagent
-import Utils.HandlerUtil as Util
-import commands
-import base64
+import traceback
 import xml.dom.minidom
 import xml.etree.ElementTree as ET
-from collections import defaultdict
+from six import string_types
```
hosungsmsft: I don't see this in the original code. Do we need this?

hosungsmsft:
LGTM regarding LAD only (didn't look other source files).

I wanted to be able to move on with changing LAD code after your correct-ext-ver-logging fix PR was merged, but I guess I have to wait until this PR is merged... I really don't enjoy resolving merge conflicts. Can we get this in soon?

boumenot:
@jasonzio or @ahmetalpbalkan  - thoughts?

ahmetalpbalkan:
looks good.

PR #181, opened 2016-06-30
[LAD] Merge 2.3.9005 release branch to master

hosungsmsft:
@LiweiPeng - Please review. Thanks.

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -487,7 +487,7 @@ def get_extension_operation_type(command):
     if re.match("^([-/]*)(disable)", command):
         return waagent.WALAEventOperation.Disable
     if re.match("^([-/]*)(uninstall)", command):
-        return waagent.WALAEventOperation.Uninstall
+        return waagent.WALAEventOperation.UnIsntall
     if re.match("^([-/]*)(update)", command):
         return waagent.WALAEventOperation.Update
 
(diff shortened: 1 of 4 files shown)


ahmetalpbalkan:
❤️  your commit messages. They're 💯 .

Diagnostic/diagnostic.py
```diff
@@ -25,36 +25,41 @@
...
+hutil.set_verbose_log(False)  # Explicitly set verbose flag to False. This is default anyway, but it will be made explicit and logged.
 public_settings = hutil.get_public_settings()
 private_settings = hutil.get_protected_settings()
 if not public_settings:
     public_settings = {}
 if not private_settings:
     private_settings = {}
-
+ExtensionOperationType = None
+# Better deal with the silly waagent typo, in anticipation of a proper fix of the typo later on waagent
```
ahmetalpbalkan: have you brought this up in the agent repo? If there's a fix we need to take, we can take it to the upcoming v2.1.5 release which we'll push to publishers soon.

Diagnostic/diagnostic.py
```diff
@@ -478,110 +483,125 @@ def install_service():
...
-            elif re.match("^([-/]*)(enable)", command):
+                RunGetOutput('systemctl enable mdsd-lde')
+                mdsd_lde_active = RunGetOutput('systemctl status mdsd-lde')[0] is 0
+                if not mdsd_lde_active or hutil.is_current_config_seq_greater_inused():
+                    RunGetOutput('systemctl restart mdsd-lde')
+            else:
                 # if daemon process not runs
-                hutil.log("get pid:"+str(get_mdsd_process()))
-                if len(get_mdsd_process())!=2 or  hutil.is_current_config_seq_greater_inused():
+                hutil.log("get pid:" + str(get_mdsd_process()))
```
LiweiPeng: minor: get_mdsd_process() is called twice here. it's better to assign it to a var, then use the var in the 2nd case.

LiweiPeng:
looks good to me. minor comment posted. one general issue that's unrelated to this PR: LAD uses 'tail()' to get the last 1024-byte message from mdsd.err when mdsd fails.  In some case, this 1024-byte may not be enough to show the actual root cause. probably LAD needs to be more smarter on how much mdsd.err to log.

Diagnostic/diagnostic.py
hosungsmsft: Good point. Changing...

PR #202, opened 2016-07-14
[LAD] Merge release branch to master for 2.3.9007 release

hosungsmsft:
@LiweiPeng - Please review.

### DIFF
--- a/Diagnostic/manifest.xml
+++ b/Diagnostic/manifest.xml
@@ -2,7 +2,7 @@
 <ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
   <ProviderNameSpace>Microsoft.Azure.Extensions</ProviderNameSpace>
   <Type>TestLAD</Type>
-  <Version>2.3.9006.1</Version>
+  <Version>2.3.9006.2</Version>
   <Label>Microsoft Azure Diagnostic Extension for Linux Virtual Machines</Label>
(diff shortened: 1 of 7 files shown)


Diagnostic/manifest.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version='1.0' encoding='utf-8' ?>
 <ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
-  <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
-  <Type>LinuxDiagnostic</Type>
-  <Version>2.3.9005</Version>
+  <ProviderNameSpace>Microsoft.Azure.Extensions</ProviderNameSpace>
+  <Type>TestLAD</Type>
+  <Version>2.3.9006.1</Version>
```
LiweiPeng: what's the reason for Type change and ProviderNameSpace change?

hosungsmsft: It's not really needed at all, but it just shows that I publish the same extension with a different name, so my E2E tests can be done totally independent from this original name. If I use the original name and publish a test extension version, that test version is still made available to customer VMs, so I can't use the same name for a test extension version.

LiweiPeng:
the change looks good to me. I left one comment for commit 1 but I found you changed it back in commit #3. so I signed off.

PR #208, opened 2016-07-26
VMAccess: Post correct status for Install

boumenot:
The install status was incorrectly posted as "Installed," which is not a
valid state.  The correct status is "success."

Add a change log to track changes to VMAccess over time.


ahmetalpbalkan:
@boumenot have you asked agent folks if .status is collected for install/uninstall operations? AFAIK it is not.

boumenot:
The entire reason the PR was created was due to Fabric stating the posted install status was incorrect.

ahmetalpbalkan:
so it's about telemetry rather than the .status file?

boumenot:
I had to manually test this feature, so I ran through the various scenarios outlined in the [docs][1].  I noticed that the CLI would not return when deploying 1.4.0.0.  I can't definitively say that was the issue, but I think this is more than telemetry.

[1]: https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-linux-using-vmaccess-extension/

ahmetalpbalkan:
@boumenot yep I think that's the waagent not understanding the string you posted https://github.com/Azure/WALinuxAgent/issues/323

ahmetalpbalkan:
LGTM

Issue #224, opened 2016-08-16
LAD creates duplicate OMIQuery elements in generated config when enabled via portal


jasonzio:
When the LAD configuration settings include specific performance counter configs, or if none are specified and the default (LinuxMemory, LinuxCpu, LinuxDisk tables) perf collections are applied, _and_ the config file generated to that point contains more than one OMI element, LAD inserts the generated OMIQuery elements into each occurrence of the OMI element. The result is LAD requests collection of data more than once, doubling the resource requirements for collection and increasing transmission and storage costs.

When the extension is enabled through the portal, all default configs are used; because of the way the config file is generated, there are two OMI elements when the perf counter settings are generated. If a custom config is specified, the bug bites only if the user inserts the ARM ResourceId into the config (in an attempt to enable portal graphing of collected data). This latter scenario is extremely uncommon, but pure default config via portal enablement is by far the most common scenario.

PR #225, opened 2016-08-16
Write generated perfcfg query elements to only one OMI parent element.

jasonzio:
Fixes [issue 224](https://github.com/Azure/azure-linux-extensions/issues/224). Also fixes an insignificant-to-OMI typo that offends my sense of order and rightness in the world.


LiweiPeng:
looks good to me.

PR #231, opened 2016-08-25
vmaccess: backup sshd_config and restart SSHD

boumenot:
If the user specifies a password, then VMAccess must ensure password
authentication will be accepted by SSHD.  In this case, the code failed
to backup sshd_config, and failed to restart the daemon to ensure the
config change took effect.

### DIFF
--- a/VMAccess/CHANGELOG.md
+++ b/VMAccess/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 1.4.2.0 (2016-08-25)
 - Ensure expiration (if specified) is used when creating an account
+- Backup sshd_config before any edits are made.
+- Ensure sshd_config is restarted when edits are made.
 
 ## 1.4.1.0 (2016-07-27)
 - Install operation posts incorrect status [#206]
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -34,7 +34,7 @@ ExtensionShortName = 'VMAccess'
 BeginCertificateTag = '-----BEGIN CERTIFICATE-----'
 EndCertificateTag = '-----END CERTIFICATE-----'
 OutputSplitter = ';'
-
+SshdConfigPath = '/etc/ssh/sshd_config'
 
 def main():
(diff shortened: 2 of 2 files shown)


jasonzio:
If reset_ssh is set in the private config, then the "backup" and "restart SSH" operations will be performed on the way out of the enable operation. Instead of performing the backup and restart operations in-place, consider setting backupSSHConfigRequired and restartSSHRequired flags; perform those actions, if requested, in _main on the way out.

jasonzio:
&lt;TeachGrannyToSuckEggs&gt;
```python
def _perform_deferred_actions():
    if RequireSSHBackup:
        _backup_sshd_config(RequiresSSHBackup)
    if RequireSSHRestart:
        waagent.MyDistro.restartSshService() 
```

Request backup of the SSH config file by setting the `RequireSSHBackup` global to the path; request a daemon restart by setting `RequireSSHRestart` to true. Initialize both to None; invoke `_perform_deferred_actions()` after the command loop in `main()`.

&lt;/TeachGrannyToSuckEggs&gt;

boumenot:
That's a great phrase!

The latest changes avoid the double restart, and better handle backup.

You cannot perform backup as a deferred action, you must perform it preemptively otherwise you will not backup the pristine state.

I liked the idea of making restart an action that can happen after all others, and then I started implementing it.  Referencing a global variable to set it looked off, so I localized the change to enable().

The code needs some work.  I would like to redo it first, and then revisit this suggestion.

1. There are not unit tests.
1. The code emits log statements that are inconsequential to operation executed.  Why does repair disk say it reset ssh?!?
1. Hard codes some assumptions that can probably be avoided.

jasonzio:
Ah. You're right about the backup. In fact, you want to do backup only once, the _first_ time it's requested; you want to restart SSH only after the last time it's requested. And I completely agree about not making the flags global variables. Might actually be best to hide the "already did a backup" flag inside _backup_sshd_config().

jasonzio:
Now that I've actually _read_ your code, I see what you've done. LGTM.

PR #233, opened 2016-08-30
Handle missing DeploymentId gracefully (don't add as IdentityComponent)

jasonzio:
If running in a non-hosted environment, the HostingEnvironmentConfig.xml file will be unavailable, which means the DeploymentId is unavailable (and almost certainly irrelevant). Rather than throwing an error, just skip adding the DeploymentId as an IdentityComponent in the mdsd config (i.e. don't bother adding it as an extra column present in every table row driven by portal configuration).


hosungsmsft:
LGTM

PR #239, opened 2016-09-06
vmaccess: forcefully reset CHAP to no

boumenot:


VMAccess/vmaccess.py
```diff
@@ -67,6 +67,10 @@ def install():
 def enable():
     hutil = Util.HandlerUtility(waagent.Log, waagent.Error)
     hutil.do_parse_context('Enable')
+
+    if hutil.get_extension_version() == "1.4.5.0":
+        _forcefully_reset_chap()
```
brendandixon: @boumenot Why limit to 1.4.5.0? Also, does this change increment the hot fix version?

VMAccess/vmaccess.py
```diff
@@ -127,6 +131,19 @@ def update():
     hutil.do_exit(0, 'Update', 'success', '0', 'Update Succeeded')
 
 
+# Previous versions of VMAccessForLinux set ChallengeResponseAuthentication to yes,
+# which they should not.  Version 1.4.5.0 forcefully resets the value back to 'no'.
```
brendandixon: @boumenot "forcibly" vs. "forcefully" :)

boumenot: :100:

VMAccess/vmaccess.py
boumenot: This should be a one time thing to correct our previous mistake.  I wanted to contain it to this hotfix, and then remove it later...good point though in that this does not help us if we re-publish hotfixes for the other available versions (1.0, 1.1, 1.2, and 1.3).  I'll remove it.

I always miss that I have to update manifest.xml and CHANGELOG.md.  Will do.

VMAccess/vmaccess.py
brendandixon: @boumenot Or even more simple: "Verison 1.4.5.0 reset the value back to 'no'." Adverbs are so unnecessary, most of the time...said the Grammar Maven.

boumenot: I know what to get you for Xmas. :grinning:

jasonzio:
Can you add a new config entry that would allow a user to declaratively state his/her intent to enable chap? Absence of such declarative intent would trigger "forcibly reset to NO" behavior.

lunarfs:
@jasonzio commented that an option to would be required to not "forcibly reset to NO" in the current absence of such an option you can still keep you configuration set to yes by adding a comment with no  after the config option like 
`ChallengeResponseAuthentication yes # azure VMAccessForLinux requires no in this line to keep the yes configuration`

PR #240, opened 2016-09-07
vmaccess: None check for password

boumenot:
Now that VMAccess accepts a empty password, and ignores it the old and
expected behavior of not passing it in the first place must still be
honored.


VMAccess/vmaccess.py
```diff
@@ -156,7 +156,7 @@ def _set_user_account_pub_key(protect_settings, hutil):
     if not user_pass and not cert_txt and not ovf_env.SshPublicKeys:
         raise Exception("No password or ssh_key is specified.")
 
-    if len(user_pass) == 0:
+    if user_pass and len(user_pass) == 0:
```
brendandixon: @boumenot I believe this is best stated:

`user_pass is not None and len(user_pass) == 0`

Good find.

boumenot:
Closes #241

Issue #242, opened 2016-09-11
Deprecated value being used in the syslog template


mabicca:
The last parameter in the config file ~ is deprecated and it should be replaced with the word stop, here is an example of that the config file should look like after the change:

$ModLoad omazuremdslegacy.so
$template fmt, "\"syslog\",%syslogfacility-text:::csv%,\"%syslogseverity%\",\"%timereported:::date-rfc3339%\",\"%fromhost-ip%\",#TOJSON#%rawmsg%"
$legacymdsconnections 1
$legacymdsport 29131

$ActionQueueType LinkedList
$ActionQueueDequeueBatchSize 100
$ActionQueueSize 10000
$ActionResumeRetryCount -1
$ActionQueueSaveOnShutdown on
$ActionQueueFileName lad_mdsd_queue
$ActionQueueDiscardSeverity 8
*.*;local6.none :omazuremds:;fmt

$template fmtfile,"\"%syslogtag%\",#TOJSON#%rawmsg%"
$legacymdsconnections 1
$legacymdsport 29131

$ActionQueueType LinkedList
$ActionQueueDequeueBatchSize 100
$ActionQueueSize 10000
$ActionResumeRetryCount -1
$ActionQueueSaveOnShutdown on
$ActionQueueFileName lad_mdsd_queue
$ActionQueueDiscardSeverity 8
local6.* :omazuremds:;fmtfile
& stop

hosungsmsft:
Thanks for letting us know about this. We'll look into this and respond soon.

PR #245, opened 2016-09-12
updating deprecated string in config

mabicca:
https://github.com/Azure/azure-linux-extensions/issues/242

~ has been deprecated and we should use "stop"

More info at:
http://www.rsyslog.com/doc/v8-stable/compatibility/v7compatibility.html

"omruleset and discard (~) action are deprecated
Both continue to work, but have been replaced by better alternatives.

The discard action (tilde character) has been replaced by the “stop” RainerScript directive. It is considered more intuitive and offers slightly better performance.

The omruleset module has been replaced by the “call” RainerScript directive. Call permits to execute a ruleset like a subroutine, and does so with much higher performance than omruleset did. Note that omruleset could be run off an async queue. This was more a side than a desired effect and is not supported by the call statement. If that effect was needed, it can simply be simulated by running the called rulesets actions asynchronously (what in any case is the right way to handle this).

Note that the deprecated modules emit warning messages when being used. They tell that the construct is deprecated and which statement is to be used as replacement. This does not affect operations: both modules are still fully operational and will not be removed in the v7 timeframe."


Diagnostic/rsyslog5/omazurelinuxmds.conf
```diff
@@ -25,4 +25,4 @@ $ActionQueueSaveOnShutdown on
 $ActionQueueFileName lad_mdsd_queue
 $ActionQueueDiscardSeverity 8
 local6.* :omazuremds:;fmtfile
-& ~
+& stop
```
hosungsmsft: @mabicca : Could you please double-check if this is supported on rsyslog v.5? If not, we shouldn't change this here.

mabicca: Hi @hosungsmsft yes, sorry, somehow that didn't get through properly, yes, the change should only be made for rsyslog7 and later. is there a quick way for you to fix that or you want me to do that?

hosungsmsft: Thanks for confirming @mabicca . Please allow me to apply this change only to ver. 7 & 8, to reduce your inconvenience. We are planning to release a newer version of LAD soon, so this change should be included there for sure, together with the queue file name change.

mabicca: Hi @hosungsmsft alright, yes, I wanted to see exactly what those config sections mean too , but I would probably change the filenames just to be safe ;) thanks for the updates.

hosungsmsft:
This change is applied to lad-2.3.9009 branch, which is now being merged to master, so closing this PR without taking it. Thanks.

PR #249, opened 2016-09-15
[LAD] Fix Events/OMI xml cfg regression on Python 2.6-based distros

hosungsmsft:
@jasonzio , @LiweiPeng - Please review.

### DIFF
--- a/Diagnostic/Utils/XmlUtil.py
+++ b/Diagnostic/Utils/XmlUtil.py
@@ -36,7 +36,7 @@ def addElement(xml,path,el,selector=[]):
 
 def addElementToFirstMatchingPathOnly(xml, path, el):
     element = xml.find(path)
-    if element:
+    if element is not None:
         element.append(el)
 
 
--- a/Diagnostic/Utils/XmlUtil.py
+++ b/Diagnostic/Utils/XmlUtil.py
@@ -27,7 +27,7 @@ def getXmlValue(xml,path,property):
         return element.get(property)
 
 
-def addElement(xml,path,el,addOnlyOnce=False,selector=[]):
+def addElement(xml,path,el,selector=[],addOnlyOnce=False):
     elements = xml.findall(path)
     for element in elements:
         if selector and element.get(selector[0])!=selector[1]:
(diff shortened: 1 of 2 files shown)


Diagnostic/Utils/XmlUtil.py
```diff
@@ -33,6 +33,14 @@ def addElement(xml,path,el,selector=[]):
             continue
         element.append(el)
 
+
+def addElementToFirstMatchingPathOnly(xml, path, el, selector=[]):
+    element = xml.find(path)
+    if selector and element.get(selector[0]) != selector[1]:
```
jasonzio: I would think "MatchingPath" in combination with a non-empty selector would imply "the element which matches the path and which satisfies the selector"; that is, if there's a non-empty selector, keep looking through the document until you find a path match that also satisfies the selector.

Diagnostic/Utils/XmlUtil.py
```diff
@@ -33,6 +33,14 @@ def addElement(xml,path,el,selector=[]):
...
+        return
+    element.append(el)
```
jasonzio: If xml.find(path) didn't find anything, what happens when you try to do the append?

Diagnostic/Utils/XmlUtil.py
hosungsmsft: Thanks. Will remove the selector completely.

Diagnostic/Utils/XmlUtil.py
hosungsmsft: Thanks. Will guard "element.append(el)" with "if element:".

LiweiPeng:
looks good to me.

Diagnostic/Utils/XmlUtil.py
```diff
@@ -26,18 +26,15 @@ def getXmlValue(xml,path,property):
     if element is not None:
         return element.get(property)
 
-def addElement(xml,path,el,selector=[]):
+
+def addElement(xml,path,el,addOnlyOnce=False,selector=[]):
```
jasonzio: Won't this break existing code which uses selector (e.g. Utils/ApplicationInsightsUtil.py)? I think you'd need to put addOnlyOnce at the end of the parameters to avoid breaking the signature,

hosungsmsft: Good point, though I deliberately did this after checking that there's currently no addElement() calls with selector. Moved addOnlyOnce=False to the end. Thanks.

PR #253, opened 2016-09-28
Creating Status Object and Adding telemetery

bpramod:
--Status Object created
-- telemetery added
-- Common Status JSON string is generated
-- Handled lsblk path


soumyanishan:
LGTM

PR #256, opened 2016-10-06
Added fstab watcher + unit tests + makefile

anandramakrishna:


jasonzio:
Looks like the MSBOT needs a little tweaking. You're an employee; no cla is required. :-)

Diagnostic/diagnostic.py
```diff
@@ -1166,6 +1175,51 @@ def get_deployment_id():
...
+            currentModDateTime = datetime.datetime.fromtimestamp(currentModTime)
+            
+            if (currentModTime != self.lastModTime and 
+                currentModDateTime < currentModDateTime + 
+                    datetime.timedelta(minutes=1)): 
+                tryMount = True
+                self.lastModTime = currentModTime
+
+        ret = 0
+        if (tryMount == True):
```
jasonzio: if tryMount:

Diagnostic/diagnostic.py
```diff
@@ -1166,6 +1175,51 @@ def get_deployment_id():
...
+        ch = logging.StreamHandler(errorStream)
+        ch.setLevel(logging.WARNING)
+        self.logger.addHandler(ch)
+        ch = logging.StreamHandler(outputStream)
+        ch.setLevel(logging.INFO)
+        self.logger.addHandler(ch)
+
+    def handle_fstab(self, ignoremtime=False):
+        tryMount = False
+        if ignoremtime == True:
```
jasonzio: if ignoretime:

Diagnostic/diagnostic.py
```diff
@@ -1166,6 +1175,51 @@ def get_deployment_id():
...
+            if (currentModTime != self.lastModTime and 
+                currentModDateTime < currentModDateTime + 
+                    datetime.timedelta(minutes=1)): 
+                tryMount = True
+                self.lastModTime = currentModTime
+
+        ret = 0
+        if (tryMount == True):
+            ret = subprocess.call(['sudo', 'mount', '-a', '-vf'])
+            if (ret != 0):
```
jasonzio: Python style doesn't wrap the test in parentheses. Just

if ret != 0:

Diagnostic/diagnostic.py
```diff
@@ -1166,6 +1175,51 @@ def get_deployment_id():
 
     return identity
 
+class Watcher():
```
jasonzio: I suspect we'll want to add more things-to-be-watched over time, which means this will grow. Might make sense to start off right with Watcher and the fstab_handler being in a separate module from the rest of LAD; might make unit testing easier as well.

Diagnostic/diagnostic_ut.py
```diff
@@ -0,0 +1,74 @@
+import unittest
```
jasonzio: Rather than have the unit tests and the test driver living in the same directory as the source, seems it would be better to create a "tests" directory and put them there. As LAD is improved and changed over the coming months, more tests will be written; better to have a clean home for them.

Diagnostic/diagnostic.py
```diff
@@ -1166,6 +1175,51 @@ def get_deployment_id():
...
+    def handle_fstab(self, ignoremtime=False):
+        tryMount = False
+        if ignoremtime == True:
+            tryMount = True
+        else:
+            currentModTime = os.path.getmtime('/etc/fstab')
+            currentModDateTime = datetime.datetime.fromtimestamp(currentModTime)
+            
+            if (currentModTime != self.lastModTime and 
+                currentModDateTime < currentModDateTime +
```
jasonzio: I'm clearly missing something- what are you trying to do here?

Diagnostic/diagnostic.py
anandramakrishna: done.

Diagnostic/diagnostic.py
anandramakrishna: added a comment.  I'm trying to only check with mount after at least 1 minute has elapsed since the modification

Diagnostic/diagnostic.py
anandramakrishna: ok

Diagnostic/diagnostic.py
anandramakrishna: ok

Diagnostic/diagnostic_ut.py
anandramakrishna: ok.  done.

Diagnostic/watcherutil.py
```diff
@@ -0,0 +1,67 @@
...
+        if ignoretime:
+            tryMount = True
+        else:
+            currentModTime = os.path.getmtime('/etc/fstab')
+            currentModDateTime = datetime.datetime.fromtimestamp(currentModTime)
+
+            # Only try to mount if it's been at least 1 minute since the 
+            # change to fstab was done, to prevent spewing out erroneous spew
+            if (currentModTime != self.lastModTime and
+                currentModDateTime < currentModDateTime +
```
jasonzio: I do not think this does what you think it does. :-)

anandramakrishna: fixed.  Pls review.

Utils/HandlerUtil.py
```diff
@@ -73,6 +73,15 @@ class HandlerContext:
...
         self._version = '0.0'
+        self._config_dir = None
+        self._log_dir = None
+        self._log_file = None
+        self._status_dir = None
+        self._heartbeat_file = None
+        self._seq_no = -1
+        self._status_file = None
+        self._settings_file = None
+        self._config = None
```
hosungsmsft: PR looks good to me. Thanks @jasonzio for your quick review.

@anandramakrishna , just wondering if these newly added members will be used later?

Diagnostic/diagnostic.py
```diff
@@ -738,6 +741,13 @@ def start_mdsd():
         eventhub_persist_dir_path).split(" ")
 
     try:
+        # Create monitor object that encapsulates monitoring activities
+        watcher = watcherutil.Watcher(hutil._err, hutil._log)
```
hosungsmsft: Sorry, just saw this... I don't see hutil._err, but see only hutil._error. I wonder if you meant hutil._error here? FYI, pasting HandlerUtility __init__() here:

class HandlerUtility:
    def __init__(self, log, error, s_name=None, l_name=None, extension_version=None):
        self._log = log
        self._error = error

Below is how hutil is instantiated. Not sure if it's to the VM console, though...

waagent.LoggerInit('/var/log/waagent.log','/dev/stdout')
waagent.Log("LinuxAzureDiagnostic started to handle.")
hutil = Util.HandlerUtility(waagent.Log, waagent.Error)

hosungsmsft:
@anandramakrishna : Thanks for your fix of the typo (hutil._err to hutil._error), but I'm still not sure if that's going to achieve what you wanted. Didn't you say that you wanted to log those messages to the serial console? (Sorry if I'm mistaken and just ignore this in that case). As far as I can see, the hutil._error is initialized with '/dev/stdout' and I don't think that's /dev/console or /dev/ttyS0 or whatever that should be for the serial console. Maybe I'm missing something and we should sync up later when you get back. Here's the related code snippet from diagnostic.py:

waagent.LoggerInit('/var/log/waagent.log','/dev/stdout')
waagent.Log("LinuxAzureDiagnostic started to handle.")
hutil = Util.HandlerUtility(waagent.Log, waagent.Error)

anandramakrishna:
@hosungsmsft Good catch, it's only waagent that logs by default to /dev/console, I had been looking at that code and gotten confused.  I made a tweak to explicitly add a parameter to log to console optionally.

Diagnostic/watcherutil.py
```diff
@@ -19,8 +19,9 @@
 class Watcher():
     lastModTime = os.path.getmtime('/etc/fstab')
     logger = None
+    consoleStream = None
```
hosungsmsft: I don't see this variable used anywhere else. Did you mean to use this var for something?

hosungsmsft: I don't see this variable used anywhere else. Did you mean to use this somewhere? If not, should we remove this?

hosungsmsft: I don't see this variable used anywhere else. Did you mean to use this somewhere? If not, should we remove this?

Diagnostic/watcherutil.py
```diff
@@ -19,8 +19,9 @@
 class Watcher():
     lastModTime = os.path.getmtime('/etc/fstab')
```
hosungsmsft: This caught my eyes just now (sorry for being slow), but I wonder if you wanted this to be an instance variable, not a class variable...? As far as I understand, declaring a variable inside a class, but outside __init__() will make it a class variable, probably initialized only once at the script loading time. Is that what you wanted? Of course I do see that in handle_fstab() below, you do assign something to self.ladModTime, so it might work, but I'm not sure if this variable decl/def is right to be here. Same with the "logger" class variable below...

PR #257, opened 2016-10-07
LAD 2.3.9011 Release PR

hosungsmsft:
@jasonzio , @LiweiPeng : Please review. Thanks.


LiweiPeng:
looks good to me. sign off

hosungsmsft:
@LiweiPeng , @jasonzio : Please review again the latest 4 commits.

Sorry I had to push more commits. Turns out that even Unix domain socket requires reconfiguration of SELinux (on RH-based distros including CentOS and OracleLinux). And I wanted to avoid installing any unnecessary dependencies (policycoreutils-python), but because only RH ships all needed binaries (CentOS & OracleLinux miss the two of all three) and the way each distro reports its flavor, unfortunately I had to specify policycoreutils-python to be installed on all these three RH-based distros. And we have some different SELinux config commands, because we now need to enable for a socket file, not for a TCP port.

LiweiPeng:
looks good to me. signed off

hosungsmsft:
Thanks all @LiweiPeng & @jasonzio ! Finally E2E tests passed and we are releasing this!

PR #264, opened 2016-10-24
Storage Details Computation

bpramod:
Added more fields in status object
-unique machine Id
- total used space
- number of mounts 
- flags to assure the computation

### DIFF
--- a/VMBackup/main/Utils/HandlerUtil.py
+++ b/VMBackup/main/Utils/HandlerUtil.py
@@ -85,6 +85,8 @@ class HandlerUtility:
         self._error = error
         self._short_name = short_name
         self.patching = None
+        self.storageDetailsObj = None
+        self.partitioncount = 0
 
     def _get_log_prefix(self):
         return '[%s-%s]' % (self._context._name, self._context._version)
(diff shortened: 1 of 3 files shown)


VMBackup/main/handle.py
```diff
@@ -194,6 +194,7 @@ def daemon():
     #this is using the most recent file timestamp.
     hutil.do_parse_context('Executing')
     freezer = FsFreezer(patching= MyPatching, logger = backup_logger)
+    hutil.partitioncount = len(freezer.mounts.mounts)
```
soumyanishan: check for condition when freezer.mounts is None

soumyanishan: check for condition when freezer.mounts is None

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -244,10 +247,73 @@ def save_seq(self):
...
+            none                   4       0         4   0% /sys/fs/cgroup
+            none                5120       0      5120   0% /run/lock
+            none             1761552       0   1761552   0% /run/shm
+            none              102400       0    102400   0% /run/user
+            none                  64       0        64   0% /etc/network/interfaces.dynamic.d
+            tmpfs                  4       4         0 100% /etc/ruxitagentproc
+            /dev/sdb1        7092664   16120   6693216   1% /mnt
+
+            '''
+            output = df.communicate()[0]
```
soumyanishan: check if communicate method is required or the same can be done by using poll/wait followed by df.stdout

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -288,7 +288,7 @@ def get_total_used_size(self):
             /dev/sdb1        7092664   16120   6693216   1% /mnt
 
             '''
-            output = df.communicate()[0]
+            output = df.stdout.read()
```
soumyanishan: please use it with poll or wait

PR #269, opened 2016-10-30
Fix VMBackup crash on CentOS/RHEL 6.8

krkhan:


krkhan:
Should not be merged, should be in a separate fork as development has branched

PR #275, opened 2016-11-11
LAD 2.3.9015 release PR

hosungsmsft:
@jasonzio , @LiweiPeng : Please review. Thanks.

### DIFF
--- a/Diagnostic/README.md
+++ b/Diagnostic/README.md
@@ -1,7 +1,7 @@
 # Diagnostic Extension
 Allow the owner of the Azure Virtual Machines to obtain diagnostic data for a Linux virtual machine.
 
-Latest version is 2.3.9013.
+Latest version is 2.3.9015.
 
 You can read the User Guide below for detail:
 * [Use the Linux Diagnostic Extension to monitor the performance and diagnostic data of a Linux VM](https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-linux-diagnostic-extension/)
--- a/Diagnostic/README.md
+++ b/Diagnostic/README.md
@@ -59,7 +59,7 @@ Schema for the protected configuration file looks like this:
 
 
 * `storageAccountName`: (required) the name of storage account
-* `storageAccountKey`: (required) the access key of storage account
+* `storageAccountSasToken` or `storageAccountKey`: (required) a valid account SAS token for the storage acocunt or the access key of storage account. Only one of these two should be given. Note that `storageAccountKey` may be deprecated in some future. For the `storageAccountSasToken` requirements, please see below.
 * `mdsdHttpProxy`: (optional) http proxy configuration for mdsd. Format: "http://username:password@proxy_host:proxy_port". "http:" part is optional. You may specify username and password here. If this is specified both on public and protected configurations, this protected configuration will prevail.
 
 ```json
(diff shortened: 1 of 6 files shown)


LiweiPeng:
looks good to me. sign off

Diagnostic/README.md
```diff
@@ -70,6 +70,31 @@ Schema for the protected configuration file looks like this:
...
+}
+```
+
+An account SAS token should be of the following format:
+
+```
+sv=2015-12-11&ss=bt&srt=co&sp=rwlacu&st=2016-11-09T00%3A04%3A00Z&se=9999-11-10T00%3A04%3A00Z&sig=[signed-signature-string]
+```
+
+Details on how to construct an account SAS token can be found in [this document](https://msdn.microsoft.com/en-us/library/azure/mt584140.aspx). If you need to manually generate an account SAS token, we recommend using [the Microsoft Azure Storage Explorer](http://storageexplorer.com/). Just install/run the Explorer, login with your Azure account (or add a storage account explicitly), right-click on the storage account for which you want to generate an account SAS token, and click Get Shared Access Signature and follow the dialog.
```
jasonzio: This line is far too long.

Diagnostic/README.md
```diff
@@ -70,6 +70,31 @@ Schema for the protected configuration file looks like this:
...
+
+In order for the given account SAS to work with the Linux Azure Diagnostic extension, the following requirements must be met:
+
+* Services (ss) must include Blobs (b) and Tables (t).
+* Permissions (sp) must include Read (r), Write (w), List (l), Add (a), Create (c), and Update (u).
+* Resource Types (srt) must include Container (c) and Object (o).
+* Start time (st) and Expiry time (se) should be valid for the duration of time the SAS will be used. We recommend to set the Start time with today's date and to set the Expiry time with the date 9999-12-31.
```
jasonzio: Line too long.

Diagnostic/Utils/XmlUtil.py
```diff
@@ -40,3 +42,10 @@ def addElement(xml,path,el,selector=[],addOnlyOnce=False):
 def createElement(schema):
     return ET.fromstring(schema)
 
+
+def removeElement(tree, parent_path, removed_element_name):
+    parents = tree.findall(parent_path)
+    for parent in parents:
+        removed_elements = parent.findall(removed_element_name)
+        for element in removed_elements:
+            parent.remove(element)
```
jasonzio: Is this the right pattern? Does removing an element reshape the tree such that previously-found elements cannot simply be handed to parent.remove()? Or do you need something that looks more like this:
```
for parent in parents:
    element = parent.find(removed_element_name)
    while element is not null:
        parent.remove(element)
        element = parent.find(removed_element_name)
```

Diagnostic/README.md
hosungsmsft: Fixed in the next commit.

Diagnostic/README.md
hosungsmsft: Fixed in the next commit.

Diagnostic/Utils/XmlUtil.py
hosungsmsft: I'm not sure if the two are different, but your suggested pattern seems correct too. I'll try that. After all, our use case has no multiple elements to be removed in the first place.

PR #277, opened 2016-11-17
fix the issue of failing to find sudoers.d/waagent on FreeBSD

ccic:
The phenomenon of this bug is: On FreeBSD, user cannot add a new user on portal through "reset password". If you specify a user name which is not existed, and specify the new password on the portal. The result is your existing user on VM cannot execute "sudo" because it was removed from sudoers.d/waagent.

The root cause is found:
On FreeBSD, the sudo was installed in /usr/local/, and the corresponding sudoers.d locates in /usr/local/etc/. So, the current _get_other_sudoers of vmaccess.py returns "None", but the corresponding waagent for FreeBSD can find the correct sudoers.d/waagent, and it sets a new user into that file, but vmaccess.py did not save the previous content. Thus, trigger this bug.

The patch fixes this issue.

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -35,6 +35,10 @@ BeginCertificateTag = '-----BEGIN CERTIFICATE-----'
 EndCertificateTag = '-----END CERTIFICATE-----'
 OutputSplitter = ';'
 SshdConfigPath = '/etc/ssh/sshd_config'
+SudoersBaseDir = '/etc/sudoers.d'
+if not os.path.isdir(SudoersBaseDir):
+    ## sudoers.d folder on FreeBSD
+    SudoersBaseDir = '/usr/local/etc/sudoers.d'
 
 def main():
     waagent.LoggerInit('/var/log/waagent.log', '/dev/stdout')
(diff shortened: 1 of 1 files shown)


ccic:
Any comments about this fix? The patch has been pending for more than 1.5 month.

boumenot:
This change takes a try/fail approach as opposed to validating the available implementation based on the OS.  I will not accept such an approach.

This change is considerably risky, and affects every extension that incorporates the WAAgent code, which is most of them.

This change needs to re-validated across all flavors of Linux that we support, and I see no evidence of that in this PR.  The extension has no test suite, so it is very costly for me to validate this behavior. ☹️

The README claims to support FreeBSD, but if this patch is needed to support FreeBSD it seems dubious to me that FreeBSD was ever truly supported.  Perhaps you have context on this, but the existing code is before my time.

Sorry, I will not accept the patch at this time.

ccic:
What is the reason for refusing this patch?

boumenot:
In general, it is the same reason as #278.  It seems like FreeBSD support was claimed by the earlier code authors, but did not really work, and lack of automated testing makes an additional platform, FreeBSD, a burden that I do not want to take on at this time.

PR #278, opened 2016-11-17
fix the expiration format issue on FreeBSD

ccic:
The "expiration" parameter used to create a user does not work on FreeBSD, because "pw" command used on FreeBSD requires a different date format compared with Linux. FreeBSD wants "DD-MM-YYYY", but Linux requires "YYYY-MM-DD". So, a conversion is required. This patch targets to fix it.

### DIFF
--- a/Common/WALinuxAgent-2.0.14/waagent
+++ b/Common/WALinuxAgent-2.0.14/waagent
@@ -1870,7 +1870,22 @@ class FreeBSDDistro(AbstractDistro):
         if userentry == None:
             command = "pw useradd " + user + " -m"
             if expiration != None:
-                command += " -e " + expiration.split('.')[0]
+                expdate = expiration.split('.')[0]
+                ddmmyyyy = expdate
+                ## date format check:
+                ## pw requires the format should be "dd-mm-yyyy",
+                ## but the parameter here is "yyyy-mm-dd" in order to align with Linux
+                try:
+                    yyyymmdd = datetime.datetime.strptime(expdate, '%Y-%m-%d')
+                    ddmmyyyy = yyyymmdd.strftime('%d-%m-%Y')
+                except:
+                    try:
+                        ddmmyyyydate = datetime.datetime.strptime(expdate, '%d-%m-%Y')
+                        ddmmyyyy = ddmmyyydate.strftime('%d-%m-%Y')
+                    except:
+                        Error("Valid expiration date fmt is YYYY-MM-DD or DD-MM-YYYY, but the input is '{0}'".expdate)
+                        return "Failed to create account with invalid expiration date fmt"
+                command += " -e " + ddmmyyyy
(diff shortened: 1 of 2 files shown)


boumenot:
This looks locale specific.  Is there a way to influence the value so it does not have to be locale specific, and can use the more agnostic YYYY-MM-DD format.

ccic:
No. FreeBSD pw cmd does not accept "YYYY-MM-DD" format. See its help. Currently this bug caused Azure users to fail to reset password. I suggest fixing it ASAP before we find a decent solution.

> Format of the date is either a UNIX time in decimal, or a date in `dd-mmm-yy[yy]' format, where dd is the day, mmm is the month, either in numeric or alphabetic format ('Jan', 'Feb', etc) and year is either a two or four digit year.

boumenot:
This change takes a try/fail approach as opposed to validating the available implementation based on the OS.  I will not accept such an approach.

This change is considerably risky, and affects every extension that incorporates the WAAgent code, which is most of them.

This change needs to re-validated across all flavors of Linux that we support, and I see no evidence of that in this PR.  The extension has no test suite, so it is very costly for me to validate this behavior. ☹️

The README claims to support FreeBSD, but if this patch is needed to support FreeBSD it seems dubious to me that FreeBSD was ever truly supported.  Perhaps you have context on this, but the existing code is before my time.

Sorry, I will not accept the patch at this time.

ccic:
Here I want to transform the date format from 'YYYY-MM-DD' to 'DD-MM-YYYY'. Do you expect to change 'pw' to accept 'YYYY-MM-DD'? 
I don't think it impacts Linux Distros for this change only touches 'FreeBSDDistro'. So, there is no risky.

boumenot:
No, I do not expect pw to accept a bad format, but I do expect the code to present the correct format instead trying, failing, and then trying again.  It should **know** the correct format not guess at it, right?

There is always risk.  The question is how much. 😃 Do you have any test collateral to share for FreeBSD?

I followed up with team on this to solicit their ideas and feedback.  We decided that we'll take the same approach with VMAccessForLinux that we did with CustomScriptForLinux.  

 * The extension will be re-written in Go with appropriate unit and integration tests. (See [custom-script-for-linux](https://github.com/Azure/custom-script-extension-linux/) for examples of what I mean.)
 * The extension should support both Linux and FreeBSD.

I do not have an ETA to share on when the re-write will happen.

ccic:
I agree that "not guess at it". Here it is just to validate the input format. So, you suggest pending this issue until starting to re-write with Go?

boumenot:
I do.

If you're ambitious the working template for the new extension is the v2 of the [CustomScript](https://github.com/Azure/custom-script-extension-linux) extension.  The majority of the ideas, and most of the code will be carried over.  I am happy to start the project now if you would like to contribute changes.

PR #302, opened 2016-12-25
Update handle.py - sdx path replaced with uuid

mignonsys:
Addressed an issue - uuid of newly attached disks will be assigned ONLY after luksFormat, yet we still want uuid to be written into the /var/lib/azure_crypt_mount file as dev path, coz the sdx path for attached disks CAN change after reboot.

We found the issue appeared on several of our VMs running CentOS 7.2.1511 with multiple attached data disks. The VM encryption extension mounted incorrect disks (though luksOpen throws no exception since all disks shared the same passphrase file per VM).

### DIFF
--- a/VMEncryption/main/Common.py
+++ b/VMEncryption/main/Common.py
@@ -19,8 +19,8 @@
 
 class CommonVariables:
     utils_path_name = 'Utils'
-    extension_name = 'AzureDiskEncryptionForLinuxTest'
-    extension_version = '0.1.0.999220'
+    extension_name = 'AzureDiskEncryptionForLinux'
+    extension_version = '0.1.0.999234'
     extension_type = extension_name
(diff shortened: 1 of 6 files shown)


mignonsys:
Additional info for my pull request:
We used the '**EnableEncryptionFormat**' option in our ARM template and specified scsi_number for the data disks (even though we found some changed comments from you stated that the EnableEncryptionFormat should only be used by raid disks, we'd checked the code, and found it could/should work with empty data disks without raid.. and it is really important for our use case: having multiple data disks attached to VM through ARM template, and have them automatically formatted and mapped via VM extensions.)

ejarvi:
Thanks for the contribution, this was helpful, currently considered resolved in other changes that have since been made, closing this PR for now.

PR #308, opened 2017-01-06
fix the issue of failing to find sudoers.d/waagent

ccic:
FreeBSD user on Azure cannot add a new user or reset password for new user, otherwise, the existing user will be moved out from sudoer list.

The fix does not impact Linux distros by adding check for FreeBSD.


ccic:
This patch removed an original file existing check, abandon this request and will create a new one.

PR #309, opened 2017-01-06
[VMAccess]fix the issue of failing to find sudoers.d/waagent

ccic:


ccic:
This patch fixed the issue of moving existing user out from sudoers list when adding a new user through portal.

boumenot:
This is a dupe of #277, which was closed because I am not willing to accept a patch for FreeBSD at this time based on the reasons I have outlined in #278.

The implementation of the patch is not the issue.  I would not close a PR if I wanted to change the implementation, I would close a PR if I will not accept a PR based on principle.

PR #310, opened 2017-01-06
fix the expiration format issue on FreeBSD

ccic:
change the date format from "YYYY-MM-DD" to "DD-MM-YYYY" for FreeBSD.


boumenot:
This is a dupe of #278, which was closed because I am not willing to accept a patch for FreeBSD at this time based on the reasons I have outlined in #278.

The implementation of the patch is not the issue.  I would not close a PR if I wanted to change the implementation, I would close a PR if I will not accept a PR based on principle.

PR #312, opened 2017-01-06
[LAD] Initial changes for 3.0

jasonzio:
This PR contains four sets of changes.

1. Bring existing code into compliance with PEP8: variable naming, spacing, parens, etc.
2. Move distro-specific code into a separate package (DistroSpecific). This module exposes general utility methods ("run this command") as well as task-specific methods ("find the current major version of rsyslogd").
3. Implement the Providers module and the Builtin provider. A provider is responsible for generating the necessary configurations to drive ingest of metrics. The Builtin provider uses OMI and rsyslogd today. An alternate provider might be one which enables arbitrary fluentd usage, or arbitrary capture and parsing of /proc or /sys files.
4. A small smattering of bug fixes.

**Merging this PR now will make only the bugfixes and PEP8 cleanup active.** Forthcoming PRs will integrate these new packages into the diagnostic.py program. Additional refactoring of that still-too-large module should be expected.


Diagnostic/Utils/LadDiagnosticUtil.py
```diff
@@ -85,47 +94,50 @@ def generateOMIQueries(omiQueryConfiguration):
...
+    return omi_queries
+
 
 # Get elements from DiagnosticsMonitorConfiguration in LadCfg based on element name
 def getDiagnosticsMonitorConfigurationElement(ladCfg, elementName):
-    if ladCfg:
-        if 'diagnosticMonitorConfiguration' in ladCfg:
-            if elementName in ladCfg['diagnosticMonitorConfiguration']:
-                return ladCfg['diagnosticMonitorConfiguration'][elementName]
+    if ladCfg and 'diagnosticMonitorConfiguration' in ladCfg:
```
hosungsmsft: Minor: Maybe we should start defining constants for all these strings. Also, just wanted to make sure, the only casing allowed here is this Pascal casing, right? There was an earlier case where the casing of this name was changed to Pascal casing, which caused an incident (see lines 398-401 of diagnostic.py), so just a reminder...

Diagnostic/Utils/LadDiagnosticUtil.py
```diff
@@ -85,47 +94,50 @@ def generateOMIQueries(omiQueryConfiguration):
...
-    if fileLogs:
-        if 'fileLogConfiguration' in fileLogs:
-            return fileLogs['fileLogConfiguration']
+    if fileLogs and 'fileLogConfiguration' in fileLogs:
+        return fileLogs['fileLogConfiguration']
+    return None
+
 
 # Generates OMI queries from LadCfg
 def generatePerformanceCounterConfigurationFromLadCfg(ladCfg):
```
hosungsmsft: As far as I understand, the functions here are not used/exercised in production (because Portal never sets "ladCfg" in the protected settings), so I think it'd be great if we have some unit tests on the functions here. By the way, I thought AI team left some test code for these features, but I couldn't find them. Maybe they checked them in on another repo? I'll look. Thanks.

Diagnostic/watcherutil.py
```diff
@@ -6,17 +6,25 @@
...
 import subprocess
 import os
 import datetime
+import time
 
 
-class Watcher():
+class Watcher:
 
     def __init__(self, errorStream, outputStream, logtoconsole=False):
```
hosungsmsft: Minor: Maybe we should change all these to "error_stream", "output_stream", and "log_to_console" as well? :)

Diagnostic/DistroSpecific.py
```diff
@@ -0,0 +1,319 @@
...
+import re
+from Utils.WAAgentUtil import waagent
+
+omi_universal_install_cmd = 'bash scx-1.6.2-337.universal.x64.sh --upgrade'
+
+
+class CommonActions:
+    def __init__(self, logger):
+        self.logger = logger
+        pass
```
jasonzio: Removed.

Diagnostic/DistroSpecific.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def log_run_get_output(self, cmd):
+        """
+        Execute a command in a subshell
+        :param str cmd: The command to be executed
+        :return (int, str): A tuple of (subshell exit code, contents of stdout)
+        """
+        self.logger("RunCmd " + cmd)
+        error, msg = waagent.RunGetOutput(cmd, chk_err=True)
```
jasonzio: Done.

Diagnostic/DistroSpecific.py
```diff
@@ -0,0 +1,319 @@
...
+        return int(error), str(msg)
+
+    def log_run_ignore_output(self, cmd):
+        """
+        Execute a command in a subshell
+        :param str cmd: The command to be executed
+        :return int: The subshell exit code
+        """
+        self.logger("RunCmd " + cmd)
+        error, msg = waagent.RunGetOutput(cmd, chk_err=True)
```
jasonzio: Refactored so it just uses log_run_get_output() and discards the output. Added the new parameter.

Diagnostic/DistroSpecific.py
```diff
@@ -0,0 +1,319 @@
...
+        :return (int, str): A tuple of (sum of status codes, concatenated stdout from commands)
+        """
+        errors = 0
+        output = []
+        for cmd in cmds:
+            if with_timeout:
+                err, msg = self.log_run_with_timeout(cmd, timeout)
+            else:
+                err, msg = self.log_run_get_output(cmd)
+            errors += err
```
jasonzio: Bug for bug compatibility; that's exactly what the existing code base does.

Diagnostic/DistroSpecific.py
```diff
@@ -0,0 +1,319 @@
...
+
+class CentosActions(RedhatActions):
+    def __init__(self, logger):
+        RedhatActions.__init__(self, logger)
+
+    def install_extra_packages(self, packages, with_timeout=False):
+        install_cmd = 'rpm -qi PACKAGE; if [ ! $? == 0 ]; then yum install -y PACKAGE; fi'
+        return self.log_run_multiple_cmds([install_cmd.replace("PACKAGE", p) for p in packages], with_timeout)
+
+    def install_required_packages(self):
```
jasonzio: Nor do I, but this code preserves the exact behavioral differences in the existing code base. We can puzzle out *why* those differences exist later. For now... no change.

Diagnostic/DistroSpecific.py
```diff
@@ -0,0 +1,319 @@
...
+        for cmd in cmds:
+            if with_timeout:
+                err, msg = self.log_run_with_timeout(cmd, timeout)
+            else:
+                err, msg = self.log_run_get_output(cmd)
+            errors += err
+            output.append(msg)
+        return errors, ''.join(output)
+
+    def extract_om_path_and_version(self, results):
```
jasonzio: So I can expect to see those changes in a future PR, but no change now.

Diagnostic/Providers/Builtin.py
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+_omiClassName = { 'processor': 'SCX_ProcessorStatisticalInformation',
+                  'memory': 'SCX_MemoryStatisticalInformation',
+                  'network': 'SCX_EthernetPortStatistics',
+                  'filesystem': 'SCX_FileSystemStatisticalInformation',
+                  'disk': 'SCX_DiskDriveStatisticalInformation'
+                }
+
+_instancedClasses = [ 'network', 'filesystem', 'disk', 'processor' ]
```
jasonzio: Because memory isn't instanced? You can have multiple processors, multiple disks, multiple filesystems, multiple networks... only one memory instance. That's what "instanced" means in the OMI context.

Diagnostic/Providers/Builtin.py
```diff
@@ -0,0 +1,183 @@
...
+
+class BuiltinMetric:
+    def __init__(self, counterSpec):
+        t = ProvUtil.GetCounterSetting(counterSpec, 'type')
+        if t is None:
+            self._Type = 'builtin'
+        else:
+            self._Type = t.lower()
+            if t != 'builtin':
+                raise ProvUtil.UnexpectedCounterType('Expected type "builtin" but saw type "{0}"'.format(self.Type))
```
jasonzio: Correct, and fixed.

Diagnostic/Providers/Builtin.py
```diff
@@ -0,0 +1,183 @@
...
+        self._Counter = ProvUtil.GetCounterSetting(counterSpec, 'counter').lower()
+        if self._Counter not in _builtIns[self._CounterClass]:
+            raise ProvUtil.InvalidCounterSpecification(
+                'Counter {0} not in builtin class {1}'.format(self._Counter, self._CounterClass))
+        self._InstanceId = ProvUtil.GetCounterSetting(counterSpec, 'instanceId')
+        self._Label = ProvUtil.GetCounterSetting(counterSpec, 'counterSpecifier')
+        self._SampleRate = ProvUtil.GetCounterSetting(counterSpec, 'sampleRate')
+
+    def IsType(self, t):
+        return self.Type == t.lower()
```
jasonzio: One underscore is private-by-convention. Two underscores is implementation-reserved.

Diagnostic/Providers/Builtin.py
```diff
@@ -0,0 +1,183 @@
...
+#import time
+#import traceback
+#import xml.dom.minidom
+#import xml.etree.ElementTree as ET
+#
+#import Utils.HandlerUtil as Util
+#import Utils.LadDiagnosticUtil as LadUtil
+#import Utils.XmlUtil as XmlUtil
+#import Utils.ApplicationInsightsUtil as AIUtil
+#from Utils.WAAgentUtil import waagent
```
jasonzio: As noted in the introductory comments to the PR, this module is not active as a result of this PR, but I wanted to get it reviewed sooner rather than later. Unit tests are still required before the code would be ready for integration. I've added a significant comment at the beginning of the Builtin.py provider which attempts to explain how all this works.

Diagnostic/Utils/LadDiagnosticUtil.py
jasonzio: This style of data retrieval should be hoisted to a more general function. The names should probably be turned into constants, but the names are themselves part of the negotiation configuration protocol between the Portal diagnostics config blade and the agent. Pascal casing is the norm, but sometimes mistakes get through and you just have to deal with them.

Diagnostic/Utils/LadDiagnosticUtil.py
jasonzio: Tests for this stuff will come in a forthcoming PR.

Diagnostic/Providers/Builtin.py
hosungsmsft: Thanks for clarification, but it wasn't obvious to me. I'd still prefer to have that commented (mainly about the meaning of "instanced").

Or maybe the dictionary above can be extended with a boolean flag, indicating whether it's instanced or not? I think we should start reducing duplicate string constants as much as possible. Well, of course the right way of doing that would be to define Python constants on all these strings here...

Diagnostic/watcherutil.py
jasonzio: Done.

Diagnostic/Providers/Builtin.py
jasonzio: I think you meant "attributes" rather than "constants", but I take your meaning. I'll look into that for the future. In the meantime, I've added a comment which explains this.

Diagnostic/Providers/Builtin.py
hosungsmsft: The only comment I'm seeing at the beginning of the Builtin.py is the MIT license statement...?

Diagnostic/DistroSpecific.py
hosungsmsft: I see, and I guess then we have to do this. A good phrase I learned from this... :) Thanks.

Diagnostic/Providers/Builtin.py
jasonzio: Look after the license statement.

jasonzio:
Done (squash merge with edited history).

PR #320, opened 2017-01-11
[LAD] Self-mitigate potential mdsd memory leaking

hosungsmsft:
@jasonzio @LiweiPeng : Please review. Not sure why Liwei doesn't show up when I try to enter him in the "Reviewers"...

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -854,8 +854,7 @@ def start_mdsd():
                     memory_leak_msg = "Suspected mdsd memory leak (Virtual memory usage: {0}MB). " \
                                       "Recycling mdsd to self-mitigate.".format(int((mdsd_memory_usage_in_KB+1023)/1024))
                     hutil.log(memory_leak_msg)
-                    # Add a telemetry as well. This might ding our enable success KPI,
-                    # but there doesn't seem to be any other way.
+                    # Add a telemetry for a possible statistical analysis
                     waagent.AddExtensionEvent(name=hutil.get_name(),
                                               op=waagent.WALAEventOperation.HeartBeat,
                                               isSuccess=True,
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -695,6 +695,26 @@ def start_daemon():
         hutil.error("wait daemon start time out")
 
 
+def check_suspected_memory_leak(pid):
+    memory_leak_threshold_in_KB = 2000000  # Roughly 2GB
+    memory_rss_usage_in_KB = 0
+    memory_leak_suspected = False
+
+    try:
+        # Check /proc/[pid]/status file for "VmRSS" to find out the process's RSS memory usage
+        with open("/proc/{0}/status".format(pid)) as proc_file:
+            for line in proc_file:
+                if line.startswith("VmRSS:"):
+                    memory_rss_usage_in_KB = int(line.split()[1])
+                    memory_leak_suspected = memory_rss_usage_in_KB > memory_leak_threshold_in_KB
+                    break
+    except Exception as e:
+        # Not to throw in case any statement above fails (e.g., invalid pid). Just log.
+        hutil.error("Failed to check memory usage of pid={0}.\nError: {1}\nTrace:\n{2}".format(pid, e, traceback.format_exc()))
+
+    return memory_leak_suspected, memory_rss_usage_in_KB/1024  # return mem usage in MB
+
+
 def start_mdsd():
     global EnableSyslog, ExtensionOperationType
     with open(MDSDPidFile, "w") as pidfile:
@@ -827,6 +847,21 @@ def start_mdsd():
                     break
 
                 # mdsd is now up for at least 30 seconds.
+
+                # Mitigate if memory leak is suspected.
+                mdsd_memory_leak_suspected, mdsd_memory_usage_in_MB = check_suspected_memory_leak(mdsd.pid)
+                if mdsd_memory_leak_suspected:
+                    memory_leak_msg = "Suspected mdsd memory leak (memory usage: {0}MB). Recycling mdsd to self-mitigate.".format(mdsd_memory_usage_in_MB)
+                    hutil.log(memory_leak_msg)
+                    # Add a telemetry as well. This might ding our enable success KPI,
+                    # but there doesn't seem to be any other way.
+                    waagent.AddExtensionEvent(name=hutil.get_name(),
+                                              op=ExtensionOperationType,
+                                              isSuccess=False,
+                                              version=hutil.get_extension_version(),
+                                              message=memory_leak_msg)
+                    mdsd.kill()
+                    break
 
                 # Issue #128 LAD should restart OMI if it crashes
                 omi_was_installed = omi_installed   # Remember the OMI install status from the last iteration
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -696,23 +696,23 @@ def start_daemon():
 
 
 def check_suspected_memory_leak(pid):
-    memory_leak_threshold_in_KB = 2000000  # Roughly 2GB
-    memory_rss_usage_in_KB = 0
+    memory_leak_threshold_in_KB = 2000000  # Roughly 2GB. TODO: Make it configurable or automatically calculated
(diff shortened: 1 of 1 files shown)


Diagnostic/diagnostic.py
```diff
@@ -695,6 +695,26 @@ def start_daemon():
         hutil.error("wait daemon start time out")
 
 
+def check_suspected_memory_leak(pid):
+    memory_leak_threshold_in_KB = 2000000  # Roughly 2GB
```
LiweiPeng: Using a hard-coded 2GB doesn't sound like a good idea. On some small VMs like A1, 2GB is a lot of memory. Can this be made as some config parameter somewhere?

Diagnostic/diagnostic.py
```diff
@@ -828,6 +848,21 @@ def start_mdsd():
...
+                    memory_leak_msg = "Suspected mdsd memory leak (memory usage: {0}MB). Recycling mdsd to self-mitigate.".format(mdsd_memory_usage_in_MB)
+                    hutil.log(memory_leak_msg)
+                    # Add a telemetry as well. This might ding our enable success KPI,
+                    # but there doesn't seem to be any other way.
+                    waagent.AddExtensionEvent(name=hutil.get_name(),
+                                              op=ExtensionOperationType,
+                                              isSuccess=False,
+                                              version=hutil.get_extension_version(),
+                                              message=memory_leak_msg)
+                    mdsd.kill()
```
LiweiPeng: after mdsd was killed by mdsd.kill(), where will it be restarted?

Diagnostic/diagnostic.py
```diff
@@ -828,6 +848,21 @@ def start_mdsd():
...
+                if mdsd_memory_leak_suspected:
+                    memory_leak_msg = "Suspected mdsd memory leak (memory usage: {0}MB). Recycling mdsd to self-mitigate.".format(mdsd_memory_usage_in_MB)
+                    hutil.log(memory_leak_msg)
+                    # Add a telemetry as well. This might ding our enable success KPI,
+                    # but there doesn't seem to be any other way.
+                    waagent.AddExtensionEvent(name=hutil.get_name(),
+                                              op=ExtensionOperationType,
+                                              isSuccess=False,
+                                              version=hutil.get_extension_version(),
+                                              message=memory_leak_msg)
```
jasonzio: I disagree with setting the status to "failed". As a result of the mitigation, the service continues to run. I agree with the need for telemetry, but why can't you AddExtensionEvent with success==true? What is the total space of ExtensionOperationType - if we can add a new "extension operation type" which corresponds to "auto-restart", then we can log success for that new operation and easily (I hope) do analysis of the events reported in this way.

Diagnostic/diagnostic.py
hosungsmsft: The next statement is to break out of the inner "while True:" loop. The outer loop starts mdsd before entering the inner loop again.

jasonzio: In the next iteration of the outer loop. This particular python function is far too large to be easily understood, and Python's syntax doesn't help. I had the same question and it took me five minutes of scrolling to find exactly what happens after the ```break``` statement on 864, but sure enough eventually it comes right back to the top of the function (near enough).

Diagnostic/diagnostic.py
hosungsmsft: The only place I can think of is the extension's publicSettings (JSON), but I don't think that can be used anytime soon. I think this issue can be revisited later if/when we truly want to make this feature a publicly advertised one (which I personally don't prefer). Please note that this is just a first step in trying to self-mitigate rare occasions.

LiweiPeng: sounds good

Diagnostic/diagnostic.py
LiweiPeng: sounds good. nice if can refactor but OK for now.

Diagnostic/diagnostic.py
hosungsmsft: Because setting "isSuccess=True" may incorrectly make our Enable success rate better...? Call me a purist, but I'd choose to get penalized on this, rather than make it look incorrectly better than the actual.

I thought about using another extension operation type with success=true, but didn't look more until your comment here. I don't think there's an extension op type corresponding to auto-restart, but the closest seems to be "HeartBeat". Let me use that here with success=true. Thanks.

Diagnostic/diagnostic.py
```diff
@@ -695,6 +695,26 @@ def start_daemon():
...
+            for line in proc_file:
+                if line.startswith("VmRSS:"):
+                    memory_rss_usage_in_KB = int(line.split()[1])
+                    memory_leak_suspected = memory_rss_usage_in_KB > memory_leak_threshold_in_KB
+                    break
+    except Exception as e:
+        # Not to throw in case any statement above fails (e.g., invalid pid). Just log.
+        hutil.error("Failed to check memory usage of pid={0}.\nError: {1}\nTrace:\n{2}".format(pid, e, traceback.format_exc()))
+
+    return memory_leak_suspected, memory_rss_usage_in_KB/1024  # return mem usage in MB
```
hosungsmsft: Good point. Let me just return in KB and let the caller do the conversion in logging. Thanks.

Diagnostic/diagnostic.py
jasonzio: Hosung, please open an issue to make "Amount of memory consumed by mdsd which will trigger a restart" a PublicConfig parameter. That should help me remember to do something about this.

Diagnostic/diagnostic.py
jasonzio: I'm working through a series of refactorings of diagnostic.py; there is more than one issue/comment complaining about it being non-pythonic and monolithic, and those complaints are spot on. Expect to see something here in the near future.

Diagnostic/diagnostic.py
jasonzio: 👍

Diagnostic/diagnostic.py
```diff
@@ -695,6 +695,26 @@ def start_daemon():
...
+def check_suspected_memory_leak(pid):
+    memory_leak_threshold_in_KB = 2000000  # Roughly 2GB
+    memory_rss_usage_in_KB = 0
+    memory_leak_suspected = False
+
+    try:
+        # Check /proc/[pid]/status file for "VmRSS" to find out the process's RSS memory usage
+        with open("/proc/{0}/status".format(pid)) as proc_file:
+            for line in proc_file:
+                if line.startswith("VmRSS:"):
```
hosungsmsft: Really no reason, other than thinking using VmSize would be too aggressive, but if that's OK with folks, I'm good. Changed to VMSize.

jasonzio: I believe VmSize is the more relevant measure. RSS is just what's in active memory, but VmSize includes what's paged out, and that's still an issue for overall system health.

Diagnostic/diagnostic.py
```diff
@@ -695,6 +695,26 @@ def start_daemon():
...
+                    memory_rss_usage_in_KB = int(line.split()[1])
+                    memory_leak_suspected = memory_rss_usage_in_KB > memory_leak_threshold_in_KB
```
hosungsmsft: It appears so. Please see [this](http://unix.stackexchange.com/questions/199482/does-proc-pid-status-always-use-kb).

Diagnostic/diagnostic.py
hosungsmsft: Hmm... I can definitely do that, but if you think this is really important, I think we should just do it now. And I thought of automatically calculating this based on the installed physical memory size (/proc/meminfo: "MemTotal:    xxxxxxxx kB"). Could you give me a good formula on this threshold with that information?

Diagnostic/diagnostic.py
hosungsmsft: I have to admit that I contributed quite a bit to this monolithic-ness, and still adding more... Let's do clean things up here soon. Thanks.

Diagnostic/diagnostic.py
jasonzio: The "kB" is hardcoded. That's good to know; I had some parsing code I could share, but it's a relief to know we don't have to do that.

PR #324, opened 2017-01-13
DSC Updates.

ericgable:
Add optional public parameters.
Increment version.  
Update dsc package to 1.1.1-294 and omi to 1.1.0 version.


boumenot:
Do not check binaries into git.  Why do you need these binaries? Is there an alternative way to achieve what you want?

ericgable:
That's what's been there for 11 months.  I simply updated them.

ericgable:
Closing this for now - I'll re-open when I can get a download location for the OMI and DSC packages.  Perhaps at packages.microsoft.com rather than github.
Then I'll remove the binaries and the 'packages' directory.

PR #346, opened 2017-03-03
Prepost

bpramod:
Changing permissions validation
Config files names are changed
Plugin Scripts timeout is increased

### DIFF
--- a/VMBackup/main/PluginHost.py
+++ b/VMBackup/main/PluginHost.py
@@ -126,7 +126,7 @@ class PluginHost(object):
 
                 if os.path.isfile(pcpath):
                     permissions = self.get_permissions(pcpath)
-                    if int(permissions[0]) == 7 or int(permissions[1]) > 0 or int(permissions[2]) > 0:
+                    if int(permissions[1]) > 0 or int(permissions[2]) > 0:
                         self.logger.log('Plugin Config file does not have desired permissions', True, 'Error')
                         errorCode = CommonVariables.FailedPrepostPluginConfigPermissionError
                     if not self.find_owner(pcpath) == 'root':
--- a/VMBackup/main/PluginHost.py
+++ b/VMBackup/main/PluginHost.py
@@ -79,10 +79,11 @@ class PluginHost(object):
             self.logger.log('Plugin host Config file does not exist in the location ' + self.configLocation, True, 'Error')
             self.configLocation = './main/PluginHost.conf'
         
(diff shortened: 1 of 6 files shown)


VMBackup/main/PluginHost.py
```diff
@@ -77,12 +77,13 @@ def pre_check(self):
 
         if not os.path.isfile(self.configLocation):
             self.logger.log('Plugin host Config file does not exist in the location ' + self.configLocation, True, 'Error')
-            self.configLocation = './main/PluginHost.conf'
+            self.configLocation = './main/VMSnapshotPluginHost.conf'
         
+        permissions = self.get_permissions(self.configLocation)
```
vityagi: will this throw exception if file does not exist ?

VMBackup/main/PluginHost.py
```diff
@@ -77,12 +77,13 @@ def pre_check(self):
...
         if not os.path.isfile(self.configLocation):
             self.logger.log('Plugin host Config file does not exist in the location ' + self.configLocation, True, 'Error')
             errorCode =CommonVariables.FailedPrepostPluginhostConfigNotFound
-        elif not self.validate_permissions(self.configLocation):
+        elif (int(permissions[0]) == 7) or not (int(permissions[1]) == 0 or int(permissions[1]) == 4) or not (int(permissions[2]) == 0 or int(permissions[2]) == 4):
```
vityagi: remove int(permissions[0]) == 7) as by-default this file has 744 permissions

VMBackup/main/ScriptRunner.py
```diff
@@ -81,16 +81,25 @@ def get_config(self):
         try:
             with open(self.configLocation, 'r') as configFile:
                 configData = json.load(configFile)
```
vityagi: configData is under with block. is it not required to declare outside to access it outside ?

bpramod: In python,variables are scoped to innermost method it belongs to.if/for/with blocks doesn't count . So scoping is not issue but in our case if we don't enter the "with" block and access the configData, it throws error which is caught in the exception handling.

VMBackup/main/PluginHost.py
bpramod: ok

VMBackup/main/PluginHost.py
bpramod: it is handled in the get_permissions method itself

PR #347, opened 2017-03-03
Added Makefile to zip the DSC extension bits

balukambala:
Made these changes to onboard the DSC extension onto EDP


boumenot:
Addressed in a new PR.

PR #363, opened 2017-03-30
csfl: re-release 1.5.2.1 as 1.1.1

boumenot:
There are a few customers still running 1.1, and the failure rate is rather high.  Customers are no opting in auto upgrade, so are stuck at a bad version.  I am re-releasing 1.5.2.1 as 1.1.1 in the hopes that the failure rate will go down without disturbing the customer.


boumenot:
1.1.1 has been released. This was only opened for review, and should not be merged to trunk.  Closing.

Issue #328, opened 2017-01-24
OMS Agent extension for Linux and DSC extension for Linux seems incompatable 


slavizh:
If I install first the OMS Agent extension for Linux on a Azure VM and than the DSC extension for Linux, the DSC extension for Linux is taking forever to install. I didn't waited after 10 minutes passed but I would assume it cannot be installed and it will fail. You should investigate this and find what is the issue. If I first install the DSC Extension for Linux and than the OMS agent extension for Linux there is no problem.

balukambala:
@slavizh, sorry for the inconvenience. This has been fixed. Please let us know if you face any more issues.

PR #364, opened 2017-03-31
DSCForLinux - Fixed the OMS version conflict with other extensions

balukambala:
OMI is being used in OMSAgent, Asm Extensions. Added a version comparison check before upgrading OMI.

### DIFF
--- a/DSC/dsc.py
+++ b/DSC/dsc.py
@@ -38,6 +38,12 @@ DownloadDirectory = 'download'
 
 omi_package_prefix = 'packages/omi-1.0.8.ssl_'
 dsc_package_prefix = 'packages/dsc-1.1.1-70.ssl_'
+omi_major_version = 1
+omi_minor_version = 0
+omi_build = 8
+dsc_major_version = 1
+dsc_minor_version = 1
+dsc_build = 1
 omi_version_deb = '1.0.8.4'
 omi_version_rpm = '1.0.8-4'
 dsc_version_deb = '1.1.1.70'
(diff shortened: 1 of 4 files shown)


balukambala:
@boumenot, Can you please review this change. I will get my teammate bring upto speed on this extension code so that I don't bother you. Is there any policy that someone from core team also has to review the changes?

slavizh:
This is related to the issue I've opened I think. May be should be closed once you merge
https://github.com/Azure/azure-linux-extensions/issues/328

eshaparmar:
@balukambala please add me to the review list

DSC/dsc.py
```diff
@@ -264,14 +270,27 @@ def install_dsc_packages():
         deb_install_pkg(omi_package_path + '.x64.deb', 'omi', omi_version_deb)
         deb_install_pkg(dsc_package_path + '.x64.deb', 'dsc', dsc_version_deb )
     elif distro_category == DistroCategory.redhat or distro_category == DistroCategory.suse:
-        rpm_install_pkg(omi_package_path + '.x64.rpm', 'omi-' + omi_version_rpm)
-        rpm_install_pkg(dsc_package_path + '.x64.rpm', 'dsc-' + dsc_version_rpm)
+        rpm_install_pkg(omi_package_path + '.x64.rpm', 'omi')
+        rpm_install_pkg(dsc_package_path + '.x64.rpm', 'dsc')
+
```
eshaparmar: I would recommend adding tests for the new code. I dont see any tests for that

DSC/dsc.py
```diff
@@ -264,14 +270,27 @@ def install_dsc_packages():
...
+def compare_package_version(system_package_version, package_name):
+    if package_name == 'omi':
+        version = re.match('(\d+).(\d+).(\d+)', system_package_version)
```
eshaparmar: this regex match for omi and dsc seems duplicate. You should be able to create a separate function that takes in a major/minor version and then invoke that here with omi and dsc values.

DSC/dsc.py
```diff
@@ -264,14 +270,27 @@ def install_dsc_packages():
...
+        if version is not None and (int(version.group(1)) > omi_major_version) or (int(version.group(1)) == omi_major_version and int(version.group(2)) >= omi_minor_version) or (int(version.group(1)) == omi_major_version and int(version.group(2)) == omi_minor_version and int(version.group(3)) >= omi_build):
+            return 1
+    elif package_name == 'dsc':
+       	version = re.match('(\d+).(\d+).(\d+)', system_package_version)
+        if version is not None and (int(version.group(1)) > dsc_major_version) or (int(version.group(1)) == dsc_major_version and int(version.group(2)) >= dsc_minor_version) or (int(version.group(1)) == dsc_major_version and int(version.group(2)) == dsc_minor_version and int(version.group(3)) >= dsc_build):
+            return 1
+    return 0			
 
 def rpm_install_pkg(package_path, package_name):
```
eshaparmar: I would split this function (rpm_install_pkg) to two - validate_installed_pck and install_pck. That will  make this easy to test where validate will only check if the version is installed and return if it is. In case, the right version is not installed - code will install.

DSC/dsc.py
```diff
@@ -264,14 +270,26 @@ def install_dsc_packages():
...
+
+def check_pkg_exists(package_name, major_version, minor_version, build):
+    code,output = run_cmd('rpm -q --queryformat "%{VERSION}" ' + package_name)
+    waagent.AddExtensionEvent(name=ExtensionShortName, op='InstallInProgress', isSuccess=True, message="package name: " + package_name + ";  existing package version:" + output)
+    hutil.log("package name: " + package_name + ";  existing package version:" + output)
+    if code == 0: 
+        return compare_pkg_version(output, major_version, minor_version, build)
+
+def rpm_install_pkg(package_path, package_name, major_version, minor_version, build):
+    if check_pkg_exists(package_name, major_version, minor_version, build) == 1:
```
eshaparmar: Ideally, this path should also be tested in the unit tests using mocks. Something to think about in the future.

PR #370, opened 2017-04-18
Allow VMAccess to work in a system container

baude:
When you run the WALinuxAgent in a system container, the locations are
often different than a normal system -- almost like a chroot.  This
contains minor patches to help the VMAccess extension detect and deal
with that.


boumenot:
Awesome! I am validating these changes now against VMAccessForLinux and CustomScriptForLinux.

otubo:
@boumenot Hi! Any updates on this pr?

PR #371, opened 2017-04-19
Update OmsAgentForLinux extension

lagalbra:
Remove x86 bundle from package
Add proxy support
Add documentation for VM ResourceId
Support for Fairfax
Add stopOnMultipleConnections parameter
Fix hutil procedure when error occurs
Update README.md with current version
Improve exit codes and add error code processing
Add error code for unsupported OS
Add retry logic for failed install and uninstall

@varunkumta @NarineM


OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
-                dummy_command("Update", "success", "Update succeeded")
+                operation = 'Update'
     except Exception as e:
-        err_msg = ("Failed to enable the extension with error: {0}, "
-                   "{1}").format(e, traceback.format_exc())
-        waagent.Error(err_msg)
-        hutil.error(err_msg)
-        hutil.do_exit(1, 'Enable','failed','0',
-                      'Enable failed: {0}'.format(e))
+        waagent.Error(e.message)
```
hosungsmsft: Just clarifying (and also minor, as this should not really happen anyway): So you are not exiting here immediately, because you want the later log_and_exit() to take effect? Then the "message" below might be misleading, as the None 'operation' value might throw at line 80?

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
+        hutil = parse_context(operation)
+        if is_vm_supported_for_extension():
+            exit_code = operations[operation](hutil)
+        else:
+            log_and_exit(hutil, operation, 51, 'Unsupported operation system')
+
+        if exit_code is not 0:
+            message = (operation + ' failed with exit code {0}').format(exit_code)
+
+    except ValueError as e:
```
hosungsmsft: Not 100% sure, but I thought it'd be KeyError, rather than ValueError...?

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
+    exit_code = 0
+    message = operation + ' succeeded'
+
+    # Invoke operation
+    try:
+        hutil = parse_context(operation)
+        if is_vm_supported_for_extension():
+            exit_code = operations[operation](hutil)
+        else:
+            log_and_exit(hutil, operation, 51, 'Unsupported operation system')
```
hosungsmsft: Minor: I think it might be better (for diagnostic purpose later) to include the platform.dist() result in this log...?

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
-def getbundlefilename():
-    file_name = BundleFileNameTemplate.format('x86')
-    is_64bits = sys.maxsize > 2**32
-    if is_64bits:
-        file_name = BundleFileNameTemplate.format('x64')
-    return file_name
+def is_vm_supported_for_extension():
+    '''
+    Returns for platform.linux_distribution() vary widely in format, such as '7.3.1611' returned
+    for a machine with CentOS 7, so the first provided digits must match
```
hosungsmsft: This is an interesting thing for us (LAD devs) to note as well. Thanks and later we may adopt your logic here in LAD as well.

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
+        vm_ver_split = vm_ver.split('.')
+        for supported_ver in supported_dists[supported_dist]:
+            supported_ver_split = supported_ver.split('.')
+
+            vm_ver_match = True
+            for idx, supported_ver_num in enumerate(supported_ver_split):
+                try:
+                    vm_ver_num = vm_ver_split[idx]
+                except IndexError:
+                    vm_ver_match = False
```
hosungsmsft: Overall, the logic here seems a bit too complicated. Also, I don't quite understand why vm_ver_match should be False here. Maybe I'm confused... For example, if vm_ver == "6.7" and supported_ver == "6", shouldn't the resulting vm_ver_match be True? But I think your code here will assign False in that case? Just ignore if I'm misunderstanding your version matching logic...

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
-    file_name = getbundlefilename();
     file_directory = os.path.join(os.getcwd(), PackagesDirectory)
-    file_path = os.path.join(file_directory, file_name)
+    file_path = os.path.join(file_directory, BundleFileName)
 
     os.chmod(file_path, 100)
-    cmd = InstallCommandTemplate.format(file_name)
-    ScriptUtil.run_command(hutil, ScriptUtil.parse_args(cmd), file_directory, 'Install', ExtensionShortName, hutil.get_extension_version())
+    cmd = InstallCommandTemplate.format(BundleFileName)
+    waagent.Log("Starting command %s --upgrade." %(BundleFileName))
```
hosungsmsft: Minor: I'd change the string arg with .format() and also with cmd, rather than with BundleFileName. Same comment applies in uninstall().

OmsAgent/omsagent.py
hosungsmsft: OK, I saw you are raising ValueError in enable(), so that's probably why you are catching only ValueError here, but the parse_context() may throw KeyError if necessary JSON key is not present in either the privateSettings or the publicSettings, so I'd do the same on KeyError as well.

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
-    cmd = OnboardCommandTemplate.format(workspaceId, workspaceKey)
-    exit_code = ScriptUtil.run_command(hutil, ScriptUtil.parse_args(cmd), OmsAdminWorkingDirectory, 'Onboard', ExtensionShortName, hutil.get_extension_version(), False)
-    # if onboard succeeds we continue, otherwise fail fast
+        list_exit_code = ScriptUtil.run_command(hutil,
+                                                ScriptUtil.parse_args(WorkspaceCheckCommand),
+                                                OmsAdminWorkingDirectory,
+                                                'Check If Already Onboarded', ExtensionShortName,
+                                                hutil.get_extension_version(), False,
+                                                interval = 30,
+                                                std_out_file_name = output_file.name)
```
hosungsmsft: I now finally see why you need the tempfile module... Your better bet would be probably to use waagent.RunGetOutput(), rather than using ScriptUtil.run_command(). That way, you don't need to create a temporary file and redirect stdout to the temp file. waagent.RunGetOutput() returns the resulting stdout as a string. Lots of tribal knowledge... :)

OmsAgent/Fairfax/HandlerManifest.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "name":  "OmsAgentForLinux",
+    "version": "1.3.18.5",
+    "handlerManifest": {
+      "installCommand": "omsagent_ff.py -install",
```
lagalbra: @varunkumta might be able to provide more details on this, but one of the command templates is different, so there the two source files are not, in fact, identical.

OmsAgent/omsagent.py
```diff
@@ -46,100 +49,279 @@
...
 
+
 def main():
     waagent.LoggerInit('/var/log/waagent.log','/dev/stdout', True)
     waagent.Log("%s started to handle." %(ExtensionShortName))
 
+    # Determine the operation being executed
+    operation = None
     try:
         for a in sys.argv[1:]:
```
lagalbra: Whether an error gets thrown about NoneType, or operation being undefined, and error will get thrown if operation doesn't get defined in the checks below. In my opinion, a NoneType error would be more explicit.

OmsAgent/omsagent.py
lagalbra: This shouldn't really happen anyway, and exit_code will still be non-zero, which will print out multiple types of error logs. In a case like this, the larger type of logs should be error and not success.

OmsAgent/omsagent.py
lagalbra: If an IndexError is thrown, that means `supported_ver` is more precise (has more digits) than `vm_ver`, in which case we can't say for sure that `vm_ver` is supported.

Take Ubuntu, for example. I haven't seen an example case of this, but if `platform.linux_distribution()` returned only '16', then we can't know if it is 16.04, which we support, or 16.10, which we do not.

In your example, `vm_ver` being '6.7' and `supported_ver` being '6', yes, this should result in `vm_ver_match` being True. The IndexError will not get thrown in this case because `vm_ver` has more digits.

OmsAgent/omsagent.py
lagalbra: Is this waagent method available in all versions of WAAgent >= 2.1.5?

OmsAgent/README.md
```diff
@@ -31,10 +33,14 @@ Schema for the public configuration file looks like this:
 Schema for the protected configuration file looks like this:
 
 * `workspaceKey`: (required, string) the primary/secondary shared key of the workspace
+* `proxy`: (optional, string) the proxy connection string - of the form \[user:pass@\]host\[:port\]
+* `vmResourceId`: (optional, string) the full azure resource id of the vm - of the form /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}
```
lagalbra: @varunkumta or @sug4 might be able to answer this.

OmsAgent/omsagent.py
lagalbra: Looking through this code again, this Exception will get thrown if there's an issue with the regex, not with a valid operation being detected. This may not be a situation where we need to exit immediately.

OmsAgent/omsagent.py
lagalbra: Agreed. Will update.

OmsAgent/omsagent.py
lagalbra: Can you point out the location where KeyError is thrown? I'm not seeing any instance of that error type.

OmsAgent/omsagent.py
lagalbra: I will add logic to check if operation got set, and if not, then exit.

OmsAgent/omsagent.py
lagalbra: Will do.

OmsAgent/omsagent.py
lagalbra: I will briefly test using this, and if it works, then I may refactor some other code so that stdout and stderr can get read together.

OmsAgent/omsagent.py
lagalbra: Fixed.

OmsAgent/omsagent.py
lagalbra: Resolved.

OmsAgent/omsagent.py
lagalbra: Updated.

OmsAgent/omsagent.py
lagalbra: Added regardless.

OmsAgent/omsagent.py
lagalbra: Added comments clarifying this.

OmsAgent/omsagent.py
lagalbra: Fixed.

OmsAgent/omsagent.py
lagalbra: Briefly tested and changed.

OmsAgent/omsagent.py
hosungsmsft: Not sure if I was understood correctly...? I was talking about the for loop...? I really don't think that there are more than one cmdline args, so this for loop is definitely a noop (under the current extension execution model), but then I get to wonder why we keep this for loop in the first place.

OmsAgent/omsagent.py
hosungsmsft: Ah... Mixed up the two, so got confused... I think the logic could be better coded, but it's not my position to assert that, so it looks good to me now...

lagalbra:
@varunkumta @NarineM Let me know if this looks good. I made some other improvements.

OmsAgent/omsagent.py
lagalbra: @varunkumta Was there a reason we had this for loop? I kept it in case any old scenarios relied on it.

hosungsmsft: Please note that your extension will be executed only as specified in the HandlerManifest.json. Therefore, if you don't specify multiple cmdline args in there, then that can't happen.

lagalbra: Fixed.

NarineM:
@lagalbra 
Can you point me to the code where omsadmin.sh calls into restart omsagent? Because this change removed service_control enable command from enable() API , I just want to make sure it would not break reboot scenarios.
From what I can see, the only function that calls into enable_omsagent_service() is restart_omsagent. But I do not seem to find it being called in omsadmin.sh->onboard().

NarineM:
@lagalbra 
I am thinking instead of removing enable , we most likely need to replace it with restart_omsagent call. Regardless, the updated routine should always call into service_control.sh->enable_omsagent()

lagalbra:
omsadmin.sh calls start for the new workspace ID service [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/omsadmin.sh#L530), which calls start_omsagent [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/service_control#L385), and start_omsagent calls enable_omsagent_service [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/service_control#L144). I think this should cover the registration that you're thinking of.

NarineM:
@lagalbra 
I see, that path also calls start_omsagent() , so should be sufficient enough for all waagent scenarios. I signoff.

lagalbra:
@NarineM Thanks for reviewing! Will ask Hosung to merge this PR after I've completed internal testing of new commits.

PR #380, opened 2017-05-13
OmsAgent: Implement backups for hutil and waagent methods

lagalbra:
Improve errors and add check for workspace ID and key formats
Changed bundle version in omsagent.py
Add better error messages for common error codes
Do not print out shared key with onboarding command
Explicitly allow Ubuntu 16.10 to install/enable
Corrected dpkg retry logic
Add handling for WAAgent binary version 2.0.14 RunGetOutput
(RunGetOutput added a third parameter for version 2.0.16)

@NarineM

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -433,7 +433,7 @@ def run_command_with_retries(hutil, cmd, retries, check_error = True,
         exit_code, output = run_command_and_log(hutil, cmd, check_error, log_cmd)
         if exit_code is 0:
             break
-        elif re.match('^.*dpkg.+lock.*$', output):
+        elif not re.match('^.*dpkg.+lock.*$', output):
             break
         try_count += 1
         hutil_log_info(hutil, 'Retrying command "{0}" because package manager ' \
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -26,6 +26,7 @@ import platform
 import subprocess
 import json
 import base64
+import inspect
 
 try:
     from Utils.WAAgentUtil import waagent
@@ -100,8 +101,9 @@ def main():
 
         # For common problems, provide a more descriptive message
         if exit_code is 1 and operation == 'Install':
-            message = 'Install failed with exit code 1. Please make sure ' \
-                      'curl, libcurl, and python-ctypes are installed'
+            message = 'Install failed with exit code 1. Please check that ' \
+                      'dependencies such as curl and python-ctypes are ' \
+                      'installed.'
         elif exit_code is InstallTarErrorCode and operation == 'Install':
             message = 'Install failed with exit code {0}: please install ' \
                       'tar'.format(InstallTarErrorCode)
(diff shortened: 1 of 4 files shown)


OmsAgent/omsagent.py
```diff
@@ -64,59 +84,252 @@ def main():
...
-        if exit_code is not 0:
-            message = (operation + ' failed with exit code {0}').format(exit_code)
-
-    # ValueError may be thrown by enable
-    # KeyError may be thrown by parse_context
-    except (ValueError, KeyError) as e:
+        # For common problems, provide a more descriptive message
+        if exit_code is 1 and operation == 'Install':
+            message = 'Install failed with exit code 1. Please make sure ' \
+                      'curl, libcurl, and python-ctypes are installed'
```
NarineM: Libcurl is a lib that is used by curl. no need to specifically mention the sub dependancies here. I would also suggest to remove the explicit dependency package names and just print an error saying check for dependancies such as...

OmsAgent/omsagent.py
```diff
@@ -64,59 +84,252 @@ def main():
...
+        elif exit_code is InstallTarErrorCode and operation == 'Install':
+            message = 'Install failed with exit code {0}: please install ' \
+                      'tar'.format(InstallTarErrorCode)
```
NarineM: no longer needed. I believe the code has changed to automatically install tar if it is not present.

OmsAgent/omsagent.py
```diff
@@ -64,59 +84,252 @@ def main():
...
-    '''
-    Returns for platform.linux_distribution() vary widely in format, such as '7.3.1611' returned
-    for a machine with CentOS 7, so the first provided digits must match
-    '''
+    """
+    Checks if the VM this extension is running on is supported by OMSAgent
+    Returns for platform.linux_distribution() vary widely in format, such as
+    '7.3.1611' returned for a machine with CentOS 7, so the first provided
+    digits must match
+    Though Ubuntu 16.10 is not officially supported, we will allow it to
```
NarineM: Please remove this sentence "Though.." it can be confusing, since this was on azure Linux VM extension supported list for at least a year.

NarineM:
Please update the PR comments to also mention that supported distro API is changed too. That that change can be tracked from the history.

OmsAgent/omsagent.py
lagalbra: I want to keep this message just in case. We still have a catch in the code where we return this error code if tar does not successfully install: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L821

NarineM: fine, although I doubt that would ever happen.  I have not heard a case when zip or tar would fail to install . azure distros are all LST images.

OmsAgent/omsagent.py
lagalbra: Message adjusted.

OmsAgent/omsagent.py
lagalbra: Comment adjusted.

lagalbra:
Done.

PR #391, opened 2017-06-02
DSCForLinux

balukambala:
Added more tracing for enable functionality.

### DIFF
--- a/DSC/dsc.py
+++ b/DSC/dsc.py
@@ -427,10 +427,17 @@ def download_file():
 
 def download_azure_blob(account_name, account_key, file_uri, download_dir):
     waagent.AddExtensionEvent(name=ExtensionShortName, op="EnableInProgress", isSuccess=True, message="Downloading from azure blob")
-    (blob_name, container_name) = parse_blob_uri(file_uri)
-    host_base = get_host_base_from_uri(file_uri)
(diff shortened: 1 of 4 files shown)


balukambala:
@eshaparmar, please review when you get a chance

DSC/dsc.py
```diff
@@ -471,13 +481,15 @@ def get_host_base_from_uri(blob_uri):
     return netloc[netloc.find('.'):]
 
 def download_external_file(file_uri, download_dir):
+    waagent.AddExtensionEvent(name=ExtensionShortName, op="EnableInProgress", isSuccess=True, message="Downloading from external file")
```
eshaparmar: Instead of adding events at so many places - why dont you throw error (basically bubble it up) to a common place and then throw an event. That way you dont have to make changes everywhere should you add another param to the event or change existing ones. Something like this:

Try
{
  make sub calls....
}Catch{
 //catch all errors
addextensionevent....
}

eshaparmar: Since this is going to be some code change - i will leave it upto you to decide whether you want to make these changes or not. Signing off for now

balukambala: Currently, we are getting exceptions but generic ones. We need more logging and concrete exceptions to debug issues with Enable

PR #393, opened 2017-06-06
OmsAgent: retry logic fix and small updates

lagalbra:
Retry logic was not matching output correctly - fixed and tested.
Extension version 1.3.127.5 has been published with omsagent bundle version 1.3.5-127: I have updated the files reflecting that in this PR.
Added error code 12 to signify when dpkg was still locked after our retries; currently there are several, seemingly random, error codes that can be returned when this is the case. This will save time in debugging from logs.

@NarineM

### DIFF
--- a/OmsAgent/Fairfax/omsagent_ff.py
+++ b/OmsAgent/Fairfax/omsagent_ff.py
@@ -37,7 +37,7 @@ except Exception as e:
 
 # Global Variables
 PackagesDirectory = 'packages'
-BundleFileName = 'omsagent-1.3.4-127.universal.x64.sh'
+BundleFileName = 'omsagent-1.3.5-127.universal.x64.sh'
 
 # Always use upgrade - will handle install if scx, omi are not installed or
 # upgrade if they are
@@ -731,7 +731,7 @@ def log_and_exit(hutil, operation, exit_code = 1, message = ''):
     if hutil is not None:
         hutil.do_exit(exit_code, operation, exit_status, str(exit_code), message)
     else:
-        update_status_file(operation, exit_code, exit_status, message)
+        update_status_file(operation, str(exit_code), exit_status, message)
         sys.exit(exit_code)
 
 
--- a/OmsAgent/Fairfax/omsagent_ff.py
+++ b/OmsAgent/Fairfax/omsagent_ff.py
@@ -48,8 +48,7 @@ WorkspaceCheckCommandTemplate = '{0} -l'
 OnboardCommandWithOptionalParamsTemplate = '{0} -d opinsights.azure.us -w {1} -s {2} {3}'
 OmsAgentServiceScript = '/opt/microsoft/omsagent/bin/service_control'
 DisableOmsAgentServiceCommandTemplate = '{0} disable'
-InstallPythonCtypesErrorCode = 61
-InstallTarErrorCode = 62
+DPKGLockedErrorCode = 12
 
 # Configuration
 SettingsSequenceNumber = None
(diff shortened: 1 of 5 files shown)


lagalbra:
@NarineM I left the most common extension errors I've seen in the table, as well as the extension-only error codes, but I did clean up the table like you suggested.

OmsAgent/Fairfax/omsagent_ff.py
```diff
@@ -112,6 +113,10 @@ def main():
             message = 'Install failed with exit code {0}: please install ' \
```
NarineM: Just noticed that we are duplicating error codes in extension code . Bundle has already taking care of logging a proper message when given error code is thrown. These logs are available in azure extension logs. please remove duplicate customization from extension code.

OmsAgent/Fairfax/omsagent_ff.py
```diff
@@ -731,7 +731,7 @@ def log_and_exit(hutil, operation, exit_code = 1, message = ''):
     if hutil is not None:
         hutil.do_exit(exit_code, operation, exit_status, str(exit_code), message)
     else:
-        update_status_file(operation, exit_code, exit_status, message)
+        update_status_file(operation, str(exit_code), exit_status, message)
```
NarineM: This seems a regression from the previous commit. Can you update the description of this commit to state so?
Also please send them in separate PR, it is easy for the reviewer miss something if there are more than one unrelated commits in a single PR.

OmsAgent/Fairfax/omsagent_ff.py
lagalbra: The messages logged by the omsagent bundle are not exposed as an extension message, from the perspective that if the extension install fails, these strings are what they will see in the Azure portal.
Most customers do not know where on the machine to find more descriptive logs on the machine, and we should not require them to go to the machine to figure out the next step.

OmsAgent/Fairfax/omsagent_ff.py
lagalbra: This change is from commit https://github.com/Azure/azure-linux-extensions/pull/393/commits/c90649a2cfa93db092bf9bdfb61767e39f1cff07

This change is already included in the current public version of the extension, but it hasn't been migrated to the Azure code base yet. It's a small change, and it was quicker to include it in this PR.

NarineM: Again, as I said please remove duplicated logic. extension logging the error code returned from the bundle is already sufficient enough to diagnose the issue. Also entire verbose output from the omsagent operations are logged in extension logs.

NarineM: I did not see this in your older commit, but since it is duplicated only for one error code so far, let us remove it.

NarineM: What I mean is only VM extension specific error should be logged here. Any error code that is thrown by the bundle already, should be maintained in the agent code base only.

OmsAgent/Fairfax/omsagent_ff.py
NarineM: We do not have too, error code for extension are documented on GitHub if someone needs to dig further. Also the bundle installation messages are logged in extension log. I can setup a machine for you to show.

OmsAgent/Fairfax/omsagent_ff.py
lagalbra: I agree with the principle of that; however, I think we should keep the specific message for error code 1.

This error can be generated for several scenarios, which is why it's not provided in our [error code table](https://github.com/Azure/azure-linux-extensions/tree/master/OmsAgent#error-codes-and-their-meanings), but the most common is that curl is not installed. Until the INSTALL_CURL error code change is added to the extension, we should inform the customer here to check for dependencies.

I will make the change to remove the logic for the other error codes, which have not reoccured in the Kusto error logs recently.

OmsAgent/Fairfax/omsagent_ff.py
```diff
@@ -99,19 +98,16 @@ def main():
         hutil = parse_context(operation)
         exit_code = operations[operation](hutil)
 
-        # For common problems, provide a more descriptive message
+        # Exit code 1 indicates a general problem that doesn't have a more
+        # specific error code; it often indicates a missing dependency
         if exit_code is 1 and operation == 'Install':
             message = 'Install failed with exit code 1. Please check that ' \
                       'dependencies such as curl and python-ctypes are ' \
                       'installed.'
```
NarineM: Can you remove dependency package names from the message? Put something like "check dependencies".
Also can you extend this message to say "for details check extension log"?

lagalbra: Done.

lagalbra:
Merged into Azure/master

PR #396, opened 2017-06-07
Fix "local" keys to read "locale"

### LINKED ISSUE #397: [LAD] Some elements in LAD 2.3 compat config misspell the word "locale"
Portal expectations for Azure Metrics definitions specify the locale of the display name for a metric in a key named "locale". Some of the entries in the LAD 2.3 compatibility config for LAD 3.0 misspelled that key as "local".

jasonzio:
Fixes #397

### DIFF
--- a/Diagnostic/tests/lad_2_3_compatible_portal_pub_settings.json
+++ b/Diagnostic/tests/lad_2_3_compatible_portal_pub_settings.json
@@ -264,7 +264,7 @@
             "annotation": [
               {
                 "displayName": "Filesystem transfers/sec", 
-                "local": "en-us"
+                "locale": "en-us"
               }
             ], 
             "class": "filesystem", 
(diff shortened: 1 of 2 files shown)


LiweiPeng:
thanks for fixing this. looks good to me

hosungsmsft:
Embarrassing... And probably the better place to fix this would be [this line](https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/tests/test_builtin.py#L267). Of course you'll need the input lad_2_3_metric_definitions_sample.json to generate the corrected output JSON, but just for completeness sake, I'd ask you to fix that generator code as well.

It's also interesting to note that there are two more typos of the same kind: [here ](https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/tests/test_builtin.py#L23) and [here](https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/tests/test_builtin.py#L37). I'd suggest to fix those 2 as well.

Issue #400, opened 2017-06-10
[LAD] Lists of sinks should tolerate white space


jasonzio:
The values for "sinks" should tolerate embedded white space. That,
```json
{
    "sinks": "sink1, sink2, sink3"
}
```
should work correctly.

jasonzio:
Fixed #401

Issue #399, opened 2017-06-10
[LAD] Should accept SAS token if it has a leading question mark


jasonzio:
Some tools that create SAS tokens stick the question mark on the front end, since a token is functionally a query string; some tools do not. LAD should accept either form.

Issue #398, opened 2017-06-09
[LAD] If an exception is thrown while loading the LAD config, the extension status info is incorrect


jasonzio:
If the loading of a LAD configuration results in LAD catching an exception, the extension substatus (in the guest agent status blob) is set from an uninitialized variable.
```
- VM has reported a failure when processing extension 'LinuxDiagnosticExtension'. Error message: "Extension operation Daemon failed:local variable 'config_invalid_reason' referenced before assignment". (Code: VMExtensionProvisioningError)
```

Issue #390, opened 2017-06-01
[LAD] Files not readable by omsagent can't be captured by fileLogs feature


jasonzio:
The fileLogs feature of LAD is handled by omsagent, which effectively "tail"s the logs and throws newly appended lines into mdsd so they can be written to the configured destination(s). While LAD itself runs as root/root, the omsagent runs as omsagent/omsusers. If a file can't be read by that user/group, no data is captured, and omsagent itself spews error messages to its log. (10MB/hour or more.)

LAD should check fileLogs targets for accessibility by omsagent/omsusers when it generates configs; if a file is not accessible (either because the file isn't readable or because one or more directories along the path aren't searchable), LAD should log an error and not configure omsagent to capture the file.

PR #401, opened 2017-06-10
Small fixes to open issues

jasonzio:
Use omsagent's sudo_tail input plug-in to read files regardless of mode (#390)
Set valid extension substatus if parsing the config raises an exception (#398)
Handle leading '?' in SAS token (#399)
Strip names when splitting sink names (#400)
Run all the unit tests as part of BVT
Small corrections to README
Fixed typos in LAD 2.3-compatible config sample (in tests/)


LiweiPeng:
:shipit:

PR #420, opened 2017-07-06
DSCForLinux: Updated status file with nodeid and VMUUID

balukambala:
Made changes to update status file with Node id and VMUUID after successful enable operation.@weijiel, please review


DSC/dsc.py
```diff
@@ -47,6 +49,8 @@
 dsc_build = 1
 dsc_release = 294
 package_pattern = '(\d+).(\d+).(\d+).(\d+)'
+nodeid_path = '/etc/opt/omi/conf/dsc/agentid'
+DateTimeFormat = "%Y-%m-%dT%H:%M:%SZ"
```
vrdmr: Could you use the snake notation here too (with _ as a separator, and lowercasing)

PR #421, opened 2017-07-07
Update OSPatching Documentation

jacwil:
Updates the OSPatching README to reflect the change in the configuration schema made long ago.


OSPatching/README.md
```diff
@@ -1,27 +1,25 @@
...
-* the status of VM can be checked by user-defined scripts,
-which can be stored locally, in github or Azure Storage
+* Patch the OS as a one-time task
+* The patching can be stopped before the actual patching operation begins
+* The status of VM can be checked by user-defined scripts stored locally, in GitHub, or in Azure Storage
 
 # User Guide
 
 ## 1. Configuration schema
+All settings are set in the protected configuration. No settings are available in the public configuration and it can be omitted.
```
boumenot: This is a minor change (2.0 to 2.3), but this is a breaking change.  I would not do this.  You should support both modes unless there's a need to break customers.

OSPatching/README.md
```diff
@@ -337,46 +307,44 @@ You can stop the OS updates for debugging. Once the “stop” parameter is set
...
-see the status on Azure Portal
+## Debugging
 * The operation log of the extension is `/var/log/azure/<extension-name>/<version>/extension.log` file.
+* The installation status of the extension is reported back to Azure so that the user can see the status on Azure Portal.
+  This does not mean the OSPatching Extension successfully applied the current configuration to the VM.
+* Attempting to enable the OSPatching Extension 2 or more times with the same configuration will result in nothing happening.
+  See [Enable the extension repeatedly](#3.5 Enable the extension repeatedly) section above for more details.
+
+## Mitigations
+If the OSPatching Extension is not working, your Linux distro likely has a recommended way to configure automatic updates. Some distros such as Ubuntu will have an automated way to keep your machine patched. Please refer to your distro's documentation for more information.
```
boumenot: Is this a recommendation to just use the Distro's preferred method of OS patching?  If so, I agree, but I think the wording could be better.  If you know of a existing process for the endorsed distros I would add that text here.

jacwil: As a mitigation, yes the recommendation is to just use the distro's preferred method. I can compile a list of links to each endorsed distro's auto-upgrade page and put them here.

OSPatching/README.md
jacwil: This is just an update to documentation to match what's been in source code since long ago. No breaking changes are being made and customers trying to use this extension are being bit when they try to follow the old documentation. I hope that makes sense.

PR #424, opened 2017-07-19
Add package's waagent copy to the search path

boumenot:
Most of the OSTC extensions rely on a copy of waagent for functionality,
and the default search was /usr/sbin/waagent.  Distros (Debian) have
removed this, and broken extensions.  Instead of looking in /usr/sbin look
for waagent in the package's local directory.


boumenot:
I am spamming everyone I know about this.  I believe this will fix our issues on Debian 8.  It turns out that most extensions already include waagent in the extension, but I failed to update the search path to include this local version.  This change prioritizes the local version above all others.

/cc: @brendandixon, @hglkrijger, @lagalbra, @hosungsmsft, @balukambala, @jacwil

balukambala:
Changes look good

brendandixon:
@boumenot LGTM.

PR #433, opened 2017-08-02
Include waagent 2.0.16

waldiTM:
Instead of trying and failing to use a system supplied copy of the old agent, just include a local copy always.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ EXTENSIONS = \
 	CustomScript \
 	DSC \
 	OSPatching \
+	VMAccess \
 	VMBackup
 
 clean:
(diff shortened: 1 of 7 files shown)


boumenot:
It appears that we've done the same work.  I think this is a dupe of #440.  Is there any reason to keep this PR open too?

waldiTM:
You still need to remove the code to every try to open /usr/sbin/waagent

waldiTM:
Also, how can it a dupe if it fixes all, instead of only one of the included extensions?

boumenot:
There are more extensions in this repo than the ones called out in the Makefile, so the change covers more than VMAccessForLinux.  I am not going to remove the other code because it is shared among most of the extensions in this repo.

PR #437, opened 2017-08-07
vmaccess: do not refer to temp files by name

boumenot:
Create telemetry events for the scenarios executed by VMAccess.

### DIFF
--- a/RDMAUpdate/main/Utils/HandlerUtil.py
+++ b/RDMAUpdate/main/Utils/HandlerUtil.py
@@ -52,6 +52,7 @@ Example Status Report:
 """
 
 
+import base64
 import os
 import os.path
 import sys
@@ -137,7 +138,7 @@ class HandlerUtility:
         except:
             self.error('JSON exception decoding ' + ctxt)
 
-        if config == None:
+        if config is None:
             self.error("JSON error processing settings file:" + ctxt)
         else:
             handlerSettings = config['runtimeSettings'][0]['handlerSettings']
@@ -149,14 +150,12 @@ class HandlerUtility:
                 thumb = handlerSettings['protectedSettingsCertThumbprint']
                 cert = waagent.LibDir + '/' + thumb + '.crt'
                 pkey = waagent.LibDir + '/' + thumb + '.prv'
-                f = tempfile.NamedTemporaryFile(delete=False)
-                f.close()
-                waagent.SetFileContents(f.name, protectedSettings)
-                cleartxt = waagent.RunGetOutput("base64 -d " + f.name + " | openssl smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
-                os.remove(f.name)
+                unencodedSettings = base64.standard_b64decode(protectedSettings)
+                openSSLcmd = "openssl smime -inform DER -decrypt -recip {0} -inkey {1}"
+                cleartxt = waagent.RunSendStdin(openSSLcmd.format(cert, pkey), unencodedSettings)[1]
                 if cleartxt is None:
                     self.error("OpenSSL decode error using thumbprint " + thumb)
-                    do_exit(1,'Enable','error','1', 'Failed to decode protectedSettings')
+                    self.do_exit(1, "Enable", 'error', '1', 'Failed to decrypt protectedSettings')
                 jctxt = ''
                 try:
                     jctxt = json.loads(cleartxt)
--- a/RDMAUpdate/main/Utils/HandlerUtil.py
+++ b/RDMAUpdate/main/Utils/HandlerUtil.py
@@ -57,6 +57,7 @@ import os.path
 import sys
 import json
 import time
+import tempfile
 from os.path import join
 from Utils.WAAgentUtil import waagent
 from waagent import LoggerInit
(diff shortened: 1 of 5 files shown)


mbearup:
Approved

VMAccess/vmaccess.py
```diff
@@ -95,10 +96,12 @@ def enable():
             _backup_sshd_config(SshdConfigPath)
 
         if reset_ssh:
+            waagent.AddExtensionEvent(name=hutil.get_name(), op="scenario", isSuccess=True, message="reset-ssh")
             _reset_sshd_config(SshdConfigPath)
             hutil.log("Succeeded in reset sshd_config.")
 
         if remove_user:
+            waagent.AddExtensionEvent(name=hutil.get_name(), op="scenario", isSuccess=True, message="remove-user")
```
anandramakrishna: These telemetry events are just telling you the scenario, but there does not appear to be any real success/failure.  I.e. isSuccess is always True.  Perhaps you should consider including that.

Perhaps you need a telemetry event in the exception handler portion to flag a failure?

VMAccess/vmaccess.py
```diff
@@ -255,6 +261,7 @@ def _set_user_account_pub_key(protect_settings, hutil):
                 waagent.ChangeOwner(pub_path, user_name)
                 os.remove('temp.pub')
                 os.remove('temp.crt')
+                waagent.AddExtensionEvent(name=hutil.get_name(), op="scenario", isSuccess=True, message="create-user")
```
anandramakrishna: Same message and telemetry as above.  Intentional?

VMAccess/vmaccess.py
boumenot: I am more interested in intent as opposed to success.  I can track success (or not) by using the Enable event, and correlating it with the scenario event.

VMAccess/vmaccess.py
boumenot: Ditto.

RDMAUpdate/main/Utils/HandlerUtil.py
```diff
@@ -148,13 +149,14 @@ def _parse_config(self, ctxt):
...
-                cleartxt = None
-                cleartxt = waagent.RunGetOutput("base64 -d /tmp/kk | openssl smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
-                os.remove("/tmp/kk")
-                if cleartxt == None:
-                    self.error("OpenSSh decode error using  thumbprint " + thumb)
-                    do_exit(1,operation,'error','1', operation + ' Failed')
+                f = tempfile.NamedTemporaryFile(delete=False)
+                f.close()
+                waagent.SetFileContents(f.name, protectedSettings)
+                cleartxt = waagent.RunGetOutput("base64 -d " + f.name + " | openssl smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
```
jasonzio: I've seen this before, and I could've sworn we fixed this before. The right solution is to eliminate the use of a disk file completely.  Hosung fixed this back in January, and it *should* have been in such a way that all extensions were affected. Looks like it went into LAD 2.3.9019.

jasonzio: I'm wrong. Hosung fixed this in December 2016. The commit in the azure-Linux-extensions repo is [7315f17d](https://github.com/Azure/azure-linux-extensions/commit/7315f17d4616b5220749ba4b8cb011f9a55c0cb7#diff-7f84beab04579bef70043ca0cc72fb85). The change completely eliminates the use of the temporary file.

hosungsmsft: I just applied the code (that uses stdin instead of a temp file at all) you provided at that time. It's in the common Utils (top-level) directory, and this appears to be a copy of that directory from an old time...

VMAccess/vmaccess.py
```diff
@@ -324,7 +331,9 @@ def _reset_sshd_config(sshd_file_path):
                     break
 
         # Prepare cloud init config for coreos-cloudinit
-        cfg_tempfile = "/tmp/cloudinit.cfg"
+        f = tempfile.NamedTemporaryFile(delete=False)
+        f.close()
+        cfg_tempfile = f.name
```
jasonzio: It amazes me that coreos-cloudinit has something like 12 options for where to get the data from, and *none of them* is "from stdin". I don't read go well enough to see for myself - if no "data source" option is provided at all, will coreos-cloudinit read from stdin? If so, it would be best to clone the "getting rid of /tmp/kk" change I referenced elsewhere to do the same thing here.

boumenot: I do not see the value in piping this construct through STDIN.  The requirement is to **not** use a guessable temporary file, which is what this change fixes.  There are no protected settings exposed in this configuration, the intent is only to prevent others from modifying the file before application.

RDMAUpdate/main/Utils/HandlerUtil.py
boumenot: It was fixed.  Welcome to the world of copy-and-paste.  I tried to find all other occurrences and scrubbed them too.

jasonzio:
If the idea is to prevent modification before application, you’ve merely reduced the window of vulnerability, you haven't closed it. That might be enough, especially if tempfile.NamedTemporaryFile() creates the file mode 600 as part of the atomic open-if-not-exist operation.

Sent from my Windows 10 phone

boumenot:
According to the docs this is the way it works, see https://docs.python.org/2/library/tempfile.html.

VMAccess/vmaccess.py
jasonzio: OK, given our conversation about the safety of tempfile.NamedTemporaryFile(), I'm okay with this code as-is.

RDMAUpdate/main/Utils/HandlerUtil.py
jasonzio: I suppose I should wait to see the commit with the fixes before changing my vote. :-)

boumenot: :)

PR #440, opened 2017-08-13
Ship waagent with extension

boumenot:
Fix support on Debian 8 and greater by shipping waagent with OSTC extensions (at least most of them).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ EXTENSIONS = \
 	CustomScript \
 	DSC \
 	OSPatching \
+	VMAccess \
 	VMBackup
 
 clean:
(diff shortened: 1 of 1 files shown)


waldiTM:
How does this relate to #433?

jacwil:
@waldiTM are all the changes in #433 necessary for Debian 8 and greater? #433 seems to pull in the entire waagent2.0. I'm not sure if we want to manage a copy of it in this repo.

waldiTM:
@jacwil I see it as two ways to fix the same problem. It's not that there will be any changes to the old 2.0 line of the agent, so pulling it in makes sure there is a version the extensions can work with. It also makes sure that the system copy will never be touched again, making another workaround not longer needed.

boumenot:
A copy already lives in this repo, and has for quite some time.  I do not see that we're taking on anything new.

waldiTM:
Where? The code reads external files.

boumenot:
https://github.com/Azure/azure-linux-extensions/tree/master/Common

Issue #439, opened 2017-08-11
[VMAccess] users created are not listed under user group sudo


z01c:
According to the security notes - 
VMAccess Extension is designed for regaining access to a VM in the event that access is lost.
Based on this principle, **it will grant sudo permission to the account specified in the username field**.

However, users created are not listed under user group sudo.

Related log from /var/log/azure/Microsoft.OSTCExtensions.VMAccessForLinux/1.4.7.0/extension.log -

2017/08/10 22:55:13 [Microsoft.OSTCExtensions.VMAccessForLinux-1.4.7.0] sequence number is 1
2017/08/10 22:55:13 [Microsoft.OSTCExtensions.VMAccessForLinux-1.4.7.0] setting file path is/var/lib/waagent/Microsoft.OSTCExtensions.VMAccessForLinux-1.4.7.0/config/1.settings
2017/08/10 22:55:13 [Microsoft.OSTCExtensions.VMAccessForLinux-1.4.7.0] JSON config: {"runtimeSettings": [{"handlerSettings": {"protectedSettingsCertThumbprint": "92D202B54029919C6CCF3D2FA28D982802
8D33BC", "publicSettings": {"abdtest4": "abdtest4", "abdtest3": "abdtest3"}, "protectedSettings": "MIICCQYJKoZIhvcNAQcDoIIB+jCCAfYCAQAxggFpMIIBZQIBADBNMDkxNzA1BgoJkiaJk/IsZAEZFidXaW5kb3dzIEF6dXJlIE
NSUCBDZXJ0aWZpY2F0ZSBHZW5lcmF0b3ICEF5Nbcu7WHO2TteXpTszAbUwDQYJKoZIhvcNAQEBBQAEggEAkqouF/FJs9lE92G6q7Dy/Qz2gJK/ZUlks1cm3KGcVxOLd3fazDdZ3eygevN1tjr5uUAStgMI0jUcyQUxwQmnEFpDPy3o+u/p91qvx3qwmEgdUiQldA4
c89oQT8SM1C/VpyDY5J6YT3dumZD3HOsl5QhLchbOZ79dWwB2XD27+a5rmZut+Pk7HFD7BOCooOnL0rR3sNxvn+y2L3iYM3M9GL3rny4qEeAIVMItjsVdJy37lKHN2fOhBq5p18IwuT5VsBoczD8IFWaZ/OZol33R9bvF0Bv4UPUv4+4CWHqvwIBNhtNSP3faDNrZ
[...]

PR #441, opened 2017-08-14
Adduser to sudo

z01c:
This change is to add the newly created user to user group "sudo". Background info: https://github.com/Azure/azure-linux-extensions/issues/439

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -24,6 +24,7 @@ import sys
 import tempfile
 import time
 import traceback
+import subprocess
 
 import Utils.HandlerUtil as Util
 from waagentloader import load_waagent
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -209,6 +209,10 @@ def _set_user_account_pub_key(protect_settings, hutil):
                                   message="(02101)" + err_msg)
         raise Exception(err_msg + " with " + error_string)
     hutil.log("Succeeded in create the account or set the password.")
+    try:
+        subprocess.call("sudo adduser " + user_name + " sudo", shell = True)
+    except Exception as e:
+        hutil.log(str(e))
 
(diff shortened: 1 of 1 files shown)


boumenot:
Thanks for the contribution!

Two points on feedback:

 1. What distros have you tested this change on?  I do not think this will work across all the distros we support because I do not expect them all have a sudo group.  Please see the README for everything we have to support.
 1.   Adding a user to a group should be a distinct operation, and have its own configuration.  Please add a new operation to account for adding a user to a group.

z01c:
Thanks a lot for the quick feedback!
I had tested only on Ubuntu. Is it possible for me to add a new parameter called user_group in the protected configuration and call the function to add the user only when user_group is defined?
```
def _add_user_to_group(protect_settings, hutil)
    user_name = protect_settings['username']
    user_group = protect_settings.get('user_group')
    if "Ubuntu" in platform.platform():
        try:
            subprocess.call("sudo adduser " + user_name + " " + user_group + "", shell=True)
        except Exception as e:
            hutil.log(str(e))
```

boumenot:
Yes, this is what I was thinking.  

I would check to see if there's an existing method to add a user to a group.  If so, you should call that instead of rolling your own.

There is no need to call "sudo adduser" because the extension executes as root already.

I won't be able to release this change until we support the platforms listed in the README.  You can wait for me to add that support, or you can try to add it - your call.

z01c:
Thanks for the feedback. We are making a small change from our end to tackle the issue for now. After you add support for adding users to different groups via the vmaccess extension we will use that. Thanks.

z01c:
> I would check to see if there's an existing method to add a user to a group. If so, you should call that instead of rolling your own.
Is there such a method? We are working with the workaround but the real issue still exists. If there's a method we can create an ACIS action to add users to sudo group. Please let us know. Thanks.

PR #452, opened 2017-09-06
Require SAS token cred; reject storageKey cred

### LINKED ISSUE #451: LAD 3.0 returns unhelpful error if storage SAS token not passed in protected settings
If storageAccountSasToken is not specified in the protected settings, or if it's an empty string, LAD 3.0 returns a singularly unhelpful error message:
```
2017/09/05 21:52:14 [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.109] Invalid config settings given: Unknown exception occurred from generate_all_configs(). Msg: string index out of range. Can't proceed, but this will be still considered a success as it's an external error.
2017/09/05 21:52:14 [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.109] Daemon,success,0,Invalid config settings given: Unknown exception occurred from generate_all_configs(). Msg: string index out of range. Can't proceed, but this will be still considered a success as it's an external error.
```

jasonzio:
Fixes #451 
Also, reject any attempt to support storage key _a la_ LAD 2.3; while LAD 2.3 continues to support that for backwards compatibility, it's not a good thing from a security standpoint. With this change, if a user attempts to pass that to LAD 3.0, the extension will refuse to start up.

### DIFF
--- a/Diagnostic/ChangeLogs
+++ b/Diagnostic/ChangeLogs
@@ -1,3 +1,7 @@
+2017-09-05: LAD-3.0.111
+    - Ensure SAS storage token is supplied
+    - Explicitly reject deprecated use of LAD 2.3's storageAccountKey
+
 2017-08-11: LAD-3.0.109
     - Fix waagent-related issue on Debian distros
     - Add additional unit tests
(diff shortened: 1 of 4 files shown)


LiweiPeng:
looks good to me. Thanks.

PR #463, opened 2017-09-14
Add mdsd source code

jasonzio:
This source is provided to enable auditing of all elements of the Linux Diagnostic Extension. While software construction tooling is included, we don't particular expect anyone to alter or extend this particular component.


LiweiPeng:
Building the docker image based on Dockerfile failed. Did you miss the azure.list?

Step 8 : ADD azure.list /etc/apt/sources.list.d/azure.list
INFO[0104] azure.list: no such file or directory

> echo $?
1

PR #474, opened 2017-10-16
VMEncryption: Centos partition issue fix

vermashi:
Fix for https://icm.ad.msft.net/imp/v3/incidents/details/49002787/home

For this fix we wait for the partition table changes to be read before proceeding to format the new drives.


ejarvi:
LGTM

PR #510, opened 2018-01-04
Use systemd on CentOS/RHEL 7+

### LINKED ISSUE #490: mdsd-lde service won't install on CentOS/RHEL 7+
mdsd-lde service does not get installed on CentOS 7.3 when I enable LinuxDiagnostic v3 extension.
This seems to be due to `use_systemd` returning `False` for CentOS/RHEL in `Diagnostic/DistroSpecific.py`. systemd has been introduced in CentOS/RHEL 7.

a-shevtsov:
This PR fixes #490 issue - (mdsd-lde service won't install on CentOS/RHEL 7+)

### DIFF
--- a/Diagnostic/DistroSpecific.py
+++ b/Diagnostic/DistroSpecific.py
@@ -312,6 +312,27 @@ class CentosActions(RedhatActions):
         return self.install_extra_packages(('policycoreutils-python',), True)
 
 
+class Redhat7Actions(RedhatActions):
+    def __init__(self, logger):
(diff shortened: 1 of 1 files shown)


jasonzio:
Consider refactoring the oracle/oracle7 requirements; looks to me like redhat doesn't need one extra package that oracle does, and oracle7 appears to need two extra packages that redhat7 doesn't need.

a-shevtsov:
Added classes for Oracle OS

PR #514, opened 2018-01-20
add resource disk encryption and environment prechecks

ejarvi:
- encrypt format resource disk when EncryptFormatAll is used
- add reliability checks to inspect system environment prior to encryption


VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+        return (int)(self.executor.Execute(cmd, suppress_logging=True)) == 0
+
+    def resource_disk_exists(self):
+        """ true if the udev name for resourced disk exists """
+        cmd = 'test -b ' + self.RD_BASE_DEV_PATH
+        return (int)(self.executor.Execute(cmd, suppress_logging=True)) == 0
+
+    def resource_disk_partition_exists(self):
+        """ true if udev name for resource disk partition exists """
+        cmd = 'test -b ' + self.RD_DEV_PATH
```
vermashi: Maybe there is a better way to do this.. you may want to combine the following two stack overflow solutions and make it a different disk_util function called "is_block_device".
https://stackoverflow.com/questions/82831/how-do-i-check-whether-a-file-exists-using-python
https://stackoverflow.com/questions/32682477/check-if-file-path-is-block-device-in-python

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+        return (int)(self.executor.Execute(cmd, suppress_logging=True)) == 0
+
+    def format_luks(self):
+        """ set up resource disk crypt device layer using disk util """
+        if not self.resource_disk_partition_exists():
+            self.logger.log('LUKS format operation requested, but resource disk partition does not exist')
+            return False
+        return (int)(self.disk_util.luks_format(passphrase_file=self.RD_KEY_FILE, dev_path=self.RD_DEV_PATH, header_file=None)) == 0
```
vermashi: Why not just do ( == CommonVariables.process_success ).
I don't like it any more than you do but there is something to be said for keeping code consistent.

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+        if not os.path.exists(self.RD_KEY_FILE):
+            self.disk_util.make_sure_path_exists(self.RD_KEY_FILE_MOUNT_POINT)
+            key_volume_device_name = os.popen('blkid -L "' + self.RD_KEY_VOLUME_LABEL + '"').read().strip()
+            self.disk_util.mount_filesystem(key_volume_device_name, self.RD_KEY_FILE_MOUNT_POINT)
+        return os.path.exists(self.RD_KEY_FILE)
+        
+    def mount(self):
+        """ mount the file system previously made on top of the crypt layer """
+        #ensure that resource disk mount point directory has been created
+        cmd = 'mkdir -p ' + self.RD_MOUNT_POINT
```
vermashi: the disk_util.mount_filesystem() call below already takes care of this.

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+        if self.executor.ExecuteInBash(cmd) != CommonVariables.process_success:
+            self.logger.log(msg="Failed to configure resource disk entry of /etc/fstab", level=CommonVariables.WarningLevel)
+            return False        
+        return True
+
+    def unmount_resource_disk(self):
+        """ return true if mount point is mounted regardless of crypt status """
+        # after service healing multiple unmounts of key file mount point may be required
+        self.disk_util.umount(self.RD_KEY_FILE_MOUNT_POINT)
+        self.disk_util.umount(self.RD_KEY_FILE_MOUNT_POINT)
```
vermashi: Not sure why the line is repeated.. Has it got to do with the comment right above?

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def prepare(self):
+        """ prepare a non-encrypted resource disk to be encrypted """
+        self.configure_waagent()
+        self.configure_fstab()
+        if self.resource_disk_partition_exists():
+            self.disk_util.swapoff()
+            self.unmount_resource_disk()
+            self.remove_device_mapper()
+            self.clear_luks_header()
```
vermashi: why do we clear the luks header again? Won't the luksFormat immediately afterwards take care of this?
Even if we do have to do it- why use urandom instead of /dev/zero

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+    def make(self):
+        """ make a default file system on top of the crypt layer """
+        make_result = self.disk_util.format_disk(dev_path=self.mapper_path, file_system=CommonVariables.default_file_system)
+        if make_result != CommonVariables.process_success:
+            self.logger.log(msg="Failed to make file system on ephemeral disk", level=CommonVariables.ErrorLevel)
+            return False
+        # todo - drop DATALOSS_WARNING_README.txt file to disk
+        return True
+
+    def mount_key_volume(self):
```
vermashi: the BekUtil.py already has this method (sort of).

VMEncryption/main/ResourceDiskUtil.py
ejarvi: Unfortunately, disk_util.mount_filesystem() ignores the return code of make_sure_path_exists.  I want this function to fail early and log the error if it is unable to create the directory.

VMEncryption/main/ResourceDiskUtil.py
ejarvi: get_bek_passphrase_file did not seem to be a close enough fit for this method and I wanted to include the operation in test coverage

VMEncryption/main/ResourceDiskUtil.py
ejarvi: OK, will update

VMEncryption/main/ResourceDiskUtil.py
ejarvi: Nice, investigating...

VMEncryption/main/ResourceDiskUtil.py
ejarvi: When VM is restored from a service healing scenario, although I do not yet understand why, a single unmount of /mnt/azure_bek_disk is not sufficient, however on a second call it is unmounted, which is required prior to the unmount of /mnt to succeed.

VMEncryption/main/ResourceDiskUtil.py
ejarvi: Wiping the header is done to prevent luksFormat from entering any logic that deals with choices of what to do with a pre-existing header.  For the wipe,  /dev/zero would also work, urandom was chosen as per the following example:  https://wiki.archlinux.org/index.php/Dm-crypt/Drive_preparation#Wipe_LUKS_header

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+            return False
+        return True
+
+    def configure_waagent(self):
+        """ turn off waagent.conf resource disk management  """ 
+        # set ResourceDisk.MountPoint to standard /mnt mount point
+        cmd = "sed -i.bak 's|ResourceDisk.MountPoint=.*|ResourceDisk.MountPoint=/mnt|' /etc/waagent.conf"
+
+        # set ResourceDiskFormat=n to ensure waagent does not attempt a simultaneous format
+        cmd = "sed -i.bak 's|ResourceDisk.Format=y|ResourceDisk.Format=n|' /etc/waagent.conf"
```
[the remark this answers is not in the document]
ejarvi: There do not seem to be any existing .bak backups of the waagent.conf file, however in other areas where this is done, a unique identifier is added to the extension so as to avoid conflicts or overwrites.  I will look into carrying over some of that logic here as well.

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -0,0 +1,274 @@
...
+        mount_result = self.disk_util.mount_filesystem(dev_path=self.mapper_path, mount_point=self.RD_MOUNT_POINT, file_system=CommonVariables.default_file_system)
+        if mount_result != CommonVariables.process_success:
+            self.logger.log(msg="Failed to mount file system on resource disk", level=CommonVariables.ErrorLevel)
+            return False
+        return True
+
+    def configure_waagent(self):
+        """ turn off waagent.conf resource disk management  """ 
+        # set ResourceDisk.MountPoint to standard /mnt mount point
+        cmd = "sed -i.bak 's|ResourceDisk.MountPoint=.*|ResourceDisk.MountPoint=/mnt|' /etc/waagent.conf"
```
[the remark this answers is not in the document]
ejarvi: this command removes lines containing either regex (doesn't set to /mnt)

VMEncryption/main/ResourceDiskUtil.py
ejarvi: There may be a better way but I am not seeing it yet even after reviewing these good links.  The test -b seems rather universal without requiring an additional library, and compatible with udev name ( /dev/disk/azure/resource-part1 in this case)

PR #518, opened 2018-01-30
OMS Linux Agent Extension - vmResourceID auto-injection

arjun-urs:
vmResourceId: auto-injection from azure metadata service when vmResourceId is not sent from the extension.

@NarineM


NarineM:
@arjun-urs , just realized that you forgot to make same changes to Fairfax version of omsagent.py. we maintain two different copies for vm extension. Can you update the FF version as well?

PR #524, opened 2018-02-13
OmsAgent OneClick fixes

shpimpal:
* Get vmResourceId from Metadata API if not provided in protected settings
 * Retry workspace provisioning when OMS/AMS returns 404


arjun-urs:
update Fairfax as well?

@NarineM

NarineM:
yes, please make the same changes for FF version of extension as well. Thanks Arjun for the catch!

shpimpal:
Updated Fairfax version.

PR #527, opened 2018-02-15
OMS Agent ASC - remove workspace configuration from the box on uninstall

arjun-urs:
- Setting the flag to remove workspace configuration on update
- Removing all the files and subfolders of workspace under the agent etc & var 

@NarineM


OmsAgent/Fairfax/omsagent_ff.py
```diff
@@ -343,6 +351,24 @@ def enable():
 
     return exit_code
 
+
+def remove_workspace_configuration():
```
NarineM: Please add a comment about why this workaround is needed. something like this .."this is needed to distinguish between extension removal vs extension upgrade and to workaround waagent upgrade routine calling remove on old version before calling upgrade on new extension version issue.

arjun-urs: Done.

PR #529, opened 2018-02-24
Code changes for VMSS reimage

vermashi:
1) We now stamp the VHDs which are encrypted instead of the OS disk
	2) We now don't do detached LUKS headers 
	3) We store the relevant line of azure_crypt_mount on a hidden folder on the encrypted device
        4)We scan all attached devices when the extension is run and try to see if there any encrypted devices that are not in the azure_crypt_mount. If there are such devices we extract the lines from them and add them to the azure_crypt_mount file.

### DIFF
--- a/VMEncryption/main/EncryptionSettingsUtil.py
+++ b/VMEncryption/main/EncryptionSettingsUtil.py
@@ -147,7 +147,7 @@ class EncryptionSettingsUtil(object):
 
         # Get all the currently encrypted items from the Azure Crypt Mount file (hopefully this has been consolidated by now)
         existing_crypt_items = disk_util.get_crypt_items()
-        existing_crypt_dev_items = self.get_disk_items_from_crypt_items(existing_crypt_items)
+        existing_crypt_dev_items = self.get_disk_items_from_crypt_items(existing_crypt_items, disk_util)
 
         all_device_items = existing_crypt_dev_items + extra_device_items
 
--- a/VMEncryption/main/EncryptionSettingsUtil.py
+++ b/VMEncryption/main/EncryptionSettingsUtil.py
@@ -63,11 +63,10 @@ class EncryptionSettingsUtil(object):
         # that contains only 0-9, a-z, A-Z, and - (the dash symbol).
         return str(uuid.uuid4())
 
-    def create_protector_file(self, protector_name):
+    def create_protector_file(self, existing_passphrase_file, protector_name):
         """create temporary protector file corresponding to protector name"""
-        src = CommonVariables.encryption_key_mount_point + '/' + CommonVariables.encryption_key_file_name
         dst = CommonVariables.encryption_key_mount_point + '/' + protector_name
-        copyfile(src, dst)
+        copyfile(existing_passphrase_file, dst)
         import ctypes
         libc = ctypes.CDLL("libc.so.6")
         libc.sync()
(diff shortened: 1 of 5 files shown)


VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -143,32 +172,54 @@ def dict_to_name_value_array(values):
...
+                        "DiskEncryptionKeyFileName": CommonVariables.encryption_key_file_name,
+                        "DiskEncryptionKeyEncryptionKeyURL": kek_url,
+                        "DiskEncryptionKeyEncryptionAlgorithm": kek_algorithm,
+                        "MachineName": machine_name})
+                    }]
+                })
+
         data = {
-            "DiskEncryptionDataVersion": "2.0",
+            "DiskEncryptionDataVersion": "3.0",
```
ejarvi: 2.0 is available in all regions, but I believe 3.0 rollout is still underway, maybe a comment here on availability would help

VMEncryption/main/handle.py
```diff
@@ -266,26 +289,13 @@ def update_encryption_settings():
...
-            settings.post_to_wireserver()
-            settings.remove_protector_file(new_protector_name)
+            stamp_disks_with_settings(new_device_items_about_to_get_encrypted=[], encryption_config=encryption_config)
 
             # commit local encryption config
-            encryption_config.passphrase_file_name = extension_parameter.DiskEncryptionKeyFileName
-            encryption_config.secret_id = new_protector_name
-            encryption_config.secret_seq_num = hutil.get_current_seq()
-            encryption_config.commit()
+            # encryption_config.passphrase_file_name = extension_parameter.DiskEncryptionKeyFileName
```
ejarvi: consider deleting these lines (295-298), they're still in version control if anyone needs to look back

VMEncryption/main/handle.py
```diff
@@ -380,11 +390,16 @@ def toggle_se_linux_for_centos7(disable):
...
-        resource_disk_util.automount()
-        logger.log("mounted encrypted resource disk")
+    resource_disk_util = ResourceDiskUtil(hutil, logger, DistroPatcher)
+    if encryption_config.config_file_exists():
+        volume_type = encryption_config.get_volume_type().lower()
+        if volume_type == CommonVariables.VolumeTypeData.lower() or volume_type == CommonVariables.VolumeTypeAll.lower():
+            resource_disk_util.automount()
+            logger.log("mounted encrypted resource disk")
+    else:
+        # Probably a re-image scenario: Just do a best effort
```
ejarvi: Why not use automount in the re-image scenario as well?

VMEncryption/main/handle.py
```diff
@@ -513,7 +515,7 @@ def enable():
             logger.log("Installing pre-requisites")
             DistroPatcher.install_extras()
 
-            extension_parameter = ExtensionParameter(hutil, logger, DistroPatcher, encryption_environment, protected_settings, public_settings)
+            extension_parameter = ExtensionParameter(hutil, logger, DistroPatcher, encryption_environment, get_protected_settings(), public_settings)
```
ejarvi: get_public_settings() too?

VMEncryption/main/EncryptionSettingsUtil.py
ejarvi: will address in next update

PR #539, opened 2018-03-14
Update bundle version to 1.4.4.210 and fix the remove workspace config on Uninstall

jamvar:
Update the bundle version from 1.4.1-123 to 1.4.4-210 and Fix the remove_workspace_configuraton method to work properly and run on uninstall except for 'upgrade'.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -39,14 +39,14 @@ except Exception as e:
 
 # Global Variables
 PackagesDirectory = 'packages'
-BundleFileName = 'omsagent-1.4.1-123.universal.x64.sh'
+BundleFileName = 'omsagent-1.4.4-210.universal.x64.sh'
 GUIDRegex = r'[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
 GUIDOnlyRegex = r'^' + GUIDRegex + '$'
 SCOMCertIssuerRegex = r'^[\s]*Issuer:[\s]*CN=SCX-Certificate/title=SCX' + GUIDRegex + ', DC=.*$'
 SCOMPort = 1270
 PostOnboardingSleepSeconds = 5
 InitialRetrySleepSeconds = 30
-CleanupWorkspaceConfiguration = False
+IsUpgrade = False
 
 # Paths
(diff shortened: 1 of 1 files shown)


jamvar:
Updated the access for 'IsUpdateWorkspaceConfiguration' global variable in main and uninstall methods.

OmsAgent/omsagent.py
```diff
@@ -39,14 +39,14 @@
...
-BundleFileName = 'omsagent-1.4.1-123.universal.x64.sh'
+BundleFileName = 'omsagent-1.4.4-210.universal.x64.sh'
 GUIDRegex = r'[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
 GUIDOnlyRegex = r'^' + GUIDRegex + '$'
 SCOMCertIssuerRegex = r'^[\s]*Issuer:[\s]*CN=SCX-Certificate/title=SCX' + GUIDRegex + ', DC=.*$'
 SCOMPort = 1270
 PostOnboardingSleepSeconds = 5
 InitialRetrySleepSeconds = 30
-CleanupWorkspaceConfiguration = False
+IsUpgradeWorkspaceConfiguration = False
```
NarineM: Can you rename it to be "IsUpgrade"?

OmsAgent/omsagent.py
```diff
@@ -356,13 +359,14 @@ def enable():
...
-    before calling 'upgrade' on new extension version issue. 
-    In upgrade case, we need workspace configuration to persist when in 
+    Its a workaround for waagent upgrade routine calling 'remove' on an old version
+    before calling 'upgrade' on new extension version issue.
+    In upgrade case, we need workspace configuration to persist when in
     remove case we need all the files be removed.
 
     This method will remove all the files/folders from the workspace path in Etc and Var.
     """
+    public_settings, _ = get_settings()
```
NarineM: why you need this line?

NarineM: Can you squash all your commits so that they appear as single commit in this PR?

jamvar: Yes. The method needs workspace Id from the public_settings but it's not given a method to access it since it not a global name.

OmsAgent/omsagent.py
jamvar: Will do

PR #559, opened 2018-04-10
Integrate 91Ade to singlepass master

ahmadmsft:


VMEncryption/main/handle.py
```diff
@@ -361,6 +361,9 @@ def update_encryption_settings():
 
 def update():
     hutil.do_parse_context('Upadate')
```
vermashi: Although the typo was already here, I'd appreciate it if you'd fix it

PR #566, opened 2018-04-19
Wrong parameter inside VMAccess

otubo:
Parameter 'Uninstall' was being passed to hutil.do_parse_context()
inside the function install().

Signed-off-by: Eduardo Otubo <otubo@redhat.com>


boumenot:
Thank you!

Issue #542, opened 2018-03-16
LAD crashes if control_omsagent fails with a non-ASCII error message


jasonzio:
Stack trace:
```
File "/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.111/diagnostic.py", line 301, in main
    start_mdsd(configurator)
File "/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.111/diagnostic.py", line 357, in start_mdsd
    dependencies_err, dependencies_msg = setup_dependencies_and_mdsd(configurator)
File "/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.111/diagnostic.py", line 155, in setup_dependencies_and_mdsd
    hutil.log, hutil.error)
File ["/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.111/Utils/omsagent_util.py", line 454](https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/Utils/omsagent_util.py#L454), in setup_omsagent
    cmd_exit_code, cmd_output = control_omsagent('restart', run_command)
File ["/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.111/Utils/omsagent_util.py", line 84](https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/Utils/omsagent_util.py#L84), in control_omsagent
    return 1, 'control_omsagent({0}) failed. Output: {1}'.format(op, cmd_output)

ERROR:UnicodeEncodeError: 'ascii' codec can't encode characters in position 227-229: ordinal not in range(128)
ERROR: [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.111] Daemon,error,0,Extension operation Daemon failed:'ascii' codec can't encode characters in position 227-229: ordinal not in range(128)
```
I don't know what the unencodable character is, but a general solution for "make the string that came back from run_command() into something that can safely be logged" would solve the problem.

nikita-rastogi:
The system locale for the language of error messages is LC_MESSAGES="en_US.UTF-8" on azure Vms.
can be checked by using command "locale" on the commandline. Whereas, the python 2.7 version uses ascii codec by default. Hence, the situation here isas follows,
[...]

PR #589, opened 2018-05-18
changed default encoding from ascii to utf-8 for the LAD process

nikita-rastogi:


nikita-rastogi:
issue: https://github.com/Azure/azure-linux-extensions/issues/542

Diagnostic/diagnostic.py
```diff
@@ -29,6 +29,10 @@
 import traceback
 import xml.etree.ElementTree as ET
 
+#https://github.com/Azure/azure-linux-extensions/issues/542 
+reload(sys)
```
jasonzio: Why this reload() here?

nikita-rastogi: As explained in these links: https://docs.python.org/2/library/sys.html#sys.setdefaultencoding , https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script 
the default encoding can be set during Python start up, when python evaluates the system wide environment. 
After this, the setdefaultencoding() method is removed from sys module. reload() forces the reloading of sys module and hence making the method available again. Thus, it offers a way to set default encoding even after python start up.

jasonzio: When every page on the internet says DON'T DO THIS, that seems like a strong recommendation to not do that. And you'll have to undo it when you move LAD to Python3 or make LAD version-agnostic, which you will need to do very soon.

If you insist on doing this, the reload(sys) and encoding change should be the very first thing you do, so that all imported modules see the new instance of the sys module with the changed default encoding.

nikita-rastogi: Hi @jasonzio, the reason changing default encoding is not recommended is because it will affect the behavior of string. The other way to fix the issue would have been to alter RunGetCommand() command executor in WALinuxAgent code to decode the response obtained by running bash shell using utf-8. and/or the encoding used by log file when the file is opened, this code also lies in WALinuxAgent.


As we are soon moving LAD to Python3 and making the change did not break LAD. 
Tested the change on Ubuntu and RHEL -> the logs were same as before and after the change (no error occured) and mdsd  and oms process were running. We concluded that this change can be considered to fix the issue till we migrate to Python 3.


I will shift the statement for changing default encoding  to the first line before any other import statement in diagnostic.py as suggested.

nikita-rastogi: Hi @jasonzio @abhijeetgaiha  , I tried to fix the codec issue with minimal impact on other pieces, i.e. without changing system wide default encoding. For this particular case the unicodeencodeerror got replaced by meaningful error messages by making following code change:
1. waagent.ToXml(): https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/Utils/omsagent_util.py#L440 
strEventsData+=strRecordNoQuoteFormat.format(attName,attValue,strMtStr) -> strEventsData+=strRecordNoQuoteFormat.format(attName,attValue.decode('utf-8'),strMtStr)

2. Diagnostics/Utils/omsagent_util.py: https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/Utils/omsagent_util.py#L440 
return 4, 'setup_omsagent(): Failed at configuring in_syslog. Exit code={0}, Output={1}'.format(cmd_exit_code,   cmd_output) -> return 4, 'setup_omsagent(): Failed at configuring in_syslog. Exit code={0}, Output={1}'.format(cmd_exit_code, cmd_output.encode('utf-8'))
**Remark: ToXml() is called while saving the event to xml file.

I am not sure if this will solve all the possible unicodeencodeerrors. But it will solve a subset for sure. Following steps can be taken to fix the issue in a more meaningful manner:
1. track down the message returned by waagent.RunGetOutput() and encode it using 'utf-8' codec wherever following syntax occurs: "".format(message_returned_by_RunGetOutput)

jasonzio: Your other solution would have been to build a wrapper around RunGetCommand() that, if you're running python2, does the correct utf-8 conversion; if you're running Python3, does the right thing for that environment. Basically, write the correct conversion code once, maintain it in one place, and change all the callers to call the new function.

boumenot: I am lurking, but I saw this and wanted to share. The Linux GA has to handle 2 and 3.  Here are a two examples where we deal with the pain.

https://github.com/Azure/WALinuxAgent/blob/0d04f47d149bb71cd3180a1d2d3ed04d096fe888/azurelinuxagent/common/future.py

https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/utils/textutil.py

abhijeetgaiha:
I've reverted this change now. There is another PR with a different approach to the fix: https://github.com/Azure/azure-linux-extensions/pull/593

PR #619, opened 2018-07-19
Format telemetry event message and success strings for JSON

kwirkykat:
This will fix DSC telemetry when an error occurs.
All new lines, tab characters, and quotes are now escaped from the message so that the message content will not affect the validity of the telemetry event JSON.
The operation status is also now ensured to be a lower case string since we were having some problems with the uppercase booleans in Python not being compatible with the lower case booleans of JSON.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/azure-linux-extensions/619)

### DIFF
--- a/OmsAgent/watcherutil.py
+++ b/OmsAgent/watcherutil.py
@@ -104,7 +104,10 @@ class Watcher:
     		]
             }}"""
 
-	return template.format(operation, operation_success, message, duration)
+        operation_success_as_string = str(operation_success).lower()
+        formatted_message = message.replace("\n", "\\n").replace("\t", "\\t").replace('"', '\\"')
+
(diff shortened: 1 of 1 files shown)


OmsAgent/watcherutil.py
```diff
@@ -104,7 +104,10 @@ def create_telemetry_event(self, operation, operation_success, message, duration
     		]
             }}"""
 
-	return template.format(operation, operation_success, message, duration)
+        operation_success_as_string = str(operation_success).lower()
+        formatted_message = message.replace("\n", "\\n").replace("\t", "\\t").replace('"', '\\"')
```
abhijeetgaiha: Shouldn't " be replaced with \" instead of \\" ?

kwirkykat: Ok will submit in another PR

PR #625, opened 2018-08-02
OMS agent self monitoring feature

shkhisti:
It adds two self monitoring features
1. Monitor heartbeat and specific error logs from oms agent logs
2. Monitor for resource usage of the agent

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -103,7 +103,7 @@ AutoManagedWorkspaceCreationSleepSeconds = 20
 
 # vmResourceId Metadata Service
 VMResourceIDMetadataHost = '169.254.169.254'
-VMResourceIDMetadataEndpoint = 'http://{0}/metadata/instance?api-version=2017-08-01'.format(VMResourceIDMetadataHost)
+VMResourceIDMetadataEndpoint = 'http://{0}/metadata/instance?api-version=2017-12-01'.format(VMResourceIDMetadataHost)
 
 # Change permission of log path - if we fail, that is not an exit case
 try:
@@ -431,7 +431,11 @@ def get_vmresourceid_from_metadata():
 
     try:
         response = json.loads(urllib2.urlopen(req).read())
-        return '/subscriptions/{0}/resourceGroups/{1}/providers/Microsoft.Compute/virtualMachines/{2}'.format(response['compute']['subscriptionId'],response['compute']['resourceGroupName'],response['compute']['name'])
+        if response['compute']['vmScaleSetName']:
+            return '/subscriptions/{0}/resourceGroups/{1}/providers/Microsoft.Compute/virtualMachineScaleSets/{2}'.format(response['compute']['subscriptionId'],response['compute']['resourceGroupName'],response['compute']['vmScaleSetName'])
+        else: 
+            return '/subscriptions/{0}/resourceGroups/{1}/providers/Microsoft.Compute/virtualMachines/{2}'.format(response['compute']['subscriptionId'],response['compute']['resourceGroupName'],response['compute']['name'])
+
     except urllib2.HTTPError as e:
         hutil_log_error('Request to Metadata service URL ' \
                         'failed with an HTTPError: {0}'.format(e))
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -30,6 +30,7 @@ import inspect
 import urllib
 import urllib2
 import watcherutil
+from threading import Thread
 
 try:
     from Utils.WAAgentUtil import waagent
@@ -217,7 +218,15 @@ def telemetry():
         f.write(str(py_pid) + '\n')
 
     watcher = watcherutil.Watcher(HUtilObject.error, HUtilObject.log, log_to_console=True)
-    watcher.watch()
+
+    watcher_thread = Thread(target = watcher.watch)
+    self_mon_thread = Thread(target = watcher.monitor_health)
+
+    watcher_thread.start()
+    self_mon_thread.start()
+
+    watcher_thread.join()
+    self_mon_thread.join()
 
 def dummy_command():
     """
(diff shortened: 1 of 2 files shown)


OmsAgent/omsagent.py
```diff
@@ -217,7 +218,15 @@ def telemetry():
         f.write(str(py_pid) + '\n')
 
     watcher = watcherutil.Watcher(HUtilObject.error, HUtilObject.log, log_to_console=True)
-    watcher.watch()
+
+    watcher_thread = Thread(target = watcher.watch)
+    self_mon_thread = Thread(target = watcher.monitor_health)
```
abhijeetgaiha: Why don't we do everything on the watcher thread? The health monitoring methods can be called from there too.

OmsAgent/watcherutil.py
```diff
@@ -74,55 +163,55 @@ def write_waagent_event(self, event):
...
         status_files = [
                 "/var/opt/microsoft/omsagent/log/ODSIngestion.status",
                 "/var/opt/microsoft/omsagent/log/ODSIngestionBlob.status",
                 "/var/opt/microsoft/omsagent/log/ODSIngestionAPI.status",
-                "/var/opt/microsoft/omsconfig/status/dscperformconsistency",
+                "/var/opt/microsoft/omsconfig/status/dscperformrequiredconfigurationchecks",
                 "/var/opt/microsoft/omsconfig/status/dscperforminventory",
-                "/var/opt/microsoft/omsconfig/status/dscsetlcm"
-            ]
+                "/var/opt/microsoft/omsconfig/status/dscsetdsclocalconfigurationmanager"
```
abhijeetgaiha: These file names should not be changed,

OmsAgent/watcherutil.py
```diff
@@ -162,3 +251,261 @@ def watch(self):
...
         pass
+
+    def monitor_heartbeat(self, self_mon_info, log_file_marker):
+        """
+            Monitor heartbeat health. OMS output plugin will update the timestamp
+            of new heartbeat file every 5 minutes. We will check if it is updated
+            If not, we will look into omsagent logs and look for specific error logs
+            which indicate we are in non recoverable state.             
+        """
+        heartbeat_file = '/var/opt/microsoft/omsagent/log/heartbeat'
```
abhijeetgaiha: The ODSIngestion.status file is already monitoring heartbeat indirectly since it uses the same channel. There isn't any need for adding another file just for the heartbeat,

OmsAgent/watcherutil.py
```diff
@@ -162,3 +251,261 @@ def watch(self):
...
+                self._hutil_error('Error in monitoring health of the omsagent. Exception={0}'.format(e))
+
+            finally:                        
+                time.sleep(sleepTime)                     
+
+    def take_corrective_action(self, self_mon_info):
+        """
+            Take a corrective action. 
+        """
+        run_command_and_log(RestartOMSAgentServiceCommand)
```
abhijeetgaiha: I don't remember seeing a lot of CRIs where simply restarting the agent fixes the issue. A better approach might be to check why the data isn't flowing. e.g. if the agent is not correctly onboarded, and try to onboard again. The ODSIngestion.status file will put out the right error message in this situation. if it's a transient error, then there isn't any need to restart.

OmsAgent/watcherutil.py
```diff
@@ -162,3 +251,261 @@ def watch(self):
...
+            We will also read oms agent logs to determine some error conditions.
+            We don't want to interfare with log watcher function.
+            So we will start this on a new thread.
+        """
+
+        self_mon_info = SelfMonitorInfo()
+        log_file_marker = LogFileMarker()
+        
+        # check every 6 minutes. we want to be bit pessimistic while looking for health, especially heartbeats which is emitted every 5 minutes.
+        sleepTime =  6 * 60
```
abhijeetgaiha: A 15 minute interval might be better here.

OmsAgent/omsagent.py
shkhisti: In my opinion telemetry and self monitoring threads have different responsibility and should be separate. One should not block other. Actualy i am even thinking of taking self monitoring our of watcherutil into separate python script , but will do that later. That will help me to put it in the bundle instead of extension.

OmsAgent/watcherutil.py
shkhisti: I think some mistake in my merge, i will correct those, thanks

OmsAgent/watcherutil.py
shkhisti: I want a signal which explicitly says that hertbeat is being emitted periodically. ODSIngestion status file does not tell me that with confidence. It can be updated when any other event is ingested. 
For scope of this change i need definiteve signal saying heartbeat is flowing end to end.

OmsAgent/watcherutil.py
shkhisti: Infact in contrary to this i see bunch of heartbeat cri's while i was oncall which were precisely due to fluentd being in bad shape.
There are two classes of CRI's this PR handles 
1. Heartbeat stops flowing randomly due to buffer corruption/disk full [false negative] / inode [false negatuve] . Fluentd restart helps in these cases as it clears buffers or any state causing this issue.
2. Fluentd suddently stops working/ does not proceed etc. 

In both these cases control never reaches out_ods.  Fluentd's buffering layer can not forward anything to us or it just stops working. 

Obviously there is other classes of issues which we can monitor for in future but out of scope of this PR.

OmsAgent/watcherutil.py
shkhisti: Heartbeat is sent every 5 minutes for sure with buffering it can take additional few seconds so 6 minutes should be more than enough. also it is better to catch the issues sooner than later.

OmsAgent/watcherutil.py
abhijeetgaiha: Got it. I wasn't aware of this class of issues. We might also consider an update to fluentd.

OmsAgent/watcherutil.py
abhijeetgaiha: Are you adding this file in the input plugin? If so, this is fine.

OmsAgent/omsagent.py
abhijeetgaiha: I was thinking along the lines of that every periodic activity carried out by the extension should be plugged into a unified framework. In any case, if you are satisfied with this approach, that's fine.

PR #630, opened 2018-08-17
ADE SinglePass fixes

vermashi:
Various fixed for ADE singlepass
 - transitioning fix
 - fix to disable encryption
 - fix to encryption substatus reporting
 - fix synchronization issue that was causing 500 errors from the wire server

### DIFF
--- a/VMEncryption/main/DiskUtil.py
+++ b/VMEncryption/main/DiskUtil.py
@@ -802,14 +802,13 @@ class DiskUtil(object):
         # the sets being disjoint would mean the candidate parent is not parent of any of the candidate children. So we return the opposite of that
         return not actual_children_dev_path_set.isdisjoint(children_dev_path_set)
 
-    def get_azure_data_disk_controller_and_lun_numbers(self, dev_items):
+    def get_all_azure_data_disk_controller_and_lun_numbers(self):
         """
         Return the controller ids and lun numbers for data disks that show up in the dev_items
         """
         list_devices = []
         azure_links_dir = '/dev/disk/azure'
 
-        dev_real_paths = set([os.path.realpath(self.get_device_path(di.name)) for di in dev_items])
         if not os.path.exists(azure_links_dir):
             return list_devices
 
@@ -831,8 +830,26 @@ class DiskUtil(object):
                         except ValueError:
                             # parsing will fail if "symlink" was a partition (e.g. "lun0-part1")
                             continue # so just ignore it
-                    if self.is_parent_of_any(os.path.realpath(full_path), dev_real_paths):
                         list_devices.append((controller_id, lun_number))
+        return list_devices
+
+    def get_azure_data_disk_controller_and_lun_numbers(self, dev_items):
+        """
+        Return the controller ids and lun numbers for data disks that show up in the dev_items
+        """
+
+        all_controller_and_lun_numbers = self.get_all_azure_data_disk_controller_and_lun_numbers()
+        dev_real_paths = set([os.path.realpath(self.get_device_path(di.name)) for di in dev_items])
+
+        list_devices = []
+        azure_links_dir = '/dev/disk/azure'
+
+        for controller_id, lun_number in all_controller_and_lun_numbers:
+            scsi_dir = os.path.join(azure_links_dir, 'scsi' + str(controller_id))
+            symlink = os.path.join(scsi_dir, 'lun' + str(lun_number))
+            if self.is_parent_of_any(os.path.realpath(symlink), dev_real_paths):
+                list_devices.append((controller_id, lun_number))
+
         return list_devices
 
     def get_device_items_sles(self, dev_path):
--- a/VMEncryption/main/DiskUtil.py
+++ b/VMEncryption/main/DiskUtil.py
@@ -246,50 +246,6 @@ class DiskUtil(object):
                 items = add_child(items, child)
         return items
 
-
-    def get_azure_vhd_dev_items(self, device_items):
-        """
-        Returns all the underlying azure-vhd level device items for given device items (that may be "children" of the vhds)
-
-        Gets a list of dev_items say ("/dev/sdd","/dev/sdc1","/dev/mapper/vg0-lv0")
-        And figures out which Azure VHDs are covered by these device items and returns their dev_items (in the below example: [/dev/sdb, /dev/sdc, /dev/sdd])
-
-        In order to do this we construct an lsblk tree object using get_lsblk_tree method to identify who is who's "child".
-        """
-        lsblk_tree = self.get_lsblk_tree()
-
-        # Let's set up a set of realpaths of each dev_items's real_path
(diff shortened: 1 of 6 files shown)


VMEncryption/main/DiskUtil.py
```diff
@@ -397,44 +397,44 @@ def get_crypt_items(self):
...
+            if not crypt_item.dev_path:
+                raise Exception("Could not locate block device for rootfs")
 
-                crypt_item.luks_header_path = "/boot/luks/osluksheader"
+            crypt_item.luks_header_path = "/boot/luks/osluksheader"
 
-                if not os.path.exists(crypt_item.luks_header_path):
-                    crypt_item.luks_header_path = crypt_item.dev_path
+            if not os.path.exists(crypt_item.luks_header_path):
+                crypt_item.luks_header_path = crypt_item.dev_path
```
mayank88mahajan: without the osluksheader how are you updating the header path to dev path? Is this for in-place headers?

VMEncryption/main/DiskUtil.py
```diff
@@ -397,44 +397,44 @@ def get_crypt_items(self):
 
                     crypt_items.append(crypt_item)
 
-            encryption_status = json.loads(self.get_encryption_status())
+        encryption_status = json.loads(self.get_encryption_status())
 
-            if encryption_status["os"] == "Encrypted" and not rootfs_crypt_item_found:
-                crypt_item = CryptItem()
-                crypt_item.mapper_name = "osencrypt"
+        if encryption_status["os"] == "Encrypted" and not rootfs_crypt_item_found:
```
mayank88mahajan: Can you write a detailed comment on the logic below? Difficult to understand just reading the code :)

mayank88mahajan:
>         def controller_id_and_lun_to_settings_data(scsi_controller, lun_number):

alignment from previous def: is off

---
Refers to: VMEncryption/main/EncryptionSettingsUtil.py:180 in 6c83ce5. [](commit_id = 6c83ce550984d75dd6b223239b5179cdfc0f81b4, deletion_comment = False)

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -259,11 +262,42 @@ def post_to_wireserver(self, data):
...
+                    "SecretTags": self._dict_to_name_value_array({
+                        "DiskEncryptionKeyFileName": "nullProtector.bek",
+                        "MachineName": machine_name})
+                    }]
+                }
+
+        data_disks_settings_data = [ controller_id_and_lun_to_settings_data(scsi_controller, lun_number)
+                                    for (scsi_controller, lun_number) in data_disk_controller_ids_and_luns]
+
+        data = {"DiskEncryptionDataVersion": "3.0",
```
mayank88mahajan: >"3.0" [](start = 45, length = 5)

maybe make this as a constant variable: current_data_version

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -259,11 +262,42 @@ def post_to_wireserver(self, data):
...
+                "Volumes": [{
+                    "VolumeType": "DataVolume",
+                    "ProtectorFileName": "nullProtector.bek",
+                    "SecretTags": self._dict_to_name_value_array({
+                        "DiskEncryptionKeyFileName": "nullProtector.bek",
+                        "MachineName": machine_name})
+                    }]
+                }
+
+        data_disks_settings_data = [ controller_id_and_lun_to_settings_data(scsi_controller, lun_number)
```
mayank88mahajan: what about OS disk on IDE controller?

VMEncryption/main/handle.py
```diff
@@ -188,16 +190,11 @@ def stamp_disks_with_settings(new_device_items_about_to_get_encrypted, os_item_t
...
         extra_device_items=items_to_encrypt,
         disk_util=disk_util)
-    try:
-        settings.post_to_wireserver(data)
-    except:
-        # if host does not support 3.0 protocol version, retry with 2.0
-        data["DiskEncryptionDataVersion"] = "2.0"
-        settings.post_to_wireserver(data)
+
+    settings.post_to_wireserver(data)
```
mayank88mahajan: try-catch the error from this code path to throw a meaningful error instead of a stack trace?

VMEncryption/main/handle.py
```diff
@@ -211,7 +208,8 @@ def stamp_disks_with_settings(new_device_items_about_to_get_encrypted, os_item_t
     encryption_config.secret_seq_num = hutil.get_current_seq()
     encryption_config.commit()
 
-
+def is_stamped_with_current_config(encryption_config):
```
mayank88mahajan: what is stamped with current config? Rename function to reflect that.

VMEncryption/main/DiskUtil.py
vermashi: I am just fixing indentation for some pre-existing code here- so I'm not sure what's going on here either but I'd also guess the same. This is probably assuming that if the header is detached and not located at osluksheader then its attached.

VMEncryption/main/DiskUtil.py
vermashi: I'll try to parse and write it out.

VMEncryption/main/DiskUtil.py
```diff
@@ -1028,6 +1033,25 @@ def get_azure_data_disk_controller_and_lun_numbers(self, vhd_dev_items):
                         list_devices.append((controller_id, lun_number))
         return list_devices
 
+    def get_azure_data_disk_controller_and_lun_numbers(self, vhd_dev_items):
+        """
+        Return the controller ids and lun numbers for data disks that show up in the vhd_dev_items
+        """
+
+        all_controller_and_lun_numbers = self.get_all_azure_data_disk_controller_and_lun_numbers()
```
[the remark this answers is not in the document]
vermashi: I think the comment is a little misplaced. Which tokens do you want me to refactor out to the constants file?
I'm guessing '/dev/disk/azure', 'scsi' and 'lun'

VMEncryption/main/EncryptionSettingsUtil.py
vermashi: We don't allow disable when the OS is encrypted so I left it out. But, on second thought I will add that in - in case somebody somehow encrypts a partition of the OS disk with volume type data.

VMEncryption/main/EncryptionSettingsUtil.py
vermashi: Will do.

VMEncryption/main/handle.py
vermashi: The Error gets caught upstream and the context is reflected there.
For example if we ran into this while doing UpdateEncryptionSettings or during Data disk encryption

vermashi:
Rebased on top of the lsblk fix so there will be some new changes for that as well

PR #636, opened 2018-08-28
Python2/Python3 support for VMAccess extension for Linux

vrdmr:
Minor fixes in VMAccess and WAAgent to support Python2 and 3.

### DIFF
--- a/Common/WALinuxAgent-2.0.16/waagent
+++ b/Common/WALinuxAgent-2.0.16/waagent
@@ -57,8 +57,7 @@ elif sys.version_info[0] == 2:
     import httplib as httpclient
     from urlparse import urlparse
 
-
-if not hasattr(subprocess,'check_output'):
+if not hasattr(subprocess, 'check_output'):
     def check_output(*popenargs, **kwargs):
         r"""Backport from subprocess module from python 2.7"""
         if 'stdout' in kwargs:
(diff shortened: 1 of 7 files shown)


Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -2333,94 +2396,101 @@ def Ord(a):
...
 def AppendFileContents(filepath, contents):
     """
     Append 'contents' to 'filepath'.
     """
-    if type(contents) == str :
-        contents=contents.encode('latin-1')
-    try: 
-        with open(filepath, "a+") as F :
+    if type(contents) == str:
+        contents = contents.encode('latin-1').decode('latin-1')
```
vrdmr: The only real fix.

VMAccess/vmaccess.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
boumenot: The final commit should keep this as python not python3.  We don't expect distros that have moved completely to Python3 to keep the binary as python3 and/or they will introduce a symlink for python that refers to python3.

vrdmr: Sure. I'll revert this once the Tests work.

PR #640, opened 2018-09-11
vmaccess: use default username regex

boumenot:


VMAccess/vmaccess.py
```diff
@@ -106,7 +105,7 @@ def enable():
                                           op=waagent.WALAEventOperation.Enable,
                                           isSuccess=False,
                                           message="(03002)Argument error, invalid remove_user")
-                raise Exception("invalid remove_user")
+                raise Exception("'remove_user' does not match the regular expression '^[a-z][-a-z0-9_]*$'")
```
hglkrijger: consider `'remove_user' username does not match...`

PR #646, opened 2018-09-21
persist oms agent id across extension upgrades

shkhisti:


shkhisti:
Extension framework ises update irrespective of oir manifest setting.



Regards,
Shrirang


-------- Original message --------

OmsAgent/omsagent.py
```diff
@@ -251,6 +277,9 @@ def install():
     workspaceId = public_settings.get('workspaceId')
     check_workspace_id(workspaceId)
 
+    # Take the backup of the state for given workspace. 
+    restore_state(workspaceId)
```
shpimpal: Shouldn't we restore in remove() too ? As the guest agent currently calls remove followed by install to update/upgrade the extension.

PR #649, opened 2018-09-22
Adding a shell layer to find python version on the box

vrdmr:
Added a new Shell script to find the correct python version and invoke the extension script.

FYI: @narrieta, @boumenot, @roiyz-msft.

### DIFF
--- a/VMAccess/HandlerManifest.json
+++ b/VMAccess/HandlerManifest.json
@@ -2,11 +2,11 @@
   {
     "version": 1.0,
     "handlerManifest": {
-      "disableCommand": "./vmaccess.py -disable",
-      "enableCommand": "./vmaccess.py -enable",
-      "installCommand": "./vmaccess.py -install",
-      "uninstallCommand": "./vmaccess.py -uninstall",
-      "updateCommand": "./vmaccess.py -update",
+      "disableCommand": "extension_shim.sh -c ./vmaccess.py -d",
+      "enableCommand": "extension_shim.sh -c ./vmaccess.py  -e",
+      "installCommand": "extension_shim.sh -c ./vmaccess.py  -i",
+      "uninstallCommand": "extension_shim.sh -c ./vmaccess.py -u",
+      "updateCommand": "extension_shim.sh -c ./vmaccess.py -p",
       "rebootAfterInstall": false,
       "reportHeartbeat": false
     }
--- a/VMAccess/manifest.xml
+++ b/VMAccess/manifest.xml
@@ -2,7 +2,7 @@
 <ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
   <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
   <Type>VMAccessForLinux</Type>
-  <Version>1.5.0</Version>
+  <Version>1.5.1</Version>
   <Label>Microsoft Azure VM Access Extension for Linux Virtual Machines</Label>
   <HostingResources>VmRole</HostingResources>
   <MediaLink></MediaLink>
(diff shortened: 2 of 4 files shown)


VMAccess/vmaccess.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
```
narrieta: If this approach is going to be used by other extensions, consider making this a more generic script that takes as argument the python script (with its arguments) that needs to be executed.

VMAccess/vmaccess.sh
```diff
@@ -0,0 +1,109 @@
...
+    -d|--disable    disable the extension
+    -e|--enable     enable the extension
+    -p|--update     update the extension
+"
+
+function find_python(){
+    local python_version=$1
+
+    # Check if there is python defined.
+    which python > /dev/null 2>&1
```
narrieta: There are some posts (e.g. https://stackoverflow.com/questions/592620/how-to-check-if-a-program-exists-from-a-bash-script) discouraging the use of which to check for the existence of a command; apparently some implementations don't set an exit code. Consider using command instead:
```

if command -v python; then
...
fi
```

VMAccess/vmaccess.sh
```diff
@@ -0,0 +1,109 @@
...
+    return_code=`echo $?`
+
+    if [[ ${return_code} == 0 ]]
+    then
+        eval ${python_version}="python"
```
narrieta: The name of the variable 'python_version' is a little misleading... 'python_command' maybe?

VMAccess/vmaccess.sh
```diff
@@ -0,0 +1,109 @@
...
+            exit 1
+            ;;
+    esac
+done
+
+shift $((OPTIND-1))
+
+find_python PYTHON
+
+if [ -z "$PYTHON" ]
```
narrieta: find_python is also doing an 'exit 51'; we should probably remove that one

VMAccess/vmaccess.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
+
+COMMAND="./vmaccess.py"
+PYTHON=""
+
+USAGE="$(basename "$0") [-h] [-i|--install] [-u|--uninstall] [-d|--disable] [-e|--enable] [-p|--update] -- program to invoke the extension script
```
narrieta: consider moving the description of the script "program to invoke the extension script" to another line to separate it from the syntax

VMAccess/manifest.xml
```diff
@@ -2,7 +2,7 @@
 <ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
   <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
   <Type>VMAccessForLinux</Type>
-  <Version>1.5.0</Version>
+  <Version>1.6.0</Version>
```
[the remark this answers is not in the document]
vrdmr: Done.

VMAccess/vmaccess.sh
vrdmr: Good idea. I added a new `-c` flag where one can give the python script. Please take a look at the update `$USAGE` for its usage.

narrieta:
thanks for the updates, lgtm

VMAccess/HandlerManifest.json
```diff
@@ -2,11 +2,11 @@
   {
     "version": 1.0,
     "handlerManifest": {
-      "disableCommand": "./vmaccess.py -disable",
-      "enableCommand": "./vmaccess.py -enable",
-      "installCommand": "./vmaccess.py -install",
-      "uninstallCommand": "./vmaccess.py -uninstall",
-      "updateCommand": "./vmaccess.py -update",
+      "disableCommand": "./extension_shim.sh -c ./vmaccess.py -d",
```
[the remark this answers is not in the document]
vrdmr: okay. I'll change this. Thanks.

PR #650, opened 2018-09-24
CustomScriptExtension shim layer for python interpreter

Diastro:
- Adding shim layer for CSE
- Bumping version to 1.5.4

### DIFF
--- a/CustomScript/HandlerManifest.json
+++ b/CustomScript/HandlerManifest.json
@@ -2,11 +2,11 @@
   {
     "version": 1.0,
     "handlerManifest": {
-      "disableCommand": "./shim.sh -disable",
-      "enableCommand": "./shim.sh -enable",
-      "installCommand": "./shim.sh -install",
-      "uninstallCommand": "./shim.sh -uninstall",
-      "updateCommand": "./shim.sh -update",
+      "disableCommand": "shim.sh -disable",
+      "enableCommand": "shim.sh -enable",
+      "installCommand": "shim.sh -install",
+      "uninstallCommand": "shim.sh -uninstall",
+      "updateCommand": "shim.sh -update",
       "rebootAfterInstall": false,
       "reportHeartbeat": false
     }
--- a/CustomScript/HandlerManifest.json
+++ b/CustomScript/HandlerManifest.json
@@ -2,11 +2,11 @@
   {
     "version": 1.0,
     "handlerManifest": {
-      "disableCommand": "customscript.py -disable",
-      "enableCommand": "customscript.py -enable",
-      "installCommand": "customscript.py -install",
-      "uninstallCommand": "customscript.py -uninstall",
-      "updateCommand": "customscript.py -update",
+      "disableCommand": "./shim.sh -disable",
+      "enableCommand": "./shim.sh -enable",
+      "installCommand": "./shim.sh -install",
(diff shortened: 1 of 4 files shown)


CustomScript/customscript.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
boumenot: This script needs to call through the shim.  Python3 is not available on all distros, and the intent is to make customscript.py portable to Python2 only or Python3 only distros.

I don't think this change is necessary at all because shim.sh is executing customscript.py directly.

CustomScript/customscript.py
```diff
@@ -236,14 +236,14 @@ def download_files(hutil):
 def start_daemon(hutil):
     cmd = get_command_to_execute(hutil)
     if cmd:
-        args = [os.path.join(os.getcwd(), __file__), "-daemon"]
+        args = [os.path.join(os.getcwd(), "shim.sh"), "-daemon"]
```
boumenot: Good catch!

CustomScript/HandlerManifest.json
```diff
@@ -2,11 +2,11 @@
   {
     "version": 1.0,
     "handlerManifest": {
-      "disableCommand": "customscript.py -disable",
-      "enableCommand": "customscript.py -enable",
-      "installCommand": "customscript.py -install",
-      "uninstallCommand": "customscript.py -uninstall",
-      "updateCommand": "customscript.py -update",
+      "disableCommand": "./shim.sh -disable",
```
boumenot: I would drop the use of ./.  The agent constructs the path as necessary, so there's no need to be explicit.

CustomScript/shim.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# The shim scripts provide a single entry point for CSE and will invoke the customscript.py entry point using the
+# appropriate python interpreter version.
+# Arguments passed to the shim layer are redirected to the invoked script without any validation.
+
+COMMAND="./customscript.py"
+PYTHON=""
+ARG="$@"
+
+function find_python(){
```
vrdmr: I just cleaned up the [shim ](https://github.com/Azure/azure-linux-extensions/pull/649/files#diff-8e0deab0c440b69b28228e76a0fdf02fR34). You should try the cleaner `find_python`.

Just FYI - It also supports commands directly now.
```
$ bash extension_shim.sh -c ""hello.py --install""
python hello.py --install
```

CustomScript/customscript.py
vrdmr: Is `python3` purposefully there?

Diastro: This was a mistake (I was testing something else).
Removed!

Diastro: This was a mistake (I was testing something else).
Removed!

CustomScript/HandlerManifest.json
Diastro: Oh perfect. Will remove. Thanks

CustomScript/shim.sh
Diastro: Done! Thanks for the help with this script!

boumenot:
Please run it through DCR.  I can show you an easier way to do this for Python 2 and 3 without having to do a redeploy of the extension.

PR #655, opened 2018-09-26
ADE SP QOS Improvements

ejarvi:
- move time consuming operations to enable 
- remove dead code and unnecessary packages

### DIFF
--- a/VMEncryption/main/Utils/HandlerUtil.py
+++ b/VMEncryption/main/Utils/HandlerUtil.py
@@ -154,9 +154,13 @@ class HandlerUtility:
             sys.exit(0)
 
     def log(self, message):
+        # write message to stderr for inclusion in QOS telemetry
+        sys.stderr.write(message)
         self._log(self._get_log_prefix() + ': ' + message)
 
     def error(self, message):
+        # write message to stderr for inclusion in QOS telemetry
+        sys.stderr.write(message)
         self._error(self._get_log_prefix() + ': ' + message)
 
     def _parse_config(self, ctxt):
--- a/VMEncryption/main/patch/UbuntuPatching.py
+++ b/VMEncryption/main/patch/UbuntuPatching.py
@@ -64,7 +64,16 @@ class UbuntuPatching(AbstractPatching):
         cmd = " ".join(['apt-get', 'update'])
         self.command_executor.Execute(cmd)
 
-        packages = ['at', 'cryptsetup-bin', 'lsscsi', 'python-six', 'python-parted', 'procps', 'psmisc', 'libssl-dev', 'libffi-dev', 'python-dev']
+        packages = ['at',
+					'cryptsetup-bin',
+					'lsscsi',
+					'python-six',
+					'python-parted',
+					'procps',
+					'psmisc',
+					'libssl-dev',
+					'libffi-dev',
+					'python-dev']
 
         cmd = " ".join(['apt-get', 'install', '-y'] + packages)
         self.command_executor.Execute(cmd)
--- a/VMEncryption/main/patch/UbuntuPatching.py
+++ b/VMEncryption/main/patch/UbuntuPatching.py
@@ -64,7 +64,7 @@ class UbuntuPatching(AbstractPatching):
         cmd = " ".join(['apt-get', 'update'])
         self.command_executor.Execute(cmd)
 
-        packages = ['at', 'cryptsetup-bin', 'lsscsi', 'python-six', 'python-parted', 'procps', 'psmisc', 'gcc', 'libssl-dev', 'libffi-dev', 'python-dev']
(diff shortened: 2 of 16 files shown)


VMEncryption/main/patch/centosPatching.py
```diff
@@ -74,39 +74,6 @@ def __init__(self, logger, distro_info):
...
 
     def install_extras(self):
-        epel_packages_installed = False
-        attempt = 0
-
-        while not epel_packages_installed:
-            attempt += 1
-            self.logger.log("Attempt #{0} to locate EPEL packages".format(attempt))
-            if self.distro_info[1].startswith("6."):
-                if self.command_executor.Execute("rpm -q ntfs-3g python-pip"):
```
mayank88mahajan: >python-pip [](start = 65, length = 10)

Is python-pip no longer required?

VMEncryption/main/patch/centosPatching.py
```diff
@@ -74,39 +74,6 @@ def __init__(self, logger, distro_info):
...
-                        self.command_executor.Execute(epel_cmd)
-
-                    self.command_executor.Execute("yum install -y ntfs-3g python2-pip")
-
-                    if not self.command_executor.Execute("rpm -q ntfs-3g python2-pip"):
-                        epel_packages_installed = True
-                else:
-                    epel_packages_installed = True
-
         packages = ['cryptsetup',
```
mayank88mahajan: do we understand all the packages and there need here?

VMEncryption/main/patch/UbuntuPatching.py
```diff
@@ -64,7 +64,7 @@ def install_extras(self):
         cmd = " ".join(['apt-get', 'update'])
         self.command_executor.Execute(cmd)
 
-        packages = ['at', 'cryptsetup-bin', 'lsscsi', 'python-six', 'python-parted', 'procps', 'psmisc', 'gcc', 'libssl-dev', 'libffi-dev', 'python-dev', 'python-pip']
+        packages = ['at', 'cryptsetup-bin', 'lsscsi', 'python-six', 'python-parted', 'procps', 'psmisc', 'libssl-dev', 'libffi-dev', 'python-dev']
```
mayank88mahajan: change formatting to have package names on new lines like other distro code.

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -154,9 +154,13 @@ def exit_if_same_seq(self, exit_status=None):
             sys.exit(0)
 
     def log(self, message):
+        # write message to stderr for inclusion in QOS telemetry
+        sys.stderr.write(message)
```
mayank88mahajan: > sys.stderr.write(message) [](start = 7, length = 26)

do these get captured in the pivoted root state as well? Would be great to know how to view these logs on the backend - Kusto?

VMEncryption/main/patch/centosPatching.py
ejarvi: python-pip was used to install adal which is no longer required by the single pass extension

VMEncryption/main/patch/centosPatching.py
ejarvi: - ntfs-3g package is not required since the key volume is always vfat 
- python-pip package was used SUSE patching to install JWT in the ADAL scenario no longer required 
- epel was added because it includes dependencies required by python-pip which is no longer required 
- some code comments explaining the inclusion of other packages still being included would be helpful, I will work on including that.

ejarvi: - lsscsi provides scsi device identifier information 
- psmisc provides utilities like fuser to identify what processes are using files 
- cryptsetup-reencrypt provides dm-crypt utilities 
- lvm2 provides lvm support
- uuid Simple, fast generation of RFC4122 UUIDS
- at for command scheduling
- patch provides utility for patching (may now be candidate for removal post Dracut design change)
- procps-ng provides utilities for browsing procfs and entries in process table
- util-linux standard package including utilities like blkid and lsblk 
- python-six wraps differences between python 2 and 3 
- pyparted includes python bindings for GNU parted 
- libffi-devel provides high level programming interface to various calling conventions

VMEncryption/main/Utils/HandlerUtil.py
ejarvi: Guest agent monitors stderr.  Prior to this change no output was written to stderr.   If the process is running under guest agent as parent process, stderr will be captured, reported to host agent, and included in telemetry.  If the process is started as a daemon and running outside of guest agent the output may not fall within this capture chain.

VMEncryption/main/patch/UbuntuPatching.py
ejarvi: reformatted

PR #656, opened 2018-09-29
ADE DP QOS Improvements

ejarvi:
ADE Dual Pass QOS improvements in this PR include: 

- move extra package installs to the enable stage to eliminate risk of install timeouts 
- remove unused NTFS package and related calls since key volume is always vfat 
- add stderr logging for better diagnostics as a stepping stone to future QOS improvements 
- include active directory authentication library and dependencies still required for dual pass KV ops

### DIFF
--- a/VMEncryption/main/BekUtil.py
+++ b/VMEncryption/main/BekUtil.py
@@ -57,7 +57,7 @@ class BekUtil(object):
 
         for azure_device in azure_devices:
             fstype = str(azure_device.file_system).lower()
-            if fstype in ['vfat', 'ntfs']:
+            if fstype in ['vfat']:
                 try:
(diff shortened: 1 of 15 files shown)


mayank88mahajan:
>                 """

This section will be called before the daemon starts up if I am not wrong. This requires communicating with KeyVault using adal and will need to be installed first.

---
Refers to: VMEncryption/main/handle.py:636 in f3b2037. [](commit_id = f3b20378c2a277d075c6659ebf24e14bc3cde350, deletion_comment = False)

mayank88mahajan:
>                 update_encryption_settings()

same comment as SP encryption PR.

---
Refers to: VMEncryption/main/handle.py:511 in f3b2037. [](commit_id = f3b20378c2a277d075c6659ebf24e14bc3cde350, deletion_comment = False)

PR #676, opened 2018-10-23
Add input validation to the DP extension as well

vermashi:


VMEncryption/main/handle.py
```diff
@@ -473,38 +461,41 @@ def enable():
...
-
-        logger.log('Public settings:\n{0}'.format(json.dumps(public_settings, sort_keys=True, indent=4)))
-
         encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
 
         if encryption_operation in [CommonVariables.EnableEncryption, CommonVariables.EnableEncryptionFormat, CommonVariables.EnableEncryptionFormatAll]:
             logger.log("handle.py found enable encryption operation")
 
-            extension_parameter = ExtensionParameter(hutil, logger, DistroPatcher, encryption_environment, protected_settings, public_settings)
+            extension_parameter = ExtensionParameter(hutil, logger, DistroPatcher, encryption_environment, get_protected_settings(), public_settings)
```
ejarvi: get_public_settings() too?

vermashi: The variable public settings was populated a bit ago around line 475(old), 464(new)

VMEncryption/main/Common.py
```diff
@@ -46,9 +46,13 @@ class CommonVariables:
     PassphraseFileNameKey = 'BekFileName'
     KeyEncryptionKeyURLKey = 'KeyEncryptionKeyURL'
     KeyVaultURLKey = 'KeyVaultURL'
+    KeyVaultResourceIdKey = 'KeyVaultResourceId'
+    KekVaultResourceIdKey = 'KekVaultResourceId'
     AADClientIDKey = 'AADClientID'
     AADClientCertThumbprintKey = 'AADClientCertThumbprint'
     KeyEncryptionAlgorithmKey = 'KeyEncryptionAlgorithm'
+    encryption_algorithms = ['RSA-OAEP', 'RSA-OAEP-256', 'RSA1_5']
```
ejarvi: What is origin of this list?  I know they were provided as example values [here](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/azure-disk-enc-linux#property-values) but I am not sure that list is complete.     The [SDL](https://docs.microsoft.com/en-us/azure/security/azure-security-threat-modeling-tool-cryptography#padding ) documentation lists RSA-KEM as approved but is missing from this list, and RSA 1.5 on the way out.

VMEncryption/main/check_util.py
```diff
@@ -90,7 +94,83 @@ def is_invalid_lvm_os(self):
                     detected = True
         return detected
 
-    def is_precheck_failure(self):
+    def check_kv_url(self, test_url, message):
+        """basic sanity check of the key vault url"""
+        expected = "https://{keyvault-name}.vault.azure.net/"
```
ejarvi: update for national clouds  (sorry if duplicate of other review comments)

VMEncryption/main/check_util.py
```diff
@@ -90,7 +94,83 @@ def is_invalid_lvm_os(self):
...
+        if test_url is None:
+            raise Exception(message + '\nNo URL supplied')
+
+        try:
+            parse_result = urlparse.urlparse(test_url)
+        except:
+            raise Exception(message + '\nMalformed URL: ' + test_url)
+
+        if not (parse_result.scheme.lower() == "https" and \
+                re.match('[a-zA-Z0-9\-]+\.vault\.azure\.net(:443)?$', parse_result.netloc)):
```
ejarvi: update for national clouds

VMEncryption/main/check_util.py
```diff
@@ -90,7 +94,83 @@ def is_invalid_lvm_os(self):
...
+        id_splits = test_id.lower().split('/')
+
+        if not (len(id_splits) >= 9 and \
+                id_splits[0] == "" and \
+                id_splits[1] == "subscriptions" and \
+                id_splits[2] != "" and \
+                id_splits[3] == "resourcegroups" and \
+                id_splits[4] != "" and \
+                id_splits[5] == "providers" and \
+                id_splits[6] == "microsoft.keyvault" and \
```
ejarvi: this is a provider name, I'm not sure if lower case is valid or not, I have never seen it lower cased

VMEncryption/main/check_util.py
```diff
@@ -90,7 +94,83 @@ def is_invalid_lvm_os(self):
...
+
+    def check_kv_id(self, test_id, message):
+        """basic sanity check of the key vault id"""
+        # more strict checking would validate the full key vault id format
+        expected = "/subscriptions/{subid}/resourceGroups/{rgname}/providers/Microsoft.KeyVault/vaults/{vaultname}"
+
+        if test_id is None:
+            raise Exception(message + '\nNo Resource ID supplied')
+
+        id_splits = test_id.lower().split('/')
```
ejarvi: regex might be cleaner and more reliable

VMEncryption/main/check_util.py
```diff
@@ -90,7 +94,83 @@ def is_invalid_lvm_os(self):
...
+
+        if not (len(id_splits) >= 9 and \
+                id_splits[0] == "" and \
```
ejarvi: this only checks for non-empty, it does not validate that various naming conventions are being met 
https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions

VMEncryption/main/check_util.py
```diff
@@ -90,7 +94,83 @@ def is_invalid_lvm_os(self):
...
+        kv_id = public_settings.get(CommonVariables.KeyVaultResourceIdKey)
+        kek_kv_id = public_settings.get(CommonVariables.KekVaultResourceIdKey)
+        kek_algorithm = public_settings.get(CommonVariables.KeyEncryptionAlgorithmKey)
+
+        self.check_kv_url(kv_url, "Key Vault URL is required, but was missing or invalid")
+        self.check_kv_id(kv_id, "Key Vault ID is required, but was missing or invalid")
+        if kek_url:
+            self.check_kv_url(kek_url, "A KEK URL was specified, but was invalid")
+            self.check_kv_id(kek_kv_id, "A KEK URL was specified, but its KeyVault ID was invalid")
+            if kek_algorithm not in CommonVariables.encryption_algorithms:
```
ejarvi: This will enforce case sensitive check of the algorithm name (the whitelist consists of all upper case strings).  If KV is not case sensitive, then this might reject values that were used before and previously considered valid.

VMEncryption/main/Common.py
vermashi: Ah, well I think that's just what we supply forward to Key Vault for wrapping/unwrapping when using a kek.
According to [this](https://docs.microsoft.com/en-us/rest/api/keyvault/wrapkey/wrapkey#jsonwebkeyencryptionalgorithm) that list seems up to date.
As to what's the origin of this list? I just found it in the single pass extension and copied it over, thanks for bringing this to my attention!

VMEncryption/main/check_util.py
vermashi: We have removed the strict url checks that would need to be updated for nclouds. We did this because it seemed like it would be difficult to maintain and host agent will be better at catching those errors anyways. So we should instead just work on a mechanism for relaying forward the host agent side errors

VMEncryption/main/Common.py
ejarvi: Yes I think we are good then, as we're using the most secure by default and matching the currently documented API for wrapKey.   https://docs.microsoft.com/en-us/rest/api/keyvault/wrapkey/wrapkey#jsonwebkeyencryptionalgorithm

VMEncryption/main/check_util.py
vermashi: All the id_splits were converted to lower case earlier to make the checks effectively case-insensitive

VMEncryption/main/check_util.py
ejarvi: is the expected string removed or updated?  just caught my eye since the "vault.azure.net" suffix is not expected in national clouds

VMEncryption/main/check_util.py
ejarvi: Yes, my only concern here was if the value was case sensitive and required an upper case, but the user provided it in all lower case, then this check would mark that as valid even though it wasn't.  I think we can rule that out though, as based on this list which includes casing requirements, it seems everything except for storage blob names seem to be case insensitive:  https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions

VMEncryption/main/check_util.py
vermashi: Taking a look

ejarvi: I don't feel this needs to hold up the rest of the PR, it might be something to note and then pick up in a future round of changes.

vermashi: I agree, if it looks like people are really prone to making a key vault id error we can look into this more

VMEncryption/main/check_util.py
ejarvi: Object identifiers are case-insensitive but the documentation doesn't specify that for algorithm identifiers.  Will check with KV team on this  https://docs.microsoft.com/en-us/azure/key-vault/about-keys-secrets-and-certificates#objects-identifiers-and-versioning

ejarvi: KV algorithm identifiers are case insensitive (upper and lower case allowed).  Check now supports this, so we are good.  https://github.com/Azure/azure-linux-extensions/pull/676/files#diff-925b65e5f426b09d40462e43eb3b4a82R169

VMEncryption/main/check_util.py
ejarvi: surplus string, now removed, resolving conversation

ejarvi:
LGTM

PR #703, opened 2018-12-13
Seperating log to console vs log to file for CSE

Diastro:
Seperating log to console vs log to file;
This is for the new requirement for which some extension output should not be logged to console (but only  to file).
This refactor also includes changes to CSE to log script output to file but not to console.

### DIFF
--- a/Common/WALinuxAgent-2.0.16/waagent
+++ b/Common/WALinuxAgent-2.0.16/waagent
@@ -2795,7 +2795,7 @@ class Logger(object):
         """
         return (counter < 10) or ((counter < 100) and ((counter % 10) == 0)) or ((counter % 100) == 0)
 
-    def LogToFile(self, message):
+    def WriteToFile(self, message):
         """
         Write 'message' to logfile.
         """
@@ -2815,7 +2815,7 @@ class Logger(object):
             except IOError as e:
                 pass
 
-    def LogToCon(self, message):
+    def WriteToConsole(self, message):
         """
         Write 'message' to /dev/console.
         This supports serial port logging if the /dev/console
@@ -2844,17 +2844,23 @@ class Logger(object):
         """
         self.LogWithPrefix("", message)
 
-    def LogWithPrefix(self, prefix, message):
+    def LogToConsole(self, message):
         """
-        Prefix each line of 'message' with current time+'prefix'.
+        Logs message to console by pre-pending each line of 'message' with current time.
         """
-        t = time.localtime()
-        t = "%04u/%02u/%02u %02u:%02u:%02u " % (t.tm_year, t.tm_mon, t.tm_mday, t.tm_hour, t.tm_min, t.tm_sec)
-        t += prefix
+        log_prefix = self._get_log_prefix("")
         for line in message.split('\n'):
-            line = t + line
-            self.LogToFile(line)
-            self.LogToCon(line)
+            line = log_prefix + line
+            self.WriteToConsole(line)
+
+    def LogToFile(self, message):
+        """
+        Logs message to file by pre-pending each line of 'message' with current time.
+        """
+        log_prefix = self._get_log_prefix("")
+        for line in message.split('\n'):
+            line = log_prefix + line
+            self.WriteToFile(line)
 
     def NoLog(self, message):
         """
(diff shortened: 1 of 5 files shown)


Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -2795,7 +2795,7 @@ class Logger(object):
         """
         return (counter < 10) or ((counter < 100) and ((counter % 10) == 0)) or ((counter % 100) == 0)
 
-    def LogToFile(self, message):
+    def WriteToFile(self, message):
```
boumenot: This file is shared by other extensions in this repo.  It isn't clear if these changes are self-contained or not.  Please double check that no other extensions are using these methods.

Utils/ScriptUtil.py
```diff
@@ -47,31 +47,39 @@ def run_command(hutil, args, cwd, operation, extension_short_name, version, exit
...
+            hutil.do_status_report(operation, 'transitioning', '0', msg_without_cmd_output)
             time.sleep(interval)
 
         exit_code = child.returncode
         if child.returncode and child.returncode != 0:
-            msg = LogUtil.get_formatted_log("Command returned an error.",
+            msg_with_cmd_output = LogUtil.get_formatted_log("Command returned an error.",
                                             LogUtil.tail(std_out_file), LogUtil.tail(err_out_file))
-            hutil.error(msg)
+            msg_without_cmd_output = "Command returned an error. Stdout/Stderr omitted from output."
```
vrdmr: You can add the exit code to the `msg_without_cmd_output ` variable. Better for debuggability.

CustomScript/customscript.py
```diff
@@ -93,7 +93,7 @@ def dummy_command(operation, status, msg):
 
 
 def parse_context(operation):
-    hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName, console_logger=waagent.LogToCon, file_logger=waagent.LogToFile)
+    hutil = Util.HandlerUtility(waagent.Log, waagent.Error, ExtensionShortName, console_logger=waagent.LogToConsole, file_logger=waagent.LogToFile)
     hutil.do_parse_context(operation)
```
vrdmr: LogToConsole != WriteToConsole here.

vrdmr: And same for LogToFile

Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -2795,7 +2795,7 @@ class Logger(object):
         """
```
vrdmr: do you need to change the WALinuxAgent-2.0.14 as well?

Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -2903,6 +2915,13 @@ class Logger(object):
...
         self.LogWithPrefix("ERROR:", message)
 
+    def _get_log_prefix(self, prefix):
+        """
+        Generates the log prefix with timestamp+'prefix'.
+        """
+        t = time.localtime()
```
pgombar: Using UTC time instead of local time ensures consistency of timestamps. I suggest using str(datetime.datetime.utcnow())

Diastro: I don't mind making the change but I didn't want to change existing logic (this is just refactoring the duplicated code into its own method)!

Utils/ScriptUtil.py
Diastro: Good point.

Common/WALinuxAgent-2.0.16/waagent
Diastro: I just double checked. The previous "LogToFile()" method was only used by CSE and waagent.py (at least these are the only references I found from the project).

CustomScript/customscript.py
Diastro: What do you mean? Sorry I don't get your comment...

Common/WALinuxAgent-2.0.16/waagent
vrdmr: I think keeping the local time is okay. As the logs would reside on the box, it is okay. Even WAAgent logs in local time on the box.

PR #705, opened 2018-12-18
Scrub protected settings from files and logs

pgombar:
For VMAccess, prevent protected settings from appearing in extension logs and config file.

### DIFF
--- a/Utils/HandlerUtil.py
+++ b/Utils/HandlerUtil.py
@@ -151,8 +151,8 @@ class HandlerUtility:
 
     @staticmethod
     def redact_protected_settings(content):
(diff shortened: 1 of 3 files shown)


Utils/HandlerUtil.py
```diff
@@ -148,15 +149,21 @@ def log(self, message):
     def error(self, message):
         self._error(self._get_log_prefix() + message)
 
+    @staticmethod
```
vrdmr: Looks good. Is there a way you can test this code in the unit-tests? (you can create a new test file if you don't want the baggage of earlier tests, which do not run).

vrdmr: And we can sync tomorrow how you can use DCR to the E2E scenario tests.

pgombar: Added unit test for static method.

PR #714, opened 2019-01-10
Include of Arch Linux compatibility for VMBackup Extension 

jespunya:
Based on the RedHat Patch the code had been modified to include Arch Linux compatibility in order to enable the Back Up extension.


soumyapattnaik:
LGTM

AkechiShiro:
It seems this was never merged, no idea why, this was completely abandoned.

PR #717, opened 2019-01-14
Setting IsAnySnapshotFailed to false when success

bpramod:
Setting IsAnySnapshotFailed to false when success


vityagi:
Looks fine.

PR #719, opened 2019-01-16
 Diagnostics: Add blobEndpoint for storage accounts; bug fix for National clouds.

canfikret:
Blob endpoint needs to be set for json sinks to work.


Diagnostic/lad_config_all.py
```diff
@@ -462,9 +464,9 @@ def generate_all_configs(self):
             return False, "Must specify storageAccountSasToken"
         if '?' == token[0]:
             token = token[1:]
-        endpoint = get_storage_endpoint_with_account(account,
+        endpoint = get_storage_endpoints_with_account(account,
```
canfikret: review this line

PR #728, opened 2019-01-25
Not to reset ChallengeResponseAuthentication for all the enable calls

vrdmr:
Moving the resetting of _ChallengeResponseAuthentication_ flag to happen only after a new sequence number is received by the VMAccess extension.


vrdmr:
@narrieta/@pgombar - The end-to-end tests were successful (DungeonCrawler.vameru/78).

PR #729, opened 2019-01-26
nit; Python3 not correctly called

vrdmr:
Missed a small bug.


vrdmr:
@D1v38om83r - Can you please start a new deployment of the agent? Thanks.

PR #731, opened 2019-01-28
ADE DP Jan QOS

ejarvi:
- add extension shim for early exit if python 2.7 is unavailable on target system 
- move restore operation to update since it should be done prior to install 
- improve restore performance by replacing slower shutil.copy2 with cp wildcard copy
- update various exit points with telemetry specific exit codes
- reduce enable failures by failing early on lvm os validation instead of warning


VMEncryption/extension_shim.sh
```diff
@@ -0,0 +1,114 @@
...
+    PYTHON_VER=`${PYTHON} --version 2>&1`
+    if [[ "$PYTHON_VER" =~ "Python 2.7" ]]; then
+        echo $PYTHON_VER
+    else
+        echo "Python 2.7 required, found $PYTHON_VER" >&2
+        exit 51 # Not Supported
+    fi
+fi
+
+${PYTHON} ${COMMAND} ${operation}
```
vermashi: Consider adding the 2>&1 to this line, if the guest agent stderr fix doesn't work for us.

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -437,23 +437,22 @@ def archive_old_configs(self):
...
-
-                if file == 'mrseq':
-                    src = os.path.join(root, file)
-                    dest = os.path.join(os.path.join(self._context._config_dir, '..'), file)
-
-                    self.log("Copying {0} to {1}".format(src, dest))
-
-                    shutil.copy2(src, dest)
+        # restores all archived settings files carried over from prior extension versions 
+        # uses cp --preserve=timestamps for faster performance than shutil.copy2()
```
vermashi: why did we add "--preserve=timestamps"?
Might be good to have that info here in the comments.

VMEncryption/main/handle.py
```diff
@@ -345,6 +344,7 @@ def update():
     hutil.do_parse_context('Update')
     logger.log("Installing pre-requisites")
     DistroPatcher.update_prereq()
+    hutil.restore_old_configs()
```
vermashi: why are we adding this line here?

VMEncryption/setup.py
```diff
@@ -83,11 +83,11 @@
   "name": CommonVariables.extension_name,
   "version": "1.0",
   "handlerManifest": {
-    "installCommand": main_entry + " -install",
-    "uninstallCommand": main_entry + " -uninstall",
-    "updateCommand": main_entry + " -update",
-    "enableCommand": main_entry + " -enable",
-    "disableCommand": main_entry + " -disable",
+    "installCommand": "extension_shim.sh -c " + main_entry + " --install",
```
vermashi: might need an escaped quote here. Like: "extension_shim.sh -c \"{0}\" --install".format(main_entry)

VMEncryption/setup.py
```diff
@@ -83,11 +83,11 @@
   "name": CommonVariables.extension_name,
   "version": "1.0",
   "handlerManifest": {
-    "installCommand": "extension_shim.sh -c " + main_entry + " --install",
-    "uninstallCommand": "extension_shim.sh -c " + main_entry + " --uninstall",
-    "updateCommand": "extension_shim.sh -c " + main_entry + " --update",
-    "enableCommand": "extension_shim.sh -c " + main_entry + " --enable",
-    "disableCommand": "extension_shim.sh -c " + main_entry + " --disable",
+    "installCommand": "extension_shim.sh -c {0} --install".format(main_entry),
```
vermashi: I meant, do we need quotes around the {0}

VMEncryption/main/check_util.py
```diff
@@ -168,10 +149,41 @@ def validate_volume_type(self, public_settings):
         if not volume_type.lower() in map(lambda x: x.lower(), supported_types) :
             raise Exception("Unknown Volume Type: {0}, has to be one of {1}".format(volume_type, supported_types))
 
+    def validate_lvm_os(self, public_settings):
+        volume_type = public_settings.get(CommonVariables.VolumeTypeKey).lower()
```
vermashi: The "get" from public settings may return null.
And we may want to skip the check for operation that's not an enable-type operation (not one of Enable, EnableEncryptFOrmat, EFA)

VMEncryption/main/check_util.py
```diff
@@ -168,10 +149,41 @@ def validate_volume_type(self, public_settings):
         if not volume_type.lower() in map(lambda x: x.lower(), supported_types) :
             raise Exception("Unknown Volume Type: {0}, has to be one of {1}".format(volume_type, supported_types))
 
+    def validate_lvm_os(self, public_settings):
```
vermashi: Please migrate the tests for the previous method into tests for this method (and maybe add a few cases)

VMEncryption/test/test_check_util.py
```diff
@@ -121,17 +121,46 @@ def test_mount_scheme(self):
...
+
+    def test_skip_lvm_os_check_if_data_only_efa(self):
+        # skip lvm detection if data only 
+        self.cutil.validate_lvm_os({Common.CommonVariables.VolumeTypeKey: "DATA", Common.CommonVariables.EncryptionEncryptionOperationKey: Common.CommonVariables.EnableEncryptionFormatAll})
+
+    def test_skip_lvm_os_check_if_data_only_disable(self):
+        # skip lvm detection if data only 
+        self.cutil.validate_lvm_os({Common.CommonVariables.VolumeTypeKey: "DATA", Common.CommonVariables.EncryptionEncryptionOperationKey: Common.CommonVariables.DisableEncryption})
+
+    def test_skip_lvm_os_check_if_query_only_data(self):
```
vermashi: QueryEncryptionStatus generally is not accompanied by any volume type.
Please add a test case for that scenario.

VMEncryption/main/check_util.py
ejarvi: done

VMEncryption/main/check_util.py
ejarvi: added logic and tests to handle this

PR #733, opened 2019-02-01
ADE SP Jan QOS

ejarvi:
- add extension shim for early exit if python 2.7 is unavailable on target system
- move archive to disable so future versions can restore in update
- improve restore performance by replacing slower shutil.copy2 with cp wildcard copy
- update various exit points with telemetry specific exit codes
- reduce enable failures by failing early on lvm os validation instead of warning

### DIFF
--- a/VMEncryption/extension_shim.sh
+++ b/VMEncryption/extension_shim.sh
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+
+# Keeping the default command
+COMMAND=""
+PYTHON=""
+
+USAGE="$(basename "$0") [-h] [-i|--install] [-u|--uninstall] [-d|--disable] [-e|--enable] [-p|--update] [-m|--daemon]
+
+Program to find the installed python on the box and invoke a Python extension script using Python 2.7.
+
+where:
+    -h|--help       show this help text
+    -i|--install    install the extension
+    -u|--uninstall  uninstall the extension
+    -d|--disable    disable the extension
+    -e|--enable     enable the extension
+    -p|--update     update the extension
+    -m|--daemon     invoke daemon option
+    -c|--command    command to run
+
+example:
+# Install usage
+$ bash extension_shim.sh -i
+python ./main/handle.py -install
+
+# Custom executable python file
+$ bash extension_shim.sh -c ""hello.py"" -i
+python hello.py -install
+
+# Custom executable python file with arguments
+$ bash extension_shim.sh -c ""hello.py --install""
+python hello.py --install
+"
+
+function find_python(){
+    local python_exec_command=$1
+
+    # Check if there is python defined.
+    if command -v python >/dev/null 2>&1 ; then
+        eval ${python_exec_command}="python"
+    fi
+}
+
+# Transform long options to short ones for getopts support (getopts doesn't support long args)
+for arg in "$@"; do
+  shift
+  case "$arg" in
+    "--help")       set -- "$@" "-h" ;;
+    "--install")    set -- "$@" "-i" ;;
+    "--update")     set -- "$@" "-p" ;;
+    "--enable")     set -- "$@" "-e" ;;
+    "--disable")    set -- "$@" "-d" ;;
+    "--uninstall")  set -- "$@" "-u" ;;
+    "--daemon")     set -- "$@" "-m" ;;
+    *)              set -- "$@" "$arg"
+  esac
+done
+
+if [ -z "$arg" ]
+then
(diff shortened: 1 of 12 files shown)


VMEncryption/main/handle.py
```diff
@@ -61,16 +61,23 @@
...
+    # In the future, moving the config archive/restore step to disable/update is preferred 
     hutil.restore_old_configs()
     hutil.do_exit(0, 'Install', CommonVariables.extension_success_status, str(CommonVariables.success), 'Install Succeeded')
 
 def disable():
     hutil.do_parse_context('Disable')
+    # archiving old configs during disable rather than uninstall will allow subsequent versions 
+    # to restore these configs in their update step rather than their install step once all 
+    # released versions of the extension are at this version or above 
+    hutil.archive_old_configs()
```
mayank88mahajan: Did we actually figure out why we even archive old configs?

VMEncryption/main/handle.py
```diff
@@ -1552,7 +1559,7 @@ def daemon_encrypt():
         except Exception as e:
             message = "Failed to encrypt data volumes with error: {0}, stack trace: {1}".format(e, traceback.format_exc())
             logger.log(msg=message, level=CommonVariables.ErrorLevel)
-            hutil.do_exit(exit_code=0,
+            hutil.do_exit(exit_code=CommonVariables.encryption_failed,
```
mayank88mahajan: What will changing the exit codes do here? Help in filtering or something else?

VMEncryption/main/handle.py
ejarvi: If the most recent configuration settings file does not include encryption operation and volume type information, the extension will walks backwards based on file timestamp through the archived settings file until it finds the setting file that contains this information.

VMEncryption/setup.py
```diff
@@ -81,11 +81,11 @@
   "name": CommonVariables.extension_name,
   "version": "1.0",
   "handlerManifest": {
-    "installCommand": main_entry + " -install",
-    "uninstallCommand": main_entry + " -uninstall",
-    "updateCommand": main_entry + " -update",
-    "enableCommand": main_entry + " -enable",
-    "disableCommand": main_entry + " -disable",
+    "installCommand": "extension_shim.sh -c {0} --install".format(main_entry),
```
mayank88mahajan: >extension_shim.sh [](start = 23, length = 17)

are there scenarios where shell script execution is not allowed on the system?

VMEncryption/main/handle.py
ejarvi: Yes, this helps to filter the telemetry for backend QOS reporting, as it allows categorizing failure types by exit code.

VMEncryption/setup.py
ejarvi: The problem being addressed here is that the extension did not gracefully with an "unsupported" error message and the proper exit code for telemetry when installed on a custom image without the necessary python 2.7 support present.   Even on a system that cannot execute python scripts, it can still usually support a shell script.  By adding this shim, we can fail gracefully from within the shim on systems that do not support python.

VMEncryption/extension_shim.sh
```diff
@@ -0,0 +1,119 @@
...
+# Custom executable python file
+$ bash extension_shim.sh -c ""hello.py"" -i
+python hello.py -install
+
+# Custom executable python file with arguments
+$ bash extension_shim.sh -c ""hello.py --install""
+python hello.py --install
+"
+
+function find_python(){
```
vrdmr: You aren't planning to add Python3 check explicitely?

```
function find_python(){
    local python_exec_command=$1

    # Check if there is python defined.
    if command -v python >/dev/null 2>&1 ; then
        eval ${python_exec_command}="python"
    else
        # Python was not found. Searching for Python3 now.
        if command -v python3 >/dev/null 2>&1 ; then
            eval ${python_exec_command}="python3"
        fi
    fi
}
```

ejarvi: The extension code that this shim calls is not yet compatible with Python3.  We are hoping to make extension code Python3 compatible in the future.  Once it is, then we'd add it to the shim too.

vrdmr: You can use [2to3](https://docs.python.org/2/library/2to3.html) to do most of your work (transform from 2 to 3) and test it with a test version of the extension (Copy on the box and run in ad-hoc way). I used that for VMAccess for Linux, and HandlerUtils have been changed for the same as well - and then had to do very minor changes. Something for you to consider.

PR #737, opened 2019-02-12
Fail early when dependencies are missing in RHEL 7

vermashi:


VMEncryption/main/patch/redhatPatching.py
```diff
@@ -109,7 +101,16 @@ def install_extras(self):
...
+            ret_val = self.command_executor.Execute("yum install -y " + " ".join(packages))
+            if ret_val != CommonVariables.process_success and raise_on_failure:
+                msg = """
+                Failed to install all required dependencies.
+                Please check the internet connection, firewall settings
+                and package manager setup on the VM and try again.
+                Alternatively, install the following dependencies manually and then try again:
+                {}"""
+                msg = msg.format(str(packages))
+                raise Exception(msg)
```
ejarvi: If we intend to fail early here, it may be desirable to simply log the failure here and exit with the CommonVariables.missing_dependency exit code so QOS telemetry can pick it up and avoid the risk of upstream caller of this function not catching or reporting the exception properly as a failure.

VMEncryption/main/patch/redhatPatching.py
```diff
@@ -132,7 +133,7 @@ def update_prereq(self):
                 dracut_repack_needed = True
 
             if os.path.exists("/dev/mapper/osencrypt"):
-                #TODO: only do this if needed (if code and existing module are different)
+                # TODO: only do this if needed (if code and existing module are different)
```
mayank88mahajan: What is this TODO? can we remove it?

VMEncryption/main/patch/redhatPatching.py
```diff
@@ -182,23 +183,22 @@ def add_91_ade_dracut_module(command_executor):
...
 
-
     @staticmethod
     def remove_old_patching_system(logger, command_executor):
         logger.log("Removing patches and recreating initrd image")
 
         command_executor.Execute('mv /lib/dracut/modules.d/90crypt/cryptroot-ask.sh.orig /lib/dracut/modules.d/90crypt/cryptroot-ask.sh', False)
         command_executor.Execute('mv /lib/dracut/modules.d/90crypt/module-setup.sh.orig /lib/dracut/modules.d/90crypt/module-setup.sh', False)
         command_executor.Execute('mv /lib/dracut/modules.d/90crypt/parse-crypt.sh.orig /lib/dracut/modules.d/90crypt/parse-crypt.sh', False)
-
```
mayank88mahajan: nit: remove

VMEncryption/main/patch/redhatPatching.py
```diff
@@ -109,7 +101,16 @@ def install_extras(self):
             packages.remove('util-linux')
 
         if self.command_executor.Execute("rpm -q " + " ".join(packages)):
-            self.command_executor.Execute("yum install -y " + " ".join(packages))
+            ret_val = self.command_executor.Execute("yum install -y " + " ".join(packages))
```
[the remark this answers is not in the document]
vermashi: but the rpm -q is there for that no?

VMEncryption/main/patch/redhatPatching.py
vermashi: good point, I'll make it so that the exit code gets sent out correctly.

VMEncryption/main/patch/redhatPatching.py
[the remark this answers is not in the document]
ejarvi: Yes but the rpm -q call is checking against an entire list of packages.  Some of the packages in that list might already be on the system, and some might not.  If it finds even one missing package, it will drop into yum install -y which will then try to install the entire list of packages.  If one of those packages was already installed yum install -y will return a failure code even after installing the missing packages because a subset of the packages it was asked to install were already there.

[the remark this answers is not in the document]
vermashi: Interesting.
Let me add a second rpm -q then to handle that scenario.

ejarvi:
@vermashi please update this PR with current status, since it was not merged prior to 3/1 release

PR #751, opened 2019-02-26
Ade vmss volume type check fix

vermashi:


VMEncryption/main/handle.py
```diff
@@ -721,6 +699,7 @@ def enable_encryption():
                 encryption_marker = mark_encryption(command=extension_parameter.command,
                                                     volume_type=extension_parameter.VolumeType,
                                                     disk_format_query=extension_parameter.DiskFormatQuery)
+                start_daemon('EnableEncryption')
```
ejarvi: To make sure I am undrestanding this change correctly, addition of this line will fix the issue of encryption settings not being set on a subsequent call, correct?

vermashi: This particular is a little bit complicated to explain, but here it goes:
This code fragment is a leftover from the 2-pass version:
   At this point, the status would be populated with the secret url, which the client would add to the VM model, triggering a reboot. Then the extension would start again, see the "encryption marker" made in the previous line and start encryption.
In single pass we don't need this and ending this branch without this line leads to an unnecessary delay in encryption, as we have to wait for the loop in handle.py:525 to come back here and take the other branch on line 650.

vermashi:
Performing manual testing...

vermashi:
Test successful. Incrementing version number and merging.

PR #752, opened 2019-02-27
Fixing the retry counter for WireProtocol messages

vermashi:


VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -234,8 +243,8 @@ def post_to_wireserver(self, data):
                 else:
                     raise Exception("no response from encryption settings update request")
             except Exception as e:
+                retry_count += 1
                 if retry_count < retry_count_max:
-                    retry_count += 1
                     time.sleep(5) # sleep for 5 seconds before retrying.
```
ejarvi: It would be helpful for log analysis to also log the retry count and the exception message that is triggering the retry within this block.  Otherwise I believe the first 2 exception messages will get swallowed.

VMEncryption/test/test_encryption_settings_util.py
```diff
@@ -0,0 +1,27 @@
...
+    def setUp(self):
+        self.logger = console_logger.ConsoleLogger()
+        self.es_util = EncryptionSettingsUtil.EncryptionSettingsUtil(self.logger)
+
+    @mock.patch('time.sleep') # To speed up this test.
+    @mock.patch('main.EncryptionSettingsUtil.EncryptionSettingsUtil.write_settings_file')
+    @mock.patch('main.EncryptionSettingsUtil.EncryptionSettingsUtil.get_index')
+    @mock.patch('os.path.isfile', return_value=True)
+    @mock.patch('main.EncryptionSettingsUtil.EncryptionSettingsUtil.get_http_util')
+    def test_post_to_wire_server(self, get_http_util, os_path_isfile, get_index, write_settings_file, time_sleep):
```
ejarvi: Tests of an HTTP 200, HTTP 400, and HTTP 500 response from wire server (in addition to no response) would also be helpful.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -234,8 +243,8 @@ def post_to_wireserver(self, data):
...
                 else:
                     raise e
```
ejarvi: To avoid no boot scenarios later, the encryption operation should fail at this point (if after multiple retries the encryption settings could not be uploaded to the host).  Rather than re-raising the exception and hoping the caller does this, maybe it would be better to simply exit with failure right at this point ensuring the properly telemetry and status is being reflected.

vermashi: post_to_wireserver is a method used at multiple places and it would be hard to error out here with the right error codes. So, I want to keep the onus on the caller to exit with the right code and operation string.

ejarvi: I haven't had a chance to review every caller yet, but that may be necessary since it is unsafe to allow encryption to proceed if encryption settings haven't been stamped to disk.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -244,6 +244,7 @@ def post_to_wireserver(self, data):
                     raise Exception("no response from encryption settings update request")
             except Exception as e:
                 retry_count += 1
+                self.logger.log("Encountered exception while postiong encyrption settings to Wire Server (attempt #{0}):\n{1}".format(str(retry_count), str(e)))
```
ejarvi: typo: "post**ion**g en**cyr**ption" -> "posting encryption"

VMEncryption/main/DiskUtil.py
```diff
@@ -635,6 +635,10 @@ def get_mount_items(self):
 
         return items
 
+    def is_in_memfs_root(self):
+        mounts = file('/proc/mounts', 'r').read()
+        return bool(re.search(r'/\s+tmpfs', mounts))
```
ejarvi: corner case - if any standard mount point has the string "tmpfs" in the name this will erroneously return true.  If you agree, it may be worth adding this as a todo comment as there may be a better approach in the future.

vermashi: I think '\s+' in regex-speak means "atleast one space-type character".
and because it is rare for mount points to have a bunch of spaces followed by "tmpfs" I feel its okay to leave it here for now.
It should definitely be made more robust later though.
Will add the TODO.

VMEncryption/main/EncryptionSettingsUtil.py
vermashi: I looked at a the callers on this method.
And this does always halt the encryption process.
The error codes are "encryption_failed" or "unknown_error". Which while unfortunate.. is kind of true. When the "post_to_wireserver" fails, at this time, we actually do not know why is failed.

VMEncryption/main/EncryptionSettingsUtil.py
vermashi: thanks, fixed

PR #754, opened 2019-02-27
ADE DP Feb QOS - fix syntax error in ExtensionParameter.py

ejarvi:
- the syntax error in this code path has caused enable failures observed in QOS telemetry


vermashi:
LGTM.

PR #756, opened 2019-02-28
Automatically restart VM after dual pass encryption completes.

vimish:
Some distros like RHEL & CentOS are not able send VMRestartPending
status to the portal. This seems to occur due to agent not working
properly in pivot root env. As a result, the extension is not
completing as customer don't know to restart the VM due to lack
of notification.
In this fix a rebbot is added at the end of encryption. This
will remove the need of manual restart by the customer.


vimish:
@ejarvi @mayank88mahajan @vermashi  Please review the PR.

PR #763, opened 2019-03-12
Stop using by-id paths for devices getting encrypted

vermashi:


VMEncryption/main/handle.py
```diff
@@ -138,7 +138,8 @@ def disable_encryption():
             if crypt_item.dev_path.startswith("/dev/sd"):
                 logger.log('Updating crypt item entry to use mapper name')
                 logger.log('Device name before update: {0}'.format(crypt_item.dev_path))
-                crypt_item.dev_path = disk_util.query_dev_id_path_by_sdx_path(crypt_item.dev_path)
+                # crypt_item.dev_path = disk_util.query_dev_id_path_by_sdx_path(crypt_item.dev_path)
```
ejarvi: probably no need to comment this, it's in version control

VMEncryption/main/DiskUtil.py
```diff
@@ -735,6 +735,8 @@ def query_dev_sdx_path_by_uuid(self, uuid):
     def query_dev_id_path_by_sdx_path(self, sdx_path):
         """
         return /dev/disk/by-id that maps to the sdx_path, otherwise return the original path
+        Update: now we have realised that by-id is not a good way to refer to devices (they can change on reallocations or resizes).
+        Try not to use this- use get_stable_path_from_sdx instead
```
ejarvi: are there any other places in the code calling this still?

VMEncryption/main/DiskUtil.py
```diff
@@ -743,6 +745,27 @@ def query_dev_id_path_by_sdx_path(self, sdx_path):
...
+            return azure_name_table[sdx_realpath]
+
+        # Then try matching a uuid symlink. Those are probably the best
+        for disk_by_uuid in os.listdir(CommonVariables.disk_by_uuid_root):
+            disk_by_uuid_path = os.path.join(CommonVariables.disk_by_uuid_root, disk_by_uuid)
+
+            if os.path.realpath(disk_by_uuid_path) == sdx_realpath:
+                return disk_by_uuid_path
+
+        # Found nothing very persistent. Just return the original sdx path.
```
ejarvi: consider erroring or logging a warning if get_persistent_path failed to get a persistent path

VMEncryption/main/handle.py
vermashi: Good point, I will go ahead and remove it.

VMEncryption/main/DiskUtil.py
vermashi: There is this one other place in OS encryption and it doesn't affect the azure_crypt_mount file.
So, I decided to not touch it (yet).

VMEncryption/main/DiskUtil.py
vermashi: Good Idea, will do.

VMEncryption/main/DiskUtil.py
ejarvi: OK

mayank88mahajan: Add a warning log statement. If this gets used, we would know.

VMEncryption/main/handle.py
```diff
@@ -138,7 +138,7 @@ def disable_encryption():
             if crypt_item.dev_path.startswith("/dev/sd"):
                 logger.log('Updating crypt item entry to use mapper name')
                 logger.log('Device name before update: {0}'.format(crypt_item.dev_path))
-                crypt_item.dev_path = disk_util.query_dev_id_path_by_sdx_path(crypt_item.dev_path)
+                crypt_item.dev_path = disk_util.get_persistent_path_by_sdx_path(crypt_item.dev_path)
```
mayank88mahajan: Fallback to dev by-id path incase persistent paths are missing?

VMEncryption/main/DiskUtil.py
```diff
@@ -743,6 +745,30 @@ def query_dev_id_path_by_sdx_path(self, sdx_path):
...
+        # And Log it.
+        self.logger.log(msg="Failed to find a persistent path for [{0}].".format(sdx_path), level=CommonVariables.WarningLevel)
+
```
mayank88mahajan: Do we expect this condition to occur? I would recommend to fail here rather than silently proceed with the dev path (/dev/sdd1 or something). We will end up failing to mount the data disks later on.

vermashi: This can occur when they are encrypting over RAID or LVM. But that is fine because mapper names arising from those systems are persistent.
That is why I don't want to just error out there.

VMEncryption/main/handle.py
vermashi: I don't know enough about by-id to add that here just yet.
Perhaps we can do it later after a little more research.

VMEncryption/main/DiskUtil.py
vermashi: So far, there are no references to this method outside of an OS Encryption code path.
And the by-id path returned there is not added to azure_crypt_mount and doesn't contribute to the problem.

vermashi:
@mayank88mahajan @ejarvi  Thanks you all for the review. Will Merge soon. :-)

PR #767, opened 2019-03-14
ADE SP March QOS

ejarvi:
- add vfat module check to precheck

### DIFF
--- a/VMEncryption/main/check_util.py
+++ b/VMEncryption/main/check_util.py
@@ -24,6 +24,7 @@ import urlparse
 import re
 from Common import CommonVariables
 from MetadataUtil import MetadataUtil
+from CommandExecutor import CommandExecutor
 
 class CheckUtil(object):
     """Checks compatibility for disk encryption"""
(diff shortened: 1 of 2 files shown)


VMEncryption/main/check_util.py
```diff
@@ -223,11 +224,20 @@ def validate_lvm_os(self, public_settings):
...
+            executor.Execute("modprobe vfat", True)
+        except:
+            raise RuntimeError('Incompatible system, prerequisite vfat module was not found.')
+
     def precheck_for_fatal_failures(self, public_settings):
         """ run all fatal prechecks, they should throw an exception if anything is wrong """
         self.validate_key_vault_params(public_settings)
         self.validate_volume_type(public_settings)
         self.validate_lvm_os(public_settings)
+        self.validate_vfat()
```
vermashi: Is absence of vfat a fatal failure even for Data disk encryption?

VMEncryption/main/check_util.py
```diff
@@ -223,11 +224,20 @@ def validate_lvm_os(self, public_settings):
         if detected:
             raise Exception("LVM OS disk layout does not satisfy prerequisites ( see https://aka.ms/adelvm )")
 
+    def validate_vfat(self):
+        """ Check for vfat module using modprobe and raise exception if not found """
+        try:
+            executor = CommandExecutor(self.logger)
+            executor.Execute("modprobe vfat", True)
```
vermashi: Does this command remain the same for all distros?
1) Do Ubuntu and RedHat both have the command "modprobe"?
2) Is the module called "vfat" in both distros?

ejarvi: Yes, good question.  Testing showed that the command is available across all currently supported image types (RedHat,CentOS,Ubuntu,SUSE) and the vfat module was found in each of those as well. We are using modprobe elsewhere in the extension.

VMEncryption/main/check_util.py
ejarvi: I could be missing something here, but in my opinion yes, we want it to terminate early even in the data disk case.  Without vfat module support the BEK volume cannot be mounted, the key cannot be read (or written to in single pass) and data disks cannot subsequently be unlocked.   All of those downstream issues are more expensive to recover from than if the scenario fails fast upstream and provides the user with a clear and actionable message that the system does not satisfy the vfat prerequisite.

vermashi:
LGTM. Merge if testing has been successful.

PR #774, opened 2019-03-22
ADE SP March QoS

vimish:
1. Fail OS encryption if memory less than 7 GB
2. Do apt-get update only when install fails

### DIFF
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -528,10 +528,25 @@ def enable():
         public_settings = get_public_settings()
         logger.log('Public settings:\n{0}'.format(json.dumps(public_settings, sort_keys=True, indent=4)))
         cutil = CheckUtil(logger)
+        # Mount already encrypted disks before running fatal prechecks
+        disk_util = DiskUtil(hutil=hutil, patching=DistroPatcher, logger=logger, encryption_environment=encryption_environment)
+        bek_util = BekUtil(disk_util, logger)
+        existing_passphrase_file = None
+        encryption_config = EncryptionConfig(encryption_environment=encryption_environment, logger=logger)
+
+        existing_passphrase_file = bek_util.get_bek_passphrase_file(encryption_config)
+        if existing_passphrase_file is not None:
+            disk_util.consolidate_azure_crypt_mount(existing_passphrase_file)
+            mount_encrypted_disks(disk_util=disk_util,
+                                  bek_util=bek_util,
+                                  encryption_config=encryption_config,
+                                  passphrase_file=existing_passphrase_file)
+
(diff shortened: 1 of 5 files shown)


VMEncryption/main/handle.py
```diff
@@ -628,6 +628,16 @@ def enable_encryption():
                        level=CommonVariables.WarningLevel)
             hutil.redo_last_status()
             exit_without_status_report()
+    
+    cutil = CheckUtil(logger)
```
vermashi: Please add this check to the fatal checks method in CheckUtil instead.

VMEncryption/main/patch/UbuntuPatching.py
```diff
@@ -58,12 +58,17 @@ def __init__(self, logger, distro_info):
...
-        self.command_executor.Execute(cmd)
-        
         packages = ['cryptsetup-bin']
         cmd = " ".join(['apt-get', 'install', '-y'] + packages)
-        self.command_executor.Execute(cmd)
+        return_code = self.command_executor.Execute(cmd)
+        
+        # If install fails, try running apt-get update and then try install again
+        if return_code != 0:
+            cmd = " ".join(['apt-get', '-o Acquire::ForceIPv4=true', '-y', 'update'])
```
vermashi: Why are you writing out the command like this?
Just do it normally, its more readable: "apt-get -o Acquire::ForceIPv4-True -y update"

VMEncryption/main/handle.py
vermashi: I'm talking about the method already called at line 534 of this file.

vermashi: if (volume_type is OS/ALL and OSdiskStatus is NOtEncrypted (ref: line 117) and operation is EnableType (ref: line 141) )
    do the memory check ()

VMEncryption/main/handle.py
```diff
@@ -528,10 +528,25 @@ def enable():
...
+        disk_util = DiskUtil(hutil=hutil, patching=DistroPatcher, logger=logger, encryption_environment=encryption_environment)
+        bek_util = BekUtil(disk_util, logger)
+        existing_passphrase_file = None
+        encryption_config = EncryptionConfig(encryption_environment=encryption_environment, logger=logger)
+
+        existing_passphrase_file = bek_util.get_bek_passphrase_file(encryption_config)
+        if existing_passphrase_file is not None:
+            disk_util.consolidate_azure_crypt_mount(existing_passphrase_file)
+            mount_encrypted_disks(disk_util=disk_util,
+                                bek_util=bek_util,
```
vermashi: Fix indentation here

vermashi:
Looks good but look at comments

vimish:
Addressed review comments.

PR #775, opened 2019-03-22
ADE DP March QoS

vimish:
1. Fail OS encryption if memory less than 7 GB
2. Do apt-get update only when install fails

### DIFF
--- a/VMEncryption/main/check_util.py
+++ b/VMEncryption/main/check_util.py
@@ -211,13 +211,24 @@ class CheckUtil(object):
         else: 
             raise Exception(CommonVariables.AADClientIDKey + ' property was not found in settings')
             
-    def precheck_for_fatal_failures(self, public_settings):
+    def validate_memory_os_encryption(self, public_settings, encryption_status):
+        is_enable_operation = False
+        encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
+        if encryption_operation in [CommonVariables.EnableEncryption, CommonVariables.EnableEncryptionFormat, CommonVariables.EnableEncryptionFormatAll]:
+            is_enable_operation = True
+        volume_type = public_settings.get(CommonVariables.VolumeTypeKey)
+        if is_enable_operation and not volume_type.lower() == CommonVariables.VolumeTypeData.lower() and encryption_status["os"] == "NotEncrypted":
+            if self.is_insufficient_memory():
+                raise Exception("Not enough memory for enabling encryption on OS volume. 8 GB memory is recommended.")
+
+    def precheck_for_fatal_failures(self, public_settings, encryption_status):
         """ run all fatal prechecks, they should throw an exception if anything is wrong """
         self.validate_key_vault_params(public_settings)
         self.validate_volume_type(public_settings)
         self.validate_lvm_os(public_settings)
         self.validate_vfat()
         self.validate_aad(public_settings)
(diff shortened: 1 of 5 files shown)


vimish:
Addressed review comments.

VMEncryption/main/check_util.py
```diff
@@ -185,11 +185,22 @@ def validate_lvm_os(self, public_settings):
...
             raise Exception("LVM OS disk layout does not satisfy prerequisites ( see https://aka.ms/adelvm )")
 
-    def precheck_for_fatal_failures(self, public_settings):
+    def validate_memory_os_encryption(self, public_settings, encryption_status):
+        is_enable_operation = False
+        encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
+        if encryption_operation in [CommonVariables.EnableEncryption, CommonVariables.EnableEncryptionFormat, CommonVariables.EnableEncryptionFormatAll]:
+            is_enable_operation = True
+        volume_type = public_settings.get(CommonVariables.VolumeTypeKey)
+        if is_enable_operation and not volume_type.lower() == CommonVariables.VolumeTypeData.lower() and encryption_status["os"] == "NotEncrypted":
```
ejarvi: on first encrypt, is encryption_status always non-null and populated with "NotEncrypted" string?  May be good to replace this with a constant to keep in synch with other locations in code that might set it, rather than hardcoded string "NotEncrypted"

VMEncryption/main/check_util.py
```diff
@@ -185,11 +185,22 @@ def validate_lvm_os(self, public_settings):
...
+            if self.is_insufficient_memory():
+                raise Exception("Not enough memory for enabling encryption on OS volume. At least 7 GB memory is required.")
```
ejarvi: 7GB is a bare minimum requirement, 8GB is recommended.  Also the available free memory is important, a system may have total RAM but due to other processes not have sufficient free space.   Maybe change "At least 7GB memory is required" to "8 GB memory is recommended."

VMEncryption/main/patch/UbuntuPatching.py
```diff
@@ -58,8 +58,10 @@ def __init__(self, logger, distro_info):
         self.touch_path = '/usr/bin/touch'
 
     def install_adal(self):
-        self.command_executor.Execute('apt-get update')
-        self.command_executor.Execute('apt-get install -y python-pip')
+        return_code = self.command_executor.Execute('apt-get install -y python-pip')
+        if return_code != 0:
+            self.command_executor.Execute('apt-get -o Acquire::ForceIPv4=true -y update')
```
ejarvi: add a comment explaining what is being done here, and a log message indicating that this path was taken

VMEncryption/test/test_check_util.py
```diff
@@ -69,41 +69,41 @@ def test_validate_volume_type(self):
     def test_fatal_checks(self):
```
ejarvi: Unit tests seem to be missing the case where "OS" was already encrypted, or cases with with less than, equal to, or greater than the required amount of RAM.

VMEncryption/main/check_util.py
vimish: By default the encryption status is set to NotEncrypted.
Ref: https://github.com/Azure/azure-linux-extensions/blob/master/VMEncryption/main/DiskUtil.py#L546

PR #780, opened 2019-04-08
Adding support for the new Wire Protocol [BekVHDFix]

vermashi:


VMEncryption/main/BekUtil.py
```diff
@@ -44,71 +44,50 @@ def generate_passphrase(self, algorithm):
...
+                    return os.path.join(self.bek_filesystem_mount_point, file)
+
+        except Exception as e:
+            message = "Failed to get BEK from BEK VOLUME with error: {0}".format(str(e))
+            self.logger.log(message)
 
         return None
 
+    def mount_bek_volume(self):
+        self.disk_util.mount_by_label("BEK VOLUME", self.bek_filesystem_mount_point, "fmask=077")
```
ejarvi: Would be nice to define "BEK VOLUME" as a constant and use that here rather than hardcoded string

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -31,7 +33,8 @@ class EncryptionSettingsUtil(object):
 
     def __init__(self, logger):
         self.logger = logger
-        self._CURRENT_DISK_ENCRYPTION_DATA_VERSION = "3.0"
+        self._CURRENT_DISK_ENCRYPTION_DATA_VERSION = "4.0"
```
ejarvi: The new section added around 52 in Common.Py references 3.0, not 4.0.  Is that correct?

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -238,18 +245,44 @@ def post_to_wireserver(self, data):
...
+    def post_to_wireserver(self, data):
+        """ Request EnableEncryption operation on settings file via wire server """
+        self.write_settings_file(data)
+        if not os.path.isfile(self.get_settings_file_path()):
+            raise Exception(
+                'Disk encryption settings file not found: ' + self.get_settings_file_path())
+
+        http_util = self.get_http_util()
+
+        # V3 message content
```
ejarvi: New version of EncryptionSettingsUtil.py in line 36-37 references 4.0 and 3.0, this post code references a wire server version 3 and 2, is there an off-by-one somewhere?

VMEncryption/main/EncryptionSettingsUtil.py
mayank88mahajan: put an example here to clarify which version is it talking about. This one is within the JSON content itself. 

---
In reply to: [273141600](https://github.com/Azure/azure-linux-extensions/pull/780#discussion_r273141600) [](ancestors = 273141600)

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -188,6 +197,12 @@ def controller_id_and_lun_to_settings_data(scsi_controller, lun_number):
             "KekAlgorithm": kek_algorithm,
             "Disks": data_disks_settings_data
         }
+
+        self.logger.log("Settings without the protectors array: " + json.dumps(data, sort_keys=True, indent=4))
```
mayank88mahajan: >without the protectors array [](start = 34, length = 28)

We should dump the protectors Name without the key in the JSON. This will let us know if the protectors section was added with the correct protector name or not.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -238,18 +245,44 @@ def post_to_wireserver(self, data):
 
                     http_util.connection.close()
                     if result.status != httplib.OK and result.status != httplib.ACCEPTED:
-                        raise Exception("encryption settings update request was not accepted")
+                        raise Exception("Encryption settings update request was not accepted")
```
mayank88mahajan: >update request [](start = 61, length = 14)

maybe change this exception message to not confuse with update_encryption_settings workflow itself. "Encryption settings could not be persisted". Or similar to Windows extension " Failed to send disk encryption data. Check KeyVault inputs, etc, "

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -31,7 +33,8 @@ class EncryptionSettingsUtil(object):
 
     def __init__(self, logger):
         self.logger = logger
-        self._CURRENT_DISK_ENCRYPTION_DATA_VERSION = "3.0"
+        self._CURRENT_DISK_ENCRYPTION_DATA_VERSION = "4.0"
+        self._PREVIOUS_DISK_ENCRYPTION_DATA_VERSION = "3.0"
```
mayank88mahajan: how about explicitly naming the variable to correspond to the version "_DISK_ENCRYPTION_DATA_VERSION_V3" and "_DISK_ENCRYPTION_DATA_VERSION_V4" ?

vermashi: Good idea- will do

VMEncryption/main/EncryptionSettingsUtil.py
vermashi: Yes, these versions are supposed to be different. The one there is for the wire protocol format itself.
This is for the JSON blob's data version

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -178,16 +181,31 @@ def controller_id_and_lun_to_settings_data(scsi_controller, lun_number):
...
             "KeyVaultResourceId": kv_id,
             "KekUrl": kek_url,
             "KekVaultResourceId": kek_kv_id,
             "KekAlgorithm": kek_algorithm,
+            "Protectors": protectors_name_only,
             "Disks": data_disks_settings_data
         }
+
+        self.logger.log("Settings without the protectors array: " + json.dumps(data, sort_keys=True, indent=4))
+        self.logger.log("Full Settings JSON might be found later in the BEK VOLUME")
```
mayank88mahajan: We need to remove dumping the settings.json file with the encryption passphrase key to bek volume. Passphrase in the settings.json file may go unoticed and also gives a wrong impression that it does not contain any secrets while we are dumping the key in there

vermashi: fixed

PR #787, opened 2019-04-22
Fix waagent import error on CoreOS

vimish:
waagent binary is in different location on CoreOs. This is causing import error and making extension fail before it can report that CoreOs is supported. Accounting for waagent location on CoreOs so that extension can execute as expected.

### DIFF
--- a/VMEncryption/main/Utils/WAAgentUtil.py
+++ b/VMEncryption/main/Utils/WAAgentUtil.py
@@ -29,11 +29,12 @@ def searchWAAgent():
     agentPath = '/usr/sbin/waagent'
     if os.path.isfile(agentPath):
         return agentPath
-    user_paths = os.environ['PYTHONPATH'].split(os.pathsep)
-    for user_path in user_paths:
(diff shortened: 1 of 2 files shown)


VMEncryption/main/Utils/WAAgentUtil.py
```diff
@@ -27,6 +27,9 @@
 #
 def searchWAAgent():
     agentPath = '/usr/sbin/waagent'
+    if os.path.isfile(agentPath):
+        return agentPath
+    agentPath = '/usr/share/oem/bin/waagent'
```
ejarvi: If an other unsupported operating system or images are introduced that have yet another agent path outside of these two, it would be good if this logic explicitly failed with telemetry exit code indicating an unsupported environment after attempting all known agent paths.

vimish:
Changed the logic to check for waagent binary in extension_shim.sh

ejarvi:
@vimish same as the DP PR, please ensure that positive and negative testcases on both supported and unsupported OS types fail as expected with this change prior to merge

PR #788, opened 2019-04-22
Fix waagent import error on CoreOS[DualPass]

vimish:
waagent binary is in different location on CoreOs. This is causing import error and making extension fail before it can report that CoreOs is supported. Accounting for waagent location on CoreOs so that extension can execute as expected.

### DIFF
--- a/VMEncryption/extension_shim.sh
+++ b/VMEncryption/extension_shim.sh
@@ -115,5 +115,12 @@ else
     fi
 fi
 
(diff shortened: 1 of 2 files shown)


vimish:
Changed the logic to check for waagent binary in extension_shim.sh

vermashi:
LGTM. Test on CoreOS: if it passes, increment the version number and merge.

vimish:
Abandoned. Closing the PR.

PR #792, opened 2019-04-26
assigning right error oode when partial snapshot failure

bpramod:


VMBackup/main/freezesnapshotter.py
```diff
@@ -159,8 +159,8 @@ def updateErrorCode(self, blob_snapshot_info_array, all_failed):
                         error_msg = error_msg + ExtensionErrorCodeHelper.ExtensionErrorCodeHelper.StatusCodeStringBuilder(self.extensionErrorCode)
                         self.logger.log(error_msg, True, 'Error')
                         break
-                    elif blob_snapshot_info.isSuccessful == False and not all_failed:
```
vityagi: this elif should still be kept

bpramod: Yes, i think i should still keep that. will do that change.

vityagi:
Looks fine.

vityagi:
Looks good.

PR #793, opened 2019-04-29
Timeout apt-get update if it is taking more time than expected

vimish:
Timeout and fail if apt-get update does not complete in 30 seconds.
Please provide feedback on 30 sec timeout for apt-get update.

### DIFF
--- a/VMEncryption/main/CommandExecutor.py
+++ b/VMEncryption/main/CommandExecutor.py
@@ -25,6 +25,7 @@ import shlex
 import sys
 
 from subprocess import *
+from threading import Timer
 
 class ProcessCommunicator(object):
     def __init__(self):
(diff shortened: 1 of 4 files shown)


VMEncryption/main/CommandExecutor.py
```diff
@@ -52,8 +55,19 @@ def Execute(self, command_to_execute, raise_exception_on_failure=False, communic
                     self.logger.log("Process creation failed: " + str(e))
                 return -1
 
-        stdout, stderr = proc.communicate(input=input)
-        return_code = proc.returncode
+        def timeout_process():
```
ejarvi: add unit tests for this function (same as DP PR)

ejarvi: add unit tests for timely and timeout cases  (same as DP PR)

vimish:
@vermashi @ejarvi 
Added unit test

PR #794, opened 2019-04-29
Timeout apt-get update if it is taking more time than expected[DualPass]

vimish:
Timeout and fail if apt-get update does not complete in 30 seconds.
Please provide feedback on 30 sec timeout for apt-get update.

### DIFF
--- a/VMEncryption/main/CommandExecutor.py
+++ b/VMEncryption/main/CommandExecutor.py
@@ -25,6 +25,7 @@ import shlex
 import sys
 
 from subprocess import *
+from threading import Timer
 
 class ProcessCommunicator(object):
     def __init__(self):
@@ -36,7 +37,7 @@ class CommandExecutor(object):
     def __init__(self, logger):
         self.logger = logger
 
-    def Execute(self, command_to_execute, raise_exception_on_failure=False, communicator=None, input=None, suppress_logging=False):
+    def Execute(self, command_to_execute, raise_exception_on_failure=False, communicator=None, input=None, suppress_logging=False, timeout=0):
         if type(command_to_execute) == unicode:
             command_to_execute = command_to_execute.encode('ascii', 'ignore')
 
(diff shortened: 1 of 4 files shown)


VMEncryption/main/CommandExecutor.py
```diff
@@ -55,8 +58,19 @@ def Execute(self, command_to_execute, raise_exception_on_failure=False, communic
                     self.logger.log("Process creation failed: " + str(e))
                 return -1
 
-        stdout, stderr = proc.communicate(input=input)
```
vermashi: Do you think it is possible to add a quick unit test for this?
You will have to make a new test file and set up the mock for Popen to test CommandExecutor.py
but it might just be easy enough to make it worthwhile.

VMEncryption/main/handle.py
```diff
@@ -180,7 +180,15 @@ def update_encryption_settings():
     logger.log('Updating encryption settings')
 
     # re-install extra packages like cryptsetup if no longer on system from earlier enable
-    DistroPatcher.install_extras()
+    try:
+        DistroPatcher.install_cryptsetup()
```
vermashi: Why are we not installing all extra packages here? (why just cryptsetup?)

VMEncryption/main/CommandExecutor.py
ejarvi: Unit tests for both success and failure cases (a timely return and a timeout) would increase confidence here.

VMEncryption/main/handle.py
ejarvi: I am also curious why this changed from install_extras to install_cryptsetup?

ejarvi: One idea:  For dual pass we may also need to install packages to support AAD auth.  On single pass this is not necessary and only needs crypt_setup.  So this line might have been inadvertently copied over from SP change?

vimish:
@vermashi @ejarvi 
Added unit test and fixed install_cryptsetup from singlepass

PR #795, opened 2019-04-29
Fix for the get-status bug

vermashi:


VMEncryption/main/DiskUtil.py
```diff
@@ -647,38 +647,40 @@ def get_encryption_status(self):
         }
 
         mount_items = self.get_mount_items()
+        device_items = self.get_device_items(None)
+        device_items_dict = {di.mount_point: di for di in device_items}
```
mayank88mahajan: nit: rename to device_item

VMEncryption/main/DiskUtil.py
```diff
@@ -647,38 +647,40 @@ def get_encryption_status(self):
...
-                not "/" == mount_item["dest"] and \
-                not "/oldroot/mnt/resource" == mount_item["dest"] and \
-                not "/oldroot/boot" == mount_item["dest"] and \
-                not "/oldroot" == mount_item["dest"] and \
-                not "/mnt/resource" == mount_item["dest"] and \
-                not "/boot" == mount_item["dest"]:
+            device_item = device_items_dict.get(mount_item["dest"])
 
+            if device_item is not None and \
+               mount_item["fs"] in ["ext2", "ext4", "ext3", "xfs"] and \
```
mayank88mahajan: >["ext2", "ext4", "ext3", "xfs"] [](start = 35, length = 31)

not sure where else is this used but maybe have a supported_file_system list?
eg: mount_item["fs"] in supported_file_systems

VMEncryption/main/DiskUtil.py
```diff
@@ -647,38 +647,40 @@ def get_encryption_status(self):
...
+               self.is_data_disk(device_item, special_azure_devices_to_skip):
                 data_drives_found = True
```
mayank88mahajan: indentation seems wrong? 
do we have unit tests added for this?

VMEncryption/main/DiskUtil.py
```diff
@@ -1091,7 +1093,24 @@ def is_os_disk_lvm(self):
...
+            return False
+        # Resource Disk. Not considered a "data disk" exactly (is not attached via portal and we have a separate code path for encyrpting it)
+        if device_item.device_id.startswith('00000000-0001'):
+            self.logger.log(msg="skipping resource disk", level=CommonVariables.WarningLevel)
+            return False
+
+        for azure_blk_item in azure_devices:
+            if azure_blk_item.name == device_item.name:
+                self.logger.log(msg="the mountpoint is the azure disk root or resource, so skip it.")
+                return False
```
mayank88mahajan: why are there 2 distinct ways to check for root disk and resource disks?

VMEncryption/main/DiskUtil.py
ejarvi: +1 to Mayank's suggestion, also we have had customer requests to add the supported fs list to the public facing documentation (which would be good to queue up in parallel with this change)

VMEncryption/main/DiskUtil.py
```diff
@@ -1091,7 +1093,24 @@ def is_os_disk_lvm(self):
 
         return DiskUtil.os_disk_lvm
 
-    def should_skip_for_inplace_encryption(self, device_item, encrypt_volume_type):
+    def is_data_disk(self, device_item, azure_devices):
+        # Root disk
+        if device_item.device_id.startswith('00000000-0000'):
+            self.logger.log(msg="skipping root disk", level=CommonVariables.WarningLevel)
+            return False
+        # Resource Disk. Not considered a "data disk" exactly (is not attached via portal and we have a separate code path for encyrpting it)
```
ejarvi: encyrpting -> encrypting

VMEncryption/main/DiskUtil.py
vermashi: Done

VMEncryption/main/DiskUtil.py
vermashi: Done

VMEncryption/main/DiskUtil.py
vermashi: We don't yet have unit tests for this file. Some more refactoring is needed before we can set those up.

VMEncryption/main/DiskUtil.py
vermashi: Just to safe guard against one of the methods not working in Gen 2.
Doesn't hurt to have two ways even for Gen 1.

VMEncryption/main/DiskUtil.py
vermashi: Fixed

VMEncryption/main/Common.py
```diff
@@ -57,6 +57,8 @@ class CommonVariables:
     default_block_size = 52428800
     min_filesystem_size_support = 52428800 * 3
     default_file_system = 'ext4'
+    format_supported_file_systems = ['ext4', 'ext3', 'ext2', 'xfs', 'btrfs']
```
ejarvi: will there be a corresponding update to public documentation that needs to be made along with this change

VMEncryption/main/DiskUtil.py
```diff
@@ -1091,7 +1087,24 @@ def is_os_disk_lvm(self):
...
+            self.logger.log(msg="skipping root disk", level=CommonVariables.WarningLevel)
+            return False
+        # Resource Disk. Not considered a "data disk" exactly (is not attached via portal and we have a separate code path for encrypting it)
+        if device_item.device_id.startswith('00000000-0001'):
+            self.logger.log(msg="skipping resource disk", level=CommonVariables.WarningLevel)
+            return False
+
+        for azure_blk_item in azure_devices:
+            if azure_blk_item.name == device_item.name:
+                self.logger.log(msg="the mountpoint is the azure disk root or resource, so skip it.")
```
ejarvi: If the device was root or resource, wouldn't we have already logged that we skipped it in the prior two if conditions?   The devices listed in azure_devices are not known to this function, maybe the log message can inject the device item name it is skipping into the message.  Maybe this skips the BEK volume?

VMEncryption/main/DiskUtil.py
```diff
@@ -1102,14 +1115,8 @@ def should_skip_for_inplace_encryption(self, device_item, encrypt_volume_type):
...
-        if encrypt_volume_type.lower() == 'data':
-            self.logger.log(msg="enabling encryption for data volumes", level=CommonVariables.WarningLevel)
-            if device_item.device_id.startswith('00000000-0000'):
-                self.logger.log(msg="skipping root disk", level=CommonVariables.WarningLevel)
-                return True
-            if device_item.device_id.startswith('00000000-0001'):
-                self.logger.log(msg="skipping resource disk", level=CommonVariables.WarningLevel)
-                return True
+        if encrypt_volume_type.lower() == 'data' and not self.is_data_disk(device_item, special_azure_devices_to_skip):
+            return True # Skip data disks
```
ejarvi: A bit confused by the comment "Skip data disks".   I would have expected it to say "Skip non-data disk"

VMEncryption/main/handle.py
```diff
@@ -883,7 +883,7 @@ def encrypt_inplace_without_separate_header_file(passphrase_file,
             logger.log(msg="the current phase is " + str(CommonVariables.EncryptionPhaseBackupHeader),
                        level=CommonVariables.InfoLevel)
             
-            if not ongoing_item_config.get_file_system().lower() in ["ext2", "ext3", "ext4"]:
+            if not ongoing_item_config.get_file_system().lower() in CommonVariables.inplace_supported_file_systems:
                 logger.log(msg="we only support ext file systems for centos 6.5/6.6/6.7 and redhat 6.7",
```
ejarvi: Can we also change the wording of the log message so that it does not specifically identify centos and rhel distros and version numbers?   This has triggered support questions in the past for versions now out of support.  For example something generic like "[" + ongoing_item_config.get_file_system().lower() + "] is not a supported file system type for Azure Disk Encryption"

VMEncryption/main/DiskUtil.py
ejarvi: Commented before reading your prior remark about Gen 1 and Gen 2.   So maybe add that as a comment in the code that the first checks are for Gen1 and the second checks are for Gen2 images.   Also ... Is there any other type of special device here other than root or resource that might be included in this list?  It is passed dynamically to the function. Would like to make sure BEK volume is not going to be misidentified as a data disk.

VMEncryption/main/DiskUtil.py
vermashi: sorry, yeah, it should be "non-data". Will fix.

VMEncryption/main/Common.py
vermashi: I don't think so. This is just a list of formats we are comfortable running mkfs.[format] with.

PR #798, opened 2019-05-06
LAD: SLES and OpenSUSE 15 distro support

canfikret:
Python 2.7 doesn't recognize SUSE 15 since it was released before the distro. Fall-back /etc/os-release to detect.

### DIFF
--- a/Diagnostic/DistroSpecific.py
+++ b/Diagnostic/DistroSpecific.py
@@ -326,6 +326,7 @@ DistroMap = {
     'centos': CentosActions, 
     'oracle': RedhatActions,
     'SuSE:12': Suse12Actions,
+    'SuSE': Suse12Actions,
     'sles:15': Suse12Actions,
     'opensuse:15':Suse12Actions
 }
--- a/Diagnostic/DistroSpecific.py
+++ b/Diagnostic/DistroSpecific.py
@@ -325,8 +325,9 @@ DistroMap = {
     'redhat': RedhatActions, 
     'centos': CentosActions, 
     'oracle': RedhatActions,
-    'SuSE:12': Suse12Actions, 
-    'SuSE': Suse12Actions
+    'SuSE:12': Suse12Actions,
+    'sles:15': Suse12Actions,
+    'opensuse:15':Suse12Actions
 }
 
 
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -84,7 +84,25 @@ def init_distro_specific_actions():
     global g_dist_config, RunGetOutput
     dist = platform.dist()
     try:
-        g_dist_config = DistroSpecific.get_distro_actions(dist[0], dist[1], hutil.log)
+        if dist[0] != '':
+            name = dist[0]
+            version = dist[1]
+        else:
+            try:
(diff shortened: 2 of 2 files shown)


canfikret:
@simathih could you review?

Diagnostic/diagnostic.py
```diff
@@ -84,7 +84,25 @@ def init_distro_specific_actions():
...
     dist = platform.dist()
     try:
-        g_dist_config = DistroSpecific.get_distro_actions(dist[0], dist[1], hutil.log)
+        if dist[0] != '':
+            name = dist[0]
+            version = dist[1]
+        else:
+            try:
+                # platform.dist() in python 2.7.15 does not recognize SLES/OpenSUSE.
+                with open("/etc/os-release") as fp:
```
AaronRM: Ideally it should [fallback to /usr/lib/os-release](https://en.opensuse.org/SDB:Find_openSUSE_version) if /etc/os-release is not available.

Diagnostic/DistroSpecific.py
```diff
@@ -325,8 +325,9 @@ def install_required_packages(self):
     'redhat': RedhatActions, 
     'centos': CentosActions, 
     'oracle': RedhatActions,
-    'SuSE:12': Suse12Actions, 
-    'SuSE': Suse12Actions
```
simathih: are we removing Suse?

Diagnostic/diagnostic.py
```diff
@@ -84,7 +84,25 @@ def init_distro_specific_actions():
...
+                    line = fp.readline()
```
AaronRM: According to [the documentation](https://en.opensuse.org/SDB:Find_openSUSE_version), we should not be:

1. Assuming that the first two lines of /etc/os-release are NAME and VERSION. (As far as I can tell, no order is guaranteed)
2. Relying on NAME or VERSION as these [are meant to be human-readable fields](https://en.opensuse.org/SDB:Find_openSUSE_version). Instead, we should be looking at ID and VERSION_ID.

Diagnostic/DistroSpecific.py
```diff
@@ -325,8 +325,9 @@ def install_required_packages(self):
...
+    'SuSE:12': Suse12Actions,
+    'sles:15': Suse12Actions,
+    'opensuse:15':Suse12Actions
```
AaronRM: Do we need to be concerned with older versions as well?

Diagnostic/diagnostic.py
canfikret: this PR is to enable SuSE 15; which has /etc/os-release ...

Fican-SLES15:/var/lib/waagent # ls /usr/lib/os-release
ls: cannot access '/usr/lib/os-release': No such file or directory

Some distros use /usr/lib/os-release; but usually it is a symlink to /etc/os-release

Diagnostic/DistroSpecific.py
canfikret: SuSE 12 is already in the support matrix. (It is the lowest we support)

Diagnostic/diagnostic.py
canfikret: Thx. Let me read the man page.

PR #799, opened 2019-05-08
fix for python 2.6 break and writing to stdout

simathih:
- sys.version.major is not available in versions <2.7
- output message to stdout which waagent picks up

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -427,7 +427,7 @@ def enable():
         os.chown(etc_final_path, uid, gid)
 
         # octal numbers are represented differently in python 3
-        if sys.version_info.major > 2:
+        if sys.version_info >= (3,):
             os.chmod(etc_final_path, 0o750)
         else:    
             os.chmod(etc_final_path, 0750)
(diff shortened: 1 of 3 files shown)


OmsAgent/omsagent.py
```diff
@@ -943,6 +943,16 @@ def run_command_and_log(cmd, check_error = True, log_cmd = True):
         hutil_log_info('Output of command "{0}": \n{1}'.format(cmd, output))
     else:
         hutil_log_info('Output: \n{0}'.format(output))
+
+    # also write output to STDOUT since WA agent uploads that to Azlinux Kusto DB
+    try:
+        if sys.version_info >= (3,):
```
abenbachir: use print(output) it works on both python 2.x and 3.x

simathih: as per doc, print with parantheses is supported only in python 3

simathih: print in python 2 doesn't take parentheses https://docs.python.org/2/tutorial/inputoutput.html

simathih: print in python 2 doesn't take parentheses https://docs.python.org/2/tutorial/inputoutput.html

PR #801, opened 2019-05-09
Resource Disk reboot fix [SinglePass]

vermashi:
Includes:
- Consistent name for encrypted resource disk's device mapper
- Refactoring of tests
- New Tests for Resource disk
- Refactoring for resource disk code


VMEncryption/test/test_resource_disk_util.py
```diff
@@ -0,0 +1,199 @@
...
+from main.Common import CommonVariables
+from console_logger import ConsoleLogger
+
+
+class TestResourceDiskUtil(unittest.TestCase):
+    def setUp(self):
+        self.logger = ConsoleLogger()
+        self.mock_disk_util = mock.create_autospec(DiskUtil)
+        self.mock_passhprase_filename = "mock_passphrase_filename"
+        mock_public_settings = {}  # update this before tests
```
ejarvi: in reference to the "# update this before tests" comment - is this a to-do item?  maybe some commented samples of what to set this to if necessary

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -18,134 +18,90 @@
...
-                public_settings = public_settings_str
-            encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
-            if encryption_operation in [CommonVariables.EnableEncryptionFormatAll]:
-                return True
-        except:
-            self.logger.log("unable to identify current encryption operation")
+        encryption_operation = self.public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
+        if encryption_operation in [CommonVariables.EnableEncryptionFormatAll]:
+            return True
+        self.logger.log("unable to identify current encryption operation")
```
ejarvi: Wording of this log message may be confusing to someone reading the log later, since we did pull an encryption operation it just wasn't an encrypt format all operation.   Maybe change to "current encryption operation is not EnableEncryptionFormatAll"

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -154,109 +110,156 @@ def configure_waagent(self):
...
+            if mount_item["dest"] == self.RD_MOUNT_POINT and (mount_item["src"].startswith(CommonVariables.dev_mapper_root) or mount_item["src"].startswith(self.DEV_DM_PREFIX)):
                 return True
         return False
 
-    def get_rd_device_mapper(self):
-        """ retrieve current device mapper path backing the encrypted resource disk mount point """
+    def _get_rd_device_mappers(self):
+        """
+        Retreive any device mapper device on the resource disk (e.g. /dev/dm-0).
+        Can't imagine why there would be multiple device mappers here, but doesn't hurt to handle the case
```
ejarvi: tested against RAID scenarios?

vermashi: Do you mean set up a RAID volume on the resource disk and then see how the code reacts?

VMEncryption/test/test_resource_disk_util.py
vermashi: I think I was expecting to use it during tests, but it turned out ended up not needing to modify it.
I'll just remove the comment.

VMEncryption/main/ResourceDiskUtil.py
vermashi: Will do.

VMEncryption/main/ResourceDiskUtil.py
ejarvi: Yes just a check for any regressions in the case where RAID is set up with a persistent device name scheme similar to /dev/dm-0 .. some now deprecated ARM template demos previously suggested RAID (they were tied to the legacy encrypt format operation).

PR #803, opened 2019-05-09
Used space encryption and CentOS 7.6 support[DualPass]

vimish:


VMEncryption/main/oscrypto/rhel_72/encryptstates/EncryptBlockDeviceState.py
```diff
@@ -25,6 +25,7 @@
 from inspect import ismethod
 from time import sleep
 from OSEncryptionState import *
+from distutils.version import LooseVersion
```
vermashi: Is distutils something that is pre-installed with Python?

vimish: Yes...Also, I am using the same to check for kernel version comparison
https://github.com/Azure/azure-linux-extensions/blob/ade-singlepass-master/VMEncryption/main/check_util.py#L270

vermashi: Nice! Maybe we can use this to simplify all the version checking elsewhere in the code when we get the chance.

vermashi:
Looks good.
Test on Centos 7.6 and Centos 7.6 LVM and then merge it.

Issue #672, opened 2018-10-19
Traceback on SLES


rjschwei:
When trying to encrypt a SLES instance the following traceback is generated:

VM has reported a failure when processing extension 'AzureDiskEncryptionForLinux'. Error message: "Failed to enable the extension with error: 'NoneType' object has no attribute 'getheader', stack trace: Traceback (most recent call last):
  File "/var/lib/waagent/Microsoft.Azure.Security.AzureDiskEncryptionForLinux-0.1.0.999322/main/handle.py", line 691, in enable_encryption
    DiskEncryptionKeyFileName=extension_parameter.DiskEncryptionKeyFileName)
  File "/var/lib/waagent/Microsoft.Azure.Security.AzureDiskEncryptionForLinux-0.1.0.999322/main/KeyVaultUtil.py", line 63, in create_kek_secret
    bearerHeader = result.getheader("www-authenticate")
AttributeError: 'NoneType' object has no attribute 'getheader'

szarkos:
Was this on SLES12-SP3 or SLES15?

szarkos:
Hi Robert,

It seems the VM Encryption extension lacks Python3 support, which may explain this error.  I believe this support is on the backlog.

Thanks,
Steve

rjschwei:
The traceback was from SLES 15, but SLES 12 is not working either, different symptom.

We will help sort this out, I have asked one of my team members, @edlane to look at this

edlane:
I need additional info to recreate this issue.
I got encryption on SLES12 to work with this "az" CLI sequence...
```
ed-sle12sp3byos:/mnt/resource # az keyvault create --name "edlanevault" --resource-group "ed_lane"
```
... then I followed the directions for allowing key vault access for disk encryption described here: https://docs.microsoft.com/en-us/azure/security/azure-security-disk-encryption-prerequisites

... then I enabled encryption on a data disk as follows:
```
ed-sle12sp3byos:/mnt/resource #az vm encryption enable --resource-group ed_lane --name ed-sle12sp3byos --disk-encryption-keyvault edlanevault --volume-type Data
ed-sle12sp3byos:/mnt/resource #az vm encryption show --name ed-sle12sp3byos --resource-group ed_lane
{
  "disks": [
    {
      "encryptionSettings": [
        {
          "diskEncryptionKey": null,
          "enabled": false,
          "keyEncryptionKey": null
        }
      ],
[...]

PR #805, opened 2019-05-15
suse os encryption extensions

edlane:
This PR has been developed and tested **ONLY** by "side-loading" and not directly as a supported Azure Linux Extension feature.

Please review as such.

### DIFF
--- a/VMEncryption/main/DiskUtil.py
+++ b/VMEncryption/main/DiskUtil.py
@@ -293,7 +293,7 @@ class DiskUtil(object):
         """
         self.hutil.log("dev path to cryptsetup luksFormat {0}".format(dev_path))
         #walkaround for sles sp3
-        if self.distro_patcher.distro_info[0].lower() == 'suse' and self.distro_patcher.distro_info[1] == '11':
+        if self.distro_patcher.distro_info[0].lower() == 'suse' and self.distro_patcher.distro_info[1] in ['11','12','15']:
             proc_comm = ProcessCommunicator()
             passphrase_cmd = self.distro_patcher.cat_path + ' ' + passphrase_file
             self.command_executor.Execute(passphrase_cmd, communicator=proc_comm)
@@ -826,7 +826,7 @@ class DiskUtil(object):
         return device_items_to_return
 
     def get_device_items(self, dev_path):
-        if self.distro_patcher.distro_info[0].lower() == 'suse' and self.distro_patcher.distro_info[1] == '11':
+        if self.distro_patcher.distro_info[0].lower() == 'suse' and self.distro_patcher.distro_info[1] in ['11', '12', '15']:
             return self.get_device_items_sles(dev_path)
         else:
             if dev_path:
--- a/VMEncryption/main/ExtensionParameter.py
+++ b/VMEncryption/main/ExtensionParameter.py
@@ -63,6 +63,8 @@ class ExtensionParameter(object):
         """
         private settings
         """
+        if protected_settings is None:
+            protected_settings = {}
         self.AADClientSecret = protected_settings.get(CommonVariables.AADClientSecretKey)
 
         if self.AADClientSecret is None:
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -1627,6 +1627,12 @@ def daemon_encrypt():
                                                              distro_patcher=DistroPatcher,
                                                              logger=logger,
                                                              encryption_environment=encryption_environment)
+        elif distro_name == 'SuSE' and distro_version in ['12', '15']:
+            from oscrypto.suse_12 import Suse12EncryptionStateMachine
+            os_encryption = Suse12EncryptionStateMachine(hutil=hutil,
+                                                             distro_patcher=DistroPatcher,
+                                                             logger=logger,
+                                                             encryption_environment=encryption_environment)
         else:
             message = "OS volume encryption is not supported on {0} {1}".format(distro_name,
                                                                                 distro_version)
--- a/VMEncryption/main/oscrypto/OSEncryptionState.py
+++ b/VMEncryption/main/oscrypto/OSEncryptionState.py
@@ -161,6 +161,11 @@ class OSEncryptionState(object):
         if matches:
             return matches[0]
 
+        # for path-to-UUID style OSes like Suse
+        matches = re.findall(r'/dev/disk/by-uuid/(.*?)\s+{0}\s+'.format(mountpoint), contents)
+        if matches:
+            return matches[0]
+
     def _get_block_device_size(self, dev):
         if not os.path.exists(dev):
             return 0
--- a/VMEncryption/main/oscrypto/suse_12/encryptstates/EncryptBlockDeviceState.py
+++ b/VMEncryption/main/oscrypto/suse_12/encryptstates/EncryptBlockDeviceState.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+#
+# VM Backup extension
+#
+# Copyright 2015 Microsoft Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.7+
+#
+
+import os
+import sys
+
+from inspect import ismethod
+from time import sleep
+from OSEncryptionState import *
+
+class EncryptBlockDeviceState(OSEncryptionState):
+    def __init__(self, context):
+        super(EncryptBlockDeviceState, self).__init__('EncryptBlockDeviceState', context)
+
+    def should_enter(self):
+        self.context.logger.log("Verifying if machine should enter encrypt_block_device state")
+
+        if not super(EncryptBlockDeviceState, self).should_enter():
+            return False
+        
+        self.context.logger.log("Performing enter checks for encrypt_block_device state")
+                
+        return True
+
+    def enter(self):
+        if not self.should_enter():
+            return
+
+        self.context.logger.log("Entering encrypt_block_device state")
+        
+        self.command_executor.Execute('mount /boot', False)
+        self._find_bek_and_execute_action('_dump_passphrase')
+        self._find_bek_and_execute_action('_luks_format')
+        self._find_bek_and_execute_action('_luks_open')
+
+        self.context.hutil.do_status_report(operation='EnableEncryptionDataVolumes',
+                                            status=CommonVariables.extension_success_status,
+                                            status_code=str(CommonVariables.success),
+                                            message='OS disk encryption started')
+
+        self.command_executor.Execute('dd if={0} of=/dev/mapper/osencrypt bs=52428800'.format(self.rootfs_block_device), True)
+
+    def should_exit(self):
+        self.context.logger.log("Verifying if machine should exit encrypt_block_device state")
+
+        if not os.path.exists('/dev/mapper/osencrypt'):
+            self._find_bek_and_execute_action('_luks_open')
+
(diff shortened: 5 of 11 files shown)


aosthof:
@vermashi Hi, I'm not sure if you're the right person but as you already provided the answers in https://github.com/Azure/azure-linux-extensions/issues/672 could you or somebody on your team review Ed's PR? BTW, what is the regular process of getting PRs reviewed so we can follow your processes in the future? Thank you.

aosthof:
@szarkos Steve, could you please help us understand the regular process of getting PRs reviewed so we can follow your processes in the future? Thank you.

szarkos:
@aosthof, I'll defer to the VM encryption team for any changes to the process, but in general you are doing everything right :)  We will discuss this PR next week when Robert is on campus.

Thanks!
Steve

aosthof:
@szarkos OK, that sounds good to me, thanks.

ejarvi:
@aosthof  Thanks for this contribution - I'd like to help champion getting this into a test build and a release build.  At a high level, here is a bit of information on how things are structured right now and that I think are important to making sure we are on the same page for moving this forward: 

First, there are currently two versions of the Microsoft.Azure.Security.AzureDiskEncryptionForLinux extension being built out of two separate branches and being maintained for customers. 
- **Version 0.1.n.n**  is built from (master)[ https://github.com/Azure/azure-linux-extensions/tree/master ade-singlepass-master] and requires the user to provide AAD credentials with access to the key vault when being enabled.  This version is sometimes referred to as "dual pass" because persisting encryption settings requires two hops that occur under the hood of the Powershell cmdlet that enables encryption.  
- **Version 1.1.n.n** is built from [ade-singlepass-master](https://github.com/Azure/azure-linux-extensions/tree/ade-singlepass-master).  This version simplifies the architecture by no longer requiring the user to provide AAD credentials at time of enable.   This model is sometimes referred to as "single pass" because unlike the prior architecture, enabling encryption only requires one step.  

The two versions are significantly different in terms of architecture dependencies and how the encryption settings are stored, so converting between one and the other is not possible.  

At time of this post the 0.1 extension version (dual pass, AAD) is being maintained for bug fixes and for users who have already installed it.   For new deployments (and new features) migration to the 1.1 (single pass, no-AAD) version is recommended. 

For testing changes, we've set up two branches that can be used as first step in a pipeline to move changes into test and then release.   The recommended steps would be to take these changes and submit the PR to one or both of the following branches: 

- 0.1 dual pass : https://github.com/Azure/azure-linux-extensions/tree/ade-dualpass-dev
- 1.1 single pass : https://github.com/Azure/azure-linux-extensions/tree/ade-singlepass-dev

Once the PR is targeting one or both of those branches, we can comment on the PR (from what I'm seeing it looks pretty good).  Once that PR is approved will create and publish a build of the test extension (which will be named Microsoft.Azure.Security.**Edp**.AzureDiskEncryptionForLinux that can be used for testing.   Finally, once those tests prove stable, we can then merge the changes from dev to master and deploy to production.  

If you have any questions that you'd like to ask outside of GitHub I'd be happy to set up a call to walk through this next week.

aosthof:
@ejarvi Thanks for sharing this detailed information, appreciated!

@edlane Please target your PR at the two dev branches mentioned above and work with the team at MSFT on the review. Thanks.

edlane:
@ejarvi,
Apparently an update for the Linux VMEncryption was released to the Azure USWest Region last week.  It looks like the update went from **Microsoft.Azure.Security.AzureDiskEncryptionForLinux-1.1.0.39** to **1.1.0.40**
This update seemed to happen in the background without notice to those doing development on the Azure platform.

**Question**: is there a way I can be advised when an Azure Linux Extension has been released or "about to be released"?  Because I must currently "side-load" code under development for the Linux Extensions this causes confusion and debugging issues when a release happens during code development.

Thanks.

ejarvi:
@edlane Great suggestion, and sorry for the growing pains on our side.   Until now our release notifications have been done with a private email list we had set up but going forward I believe we should move to posting this on Github alongside of the project.  We are currently testing a .41 deployment in our canary regions East US 2 EUAP and Central US 2 EUAP.  Results of that testing are not complete yet.   In the meantime, I'll give a heads up to other team members on this so that we can work on a better way of getting this information to you.

edlane:
@ejarvi 
Today it looks like version **Microsoft.Azure.Security.AzureDiskEncryptionForLinux-1.1.0.43** was pushed which only caused a 2 hour adventure before discovering what happened to my development instance.  I'm not sure what changes have been made to that version or if any additional code is required by me for compatibility.

My earlier question asked here is still unresolved:
https://github.com/Azure/azure-linux-extensions/pull/805#issuecomment-530093875

Thanks

ejarvi:
Will you be at the SUSE technical review later today?   If so it would be great to discuss then.  Short term I can look into adding you to our email release notification email for now until we have a way to update Github with release notifications and would be the long term solution.  Also, to minimize future impact of version updates, we may be able to set up a temporary test extension for you with a different name, so your test environment does not rely on the production extension and can be sheltered from future production release version updates.

edlane:
@ejarvi ,
Thanks for the prompt reply.
I will not be at the technical review today.  I have communicated my "Challenges and Recommendations" concerning VMEncryption to Robert and Dominic who will be at that meeting.  Feel free to share your own experiences with developing for the VMEncryption code base with them.  I look forward to a more collaborative development experience in the future with you and your team -- Thanks!

ejarvi:
@edlane - I'm also looking forward to making this a better experience for you.  I'll share the test extension approach with Robert and Dominic later today.

edlane:
@ejarvi 
Is there a description somewhere of the "test extension approach" you mentioned?

ejarvi:
@edlane  It's not CI/CD ... but if you prepare a branch in your fork with changes, synch to azure/ade-singlepass-dev, and then start a buildable PR from that, I can personally review your changes and then build an extension version out of your repo and temporarily deploy it into one of our test slots with a name like **Microsoft.Azure.Security.Edp**._AzureDiskEncryptionForLinuxTest3_.    To test this, you will override the default extension name with this test extension name when enabling encryption (there are environment variables and/or parameters to do this depending if you are using CLI or Powershell).  Also for more lightweight change testing with faster turnaround you can do tests locally within that VM which I think is similar to the approach you're currently using.  Then update your PR and I'll rebuild/redeploy for you.  Thanks to the different extension names, this will allow testing of changes in a way that is independent of version changes happening in production.   If it would help to talk through this I'd be happy to set up a call.

edlane:
@ejarvi 
I am nearly ready for a test run through your test pipeline but I need some advice before submitting a PR.
SUSE distros take this code path as part of the VMEncryption extension:
https://github.com/Azure/azure-linux-extensions/blob/master/VMEncryption/main/oscrypto/OSEncryptionState.py#L86-L89
That was based on an old disk layout for SUSE prior to our recent moving to UEFI-based images.

Currently the base disk format for a SUSE image in Azure is:
```
adeceqkquvm:/home/lane # lsblk
NAME                                   MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
fd0                                      2:0    1    4K  0 disk  
sda                                      8:0    0   30G  0 disk  
├─sda1                                   8:1    0    2M  0 part  
├─sda2                                   8:2    0  200M  0 part  /boot/efi
├─sda3                                   8:3    0    1G  0 part  /boot
└─sda4                                   8:4    0 28.8G  0 part  /
sdb                                      8:16   0   16G  0 disk  
└─sdb1                                   8:17   0   16G  0 part  /mnt/resource
sdc                                      8:32   0   48M  0 disk  
└─sdc1                                   8:33   0   47M  0 part  /mnt/azure_bek_disk
sdd                                      8:48   0    1G  0 disk  
└─3f53b453-ff93-4817-a60f-bbf664c94ce5 254:0    0 1022M  0 crypt 
sde                                      8:64   0    1G  0 disk  
sr0                                     11:0    1 1024M  0 rom   
```
The existing master branch code assumes that 
`self.rootfs_block_device = '/dev/sda2'`
`self.bootfs_block_device = '/dev/sda1'`

**BUT THE NEW DISK LAYOUT IS:**
`self.rootfs_block_device = '/dev/sda4'`
`self.bootfs_block_device = '/dev/sda3'`

**Question:** 
Because this code is also used by other "NON-SUSE" distros I cannot confidently just change these values without potentially breaking other distros.
What remedy do you suggest?

In my local test environment hacking these values to the new locations causes the encryption to succeed -- WOOT!

ejarvi:
@edlane  Great progress and thanks for the clear explanation of the issue, I think we'll want to update the OSEncryptionState.py code to identify the proper device name, ideally we would not have to fall to these default name assignments anyway.  

Earlier in the code after this line: 

https://github.com/Azure/azure-linux-extensions/blob/master/VMEncryption/main/oscrypto/OSEncryptionState.py#L61 

we could potentially add a line like the following: 
`self.bootfs_sdx_path = self._get_fs_partition("/boot")[0]`

and then refactor any subsequent logic to ensure the proper path is being taken. 

@vermashi has been developing a mock layer to test various disk layouts that he believes we could put to use when making this change to make sure it is compatible across distros

edlane:
@ejarvi 
Encountering an issue with Suse's latest SLES15 images...
SLES15 uses the newer linux `/etc/os-release` file to identify itself rather than the now deprecated `/etc/suse-release` file.

The `platform.linux_distribution()` library call used by the VMEncryption has also been deprecated in Python after 3.7 and currently fails for distributions using the newer `os-release` file.  This call fails with the empty list `('', '', '')` for all recently released images.
see https://github.com/Azure/azure-linux-extensions/blob/master/VMEncryption/main/patch/__init__.py#L37

There is another method currently used by the **Azure waagent** that works for all images I have tested including SLES15. see: https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/common/version.py#L76
```
Python 3.6.8 (default, Apr 30 2019, 13:27:23) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import azurelinuxagent.common.version
>>> azurelinuxagent.common.version.get_distro()
['sles', '15', '', 'SLES']
```

Unfortunately we cannot call the`azurelinuxagent` library from within VMEncryption because the waagent runs in python3 and the VMEncryption is forced to run in python2 by this waagent code:

https://github.com/Azure/WALinuxAgent/blob/develop/bin/waagent#L41

**Questions:**
1. Are there plans for transitioning from the deprecated `platform.linux_distribution()` libraries used by VMEncryption to a supported approach?
2. Is python3 support planned for VMEncryption module?
3. Do you have a suggestion for OS's that currently use the `/etc/os-release` method for Linux OS identification?

aosthof:
@ejarvi Hi Eric, I like to follow up with you on the questions raised by @edlane in the last comment. Did you have a chance to look into that and get answers to those questions, maybe involving your team? Resolutions to these issues are the only things preventing @edlane from submitting a PR to the Azure Encryption Repo.

ejarvi:
@aosthof  @edlane  -  Moving away from platform.linux_distribution() sounds reasonable in this case.   The linux agent team would prefer that we not take dependencies on their common library going forward (even though this has been standard practice in the past across other extensions) so the safest thing may be to replace this functionality with a common function in our extension.  If so and it is possible to do that within the existing Python 2.7 codebase, I would have no objection. 

In parallel to this I am working on a port for python3 compatibility in a private branch, that follows the strategy of achieving Python 3 functionality first and then introducing compatibility options for compatibility when running with Python 2.7.  

https://github.com/ejarvi/azure-linux-extensions/tree/ade-sp-dev-p3  (this is not yet stable, test iterations are published to the Microsoft.Azure.Security.Edp.AzureDiskEncryptionForLinuxTest2 slot).

edlane:
@ejarvi 
Thanks for the response.
Based on your feedback I will proceed with these assumptions:
1. a "common" function shared among all supported encryption distros and versions.
2. function is added to the Encryption module and NOT shared with the Linux Agent Libraries
3. function is python 2.7 compatible

**NOTE:**
The inclusion of this shared function should include testing on all linux encryption supported distros RHEL, Centos, Ubuntu.. -- not only Suse

ejarvi:
@edlane Yes, I agree that sounds like the best path forward.  Python 2.6 compatibility may be the needed target in this case for legacy RHEL/CentOS 6.8 distributions.

vimish:
@edlane 
Just wanted to notify that a new version of ADE Linux extension (1.1.0.46) is under deployment now.
It will hit all PROD regions by next week.

PR #822, opened 2019-06-04
Fix for the encryption status [DualPass]

vermashi:


VMEncryption/main/Common.py
```diff
@@ -36,6 +36,8 @@ class CommonVariables:
     min_filesystem_size_support = 52428800 * 3
     #TODO for the sles 11, we should use the ext3
     default_file_system = 'ext4'
+    format_supported_file_systems = ['ext4', 'ext3', 'ext2', 'xfs', 'btrfs']
```
ejarvi: does this trigger a need to update public documentation around xfs and btrfs support

VMEncryption/main/DiskUtil.py
```diff
@@ -957,7 +956,24 @@ def is_os_disk_lvm(self):
...
+        # Root disk
+        if device_item.device_id.startswith('00000000-0000'):
+            self.logger.log(msg="skipping root disk", level=CommonVariables.WarningLevel)
+            return False
+        # Resource Disk. Not considered a "data disk" exactly (is not attached via portal and we have a separate code path for encrypting it)
+        if device_item.device_id.startswith('00000000-0001'):
+            self.logger.log(msg="skipping resource disk", level=CommonVariables.WarningLevel)
+            return False
+
+        for azure_blk_item in azure_devices:
```
ejarvi: same as single pass -- would like to see a comment making it clear the first checks are for gen 1 images and the second are for gen 2 images, also if there is any chance of additional devices beyond root and resource disk being provided since it is being passed as a parameter.   final case log message could be updated to reflect the exact device it is skipping by name in the log message instead of ambiguous "azure disk root or resource"

VMEncryption/main/DiskUtil.py
```diff
@@ -968,14 +984,8 @@ def should_skip_for_inplace_encryption(self, device_item, encrypt_volume_type):
...
-        if encrypt_volume_type.lower() == 'data':
-            self.logger.log(msg="enabling encryption for data volumes", level=CommonVariables.WarningLevel)
-            if device_item.device_id.startswith('00000000-0000'):
-                self.logger.log(msg="skipping root disk", level=CommonVariables.WarningLevel)
-                return True
-            if device_item.device_id.startswith('00000000-0001'):
-                self.logger.log(msg="skipping resource disk", level=CommonVariables.WarningLevel)
-                return True
+        if encrypt_volume_type.lower() == 'data' and not self.is_data_disk(device_item, special_azure_devices_to_skip):
+            return True # Skip data disks
```
ejarvi: I think this comment should be skip non-data disks?

VMEncryption/main/handle.py
```diff
@@ -932,7 +931,7 @@ def encrypt_inplace_without_seperate_header_file(passphrase_file,
             logger.log(msg="the current phase is " + str(CommonVariables.EncryptionPhaseBackupHeader),
                        level=CommonVariables.InfoLevel)
 
-            if not ongoing_item_config.get_file_system().lower() in ["ext2", "ext3", "ext4"]:
+            if not ongoing_item_config.get_file_system().lower() in CommonVariables.inplace_supported_file_systems:
                 logger.log(msg="we only support ext file systems for centos 6.5/6.6/6.7 and redhat 6.7",
```
ejarvi: while in this area can we also update the log message issued in the next line to be more generic and not name specific distros and versions to be more future proof (this has led to support questions since some of those are no longer on our support list) - see same comment made on the single pass PR

PR #826, opened 2019-06-14
Update watcherutil.py to transfer omsconfighost telemetry files through individual files

zjalali:


zjalali:
Since we decided to take another approach, please ignore this PR for the moment.

PR #828, opened 2019-06-14
Fix update encryption settings flow

vimish:
1. Remove new luks keys if update operation fails.
2. Remove reboot from singlepass update settings flow.
3. Commit encryption parameters after encryption settings are stamped.
4. Keep all passphrase files in sync when updating passphrase.
5. Fix multiple wireserver requests send by Linux extension.


vimish:
Updated the PR based on an offline conversation with shiv to not update crypt item when only luks keyslot changes.

PR #835, opened 2019-06-25
Update watcherutil.py to cleanup telemetry status under /var/opt/microsoft/omsconfig/status

abenbachir:


zjalali:
looks good :)

PR #838, opened 2019-06-26
Miscellaneous bug fixes

vermashi:


VMEncryption/main/handle.py
```diff
@@ -203,7 +203,9 @@ def stamp_disks_with_settings(items_to_encrypt, encryption_config):
                         filenames.append(str(tag["Value"]))
 
     for filename in filenames:
-        shutil.copy(current_passphrase_file, os.path.join(CommonVariables.encryption_key_mount_point, filename))
+        filepath = os.path.join(CommonVariables.encryption_key_mount_point, filename)
```
ejarvi: When we pull filename from the DiskEncryptionKeyFileName tag, will it ever contain slashes?  The way os.path.join works, if subsequent parameters after the root contain a slash, the root is ignored and the join may not work as anticipated.  ( see https://stackoverflow.com/a/1945930 )

VMEncryption/main/handle.py
```diff
@@ -203,7 +203,9 @@ def stamp_disks_with_settings(items_to_encrypt, encryption_config):
...
+        if filename != current_passphrase_file:
+            shutil.copy(current_passphrase_file, filepath)
```
ejarvi: - Is destination filepath always mounted before this copy is attempted?    Asking because of an shutil failure I saw in earlier run of the SpLnxEnableDataNoKekMdLvmUpdateStopStart testcase:   ` [Errno 2] No such file or directory: '/mnt/azure_bek_disk/LinuxPassPhraseFileName', stack trace: Traceback (most recent call last): File "main/handle.py", line 322, in update_encryption_settings shutil.copy`

- When destination filepath is FAT file system, shutil might not be reliable for copy on all distros.  I haven't tested it myself but I saw one report of this failing because of a chmod/vfat quirk in the shutil implementation (see https://stackoverflow.com/questions/20020177/python-shutil-copy-fails-on-fat-file-systems-ubuntu )

VMEncryption/main/handle.py
vermashi: The tag should never have slashes, so we probably don't need to be worried about this.

VMEncryption/main/handle.py
vermashi: `current_passphrase_file = bek_util.get_bek_passphrase_file(encryption_config)`
is called a few lines prior, which mounts the BEK VOLUME if its not mounted..
Though we do need to do a bugfix to make that method a hard failure if the BEK VOLUME is missing. (Future work item)

Changed the method to "copyfile" as suggested in that Answer.
Thanks!

PR #840, opened 2019-06-28
Bug fixes ported over from single pass branch

vermashi:


VMEncryption/main/DiskUtil.py
```diff
@@ -306,12 +306,13 @@ def remove_crypt_item(self, crypt_item):
 
             filtered_mount_lines = []
             with open(crypt_file_path, 'r') as f:
+                self.logger.log("removing an entry from {0}".format(crypt_file_path))
```
ejarvi: Not sure if this is needed.  Might be misleading in some cases because it will print "removing an entry" even if it turns out there is nothing to remove and the entries are kept.

PR #841, opened 2019-07-02
Update encryption settings fixes

vimish:
1. Created a new upper function to decide between enable and update flow
2. Handled the case when new disks needs to be encrypted in update flow.
3. Block enable if volume type transition is not allowed.

### DIFF
--- a/VMEncryption/main/check_util.py
+++ b/VMEncryption/main/check_util.py
@@ -273,7 +273,24 @@ class CheckUtil(object):
                             return
             raise Exception('Distro {0} {1} is not supported for OS encryption'.format(distro_name, distro_version))
 
-    def precheck_for_fatal_failures(self, public_settings, encryption_status, DistroPatcher):
+    def validate_volume_type_for_enable(self, public_settings, existing_volume_type):
+        encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
+        if not encryption_operation in [CommonVariables.EnableEncryption, CommonVariables.EnableEncryptionFormat, CommonVariables.EnableEncryptionFormatAll]:
+            self.logger.log("Current operation is not an enable. Skipping volume type validation.")
+            return
+        if not existing_volume_type:
+            self.logger.log('Existing volume type not found. First enable.')
+            return
+        volume_type = public_settings.get(CommonVariables.VolumeTypeKey)
+        if volume_type.lower() == existing_volume_type.lower():
+            self.logger.log('No change in volume type.')
+            return
+        if volume_type.lower() == CommonVariables.VolumeTypeAll.lower():
+            self.logger.log('Upgrading volume type from {0} to {1}'.format(existing_volume_type, volume_type))
+            return
+        raise Exception('Moving from volume type {0} to volume type {1} is not allowed'.format(existing_volume_type, volume_type))
+
+    def precheck_for_fatal_failures(self, public_settings, encryption_status, DistroPatcher, existing_volume_type):
         """ run all fatal prechecks, they should throw an exception if anything is wrong """
         self.validate_key_vault_params(public_settings)
         self.validate_volume_type(public_settings)
@@ -281,6 +298,7 @@ class CheckUtil(object):
         self.validate_vfat()
         self.validate_memory_os_encryption(public_settings, encryption_status)
         self.is_supported_os(public_settings, DistroPatcher, encryption_status)
+        self.validate_volume_type_for_enable(public_settings, existing_volume_type)
 
     def is_non_fatal_precheck_failure(self):
         """ run all prechecks """
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -234,7 +234,7 @@ def get_protected_settings():
         return protected_settings_str
 
 
-def update_encryption_settings():
+def update_encryption_settings(extra_items_to_encrypt=[]):
     hutil.do_parse_context('UpdateEncryptionSettings')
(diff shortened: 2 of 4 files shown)


VMEncryption/main/check_util.py
```diff
@@ -273,14 +273,31 @@ def is_supported_os(self, public_settings, DistroPatcher, encryption_status):
...
+        encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
+        if not encryption_operation in [CommonVariables.EnableEncryption, CommonVariables.EnableEncryptionFormat, CommonVariables.EnableEncryptionFormatAll]:
+            self.logger.log("Current operation is not an enable. Skipping volume type validation.")
+            return
+        if not existing_volume_type:
+            self.logger.log('Existing volume type not found. First enable.')
+            return
+        volume_type = public_settings.get(CommonVariables.VolumeTypeKey)
+        if volume_type.lower() == existing_volume_type.lower():
+            self.logger.log('No change in data type.')
```
ejarvi: Expected: 'No change in **volume** type'

VMEncryption/main/check_util.py
```diff
@@ -273,14 +273,31 @@ def is_supported_os(self, public_settings, DistroPatcher, encryption_status):
...
+            return
+        if volume_type.lower() == CommonVariables.VolumeTypeAll.lower():
```
ejarvi: If not captured elsewhere, a log message here that the volume type is upgrading from [Data|OS] -> [All] may be helpful for debugging since it is an important transition.

VMEncryption/main/handle.py
```diff
@@ -666,6 +669,71 @@ def enable():
...
+
+def handle_encryption(public_settings, encryption_status, disk_util, bek_util, encryption_operation):
+    extension_parameter = ExtensionParameter(hutil, logger, DistroPatcher, encryption_environment, get_protected_settings(), public_settings)
+    volume_type = public_settings.get(CommonVariables.VolumeTypeKey)
+
+    if extension_parameter.config_file_exists() and extension_parameter.config_changed():
+        logger.log("Config has changed, updating encryption settings")
+        are_devices_encrypted, items_to_encrypt = are_required_devices_encrypted(volume_type, encryption_status, disk_util, bek_util, encryption_operation)
+        if not are_devices_encrypted:
+            logger.log('Required devices not encrypted for volume type {0}. Calling update for already encrypted devices'.format(volume_type))
```
ejarvi: Comment says - 'Calling update for already encrypted devices' but actually isn't it updating the encryption settings to include old devices plus new devices that are not already encrypted?

VMEncryption/main/handle.py
```diff
@@ -666,6 +669,71 @@ def enable():
...
+            logger.log('Settings updated for encrypted devices. Calling enable for new devices.')
+            enable_encryption()
+        else:
+            logger.log('Calling Update Encryption Setting.')
+            update_encryption_settings()
+    else:
+        logger.log("Config did not change or first call, enabling encryption")
+        encryption_marker = EncryptionMarkConfig(logger, encryption_environment)
+        if encryption_marker.config_file_exists():
+            logger.log('Encryption marker exist. Calling Enable')
```
ejarvi: exists

VMEncryption/test/test_check_util.py
```diff
@@ -345,3 +345,79 @@ def test_supported_os(self):
             }, MockDistroPatcher('SuSE', '12.4', ''), {"os" : "NotEncrypted"})
         except Exception as e:
             self.fail("is_unsupported_os threw unexpected exception.\nException message was:\n" + str(e))
+
+    def test_volume_type_enable(self):
```
ejarvi: nit: could be broken into more granular tests, so if one case fails it is easier to pinpoint which one

VMEncryption/main/handle.py
```diff
@@ -686,20 +754,12 @@ def enable_encryption():
...
-        disk_util.consolidate_azure_crypt_mount(existing_passphrase_file)
-        mount_encrypted_disks(disk_util=disk_util,
-                              bek_util=bek_util,
-                              encryption_config=encryption_config,
-                              passphrase_file=existing_passphrase_file)
-    else:
-        if encryption_config.config_file_exists():
-            logger.log(msg="EncryptionConfig is present, but could not get the BEK file.",
+    if existing_passphrase_file is None and encryption_config.config_file_exists():
+        logger.log(msg="EncryptionConfig is present, but could not get the BEK file.",
```
ejarvi: I would prefer "could not get the key file" instead of "BEK file" since BEK = Bitlocker Encryption Key and is more of a Windows acronym.

VMEncryption/main/check_util.py
vimish: Done.

VMEncryption/main/check_util.py
vimish: Done.

VMEncryption/main/handle.py
vimish: Changed the log messages.

VMEncryption/main/handle.py
vimish: Done.

VMEncryption/test/test_check_util.py
vimish: Done.

VMEncryption/main/handle.py
vimish: Done.

vimish:
Addressed review comments from Eric.

PR #843, opened 2019-07-03
Add distro specific actions for RHEL8

taoyama:
#821
This patch add RHEL8 specific actions for LAD though still need additional work to run on RHEL8 correctly.


Diagnostic/DistroSpecific.py
```diff
@@ -260,6 +260,12 @@ def install_lad_mdsd(self):
     def remove_lad_mdsd(self):
         return self.log_run_get_output('rpm -e lad-mdsd')
 
+class Redhat8Actions(RedhatActions):
+    def __init__(self, logger):
+        RedhatActions.__init__(self, logger)
+
+    def install_required_packages(self):
+        return (0, "no additional packages were needed")
```
canfikret: Needs packages such as python itself. However, that's a chicken-egg problem.
We're looking into updating LAD manifest to have a shell script for entry point; which will make sure python is present.

PR #852, opened 2019-07-10
Add centos 7.6 to handled distros for VM encryption.

### LINKED ISSUE #778: AzureDiskEncryptionForLinux - Add support for CentOS 7.6
Similar to PR #753.  

Can we whitelist CentOS 7.6, assuming that it would follow same compatibility as RHEL 7.6?  PR can be submitted if that would accelerate acceptance.

brandondahler:
Resolves #778 -- whitelists CentOS 7.6, assuming that it follows same compatibility as RHEL 7.6.  Was able to use patched version successfully against production servers to enable os/data disk encryption.

Wasn't sure if I should bump the extension version number like #753 did or not.  Let me know if you'd like me to add that.


ejarvi:
Thank you for the contribution, this helped us make the change which went through in a separate PR, so closing this one.

PR #855, opened 2019-07-16
handle low disk space and rpmdb errors

simathih:


OmsAgent/omsagent.py
```diff
@@ -200,7 +205,28 @@ def main():
...
+    except:
+        print('Failed to check disk usage.')
+        return 0
+        
 
+def get_free_space_mb(dirname):
+    """
+    Get the free space in MB in the directory path.
+    """
+    st = os.statvfs(dirname)
```
ers-msft: can we check for failure/-1 here?  shouldn't happen, but in case there is some crazy distro we should handle error and proceed?

PR #860, opened 2019-07-18
Truncat message to 100 before logging/uploading it

abenbachir:


OmsAgent/watcherutil.py
```diff
@@ -223,15 +223,23 @@ def upload_telemetry(self):
                             message = status_data["message"]
 
                             event = self.create_telemetry_event(operation,operation_success,message,"300000")
-                            self._hutil_log("Writing telemetry event: "+event)
                             self.write_waagent_event(event)
```
ers-msft: should we be truncating the message sent to waagent as well?  would like to ensure that is not overwhelmed as well. thanks

abenbachir: Yes, we could, is it ok to do that ?

simathih: yeah, we saw that it grew to 600MB for the customer yesterday

ers-msft: yes for now please also truncate waagent logging, we want to ensure we do not overload that either.

abenbachir: Done, we are now truncating it to 100 for waagent too

PR #861, opened 2019-07-18
ADE dual pass settings parsing 

ejarvi:
- make parsing of settings files more robust in various scenarios

### DIFF
--- a/VMEncryption/main/Utils/HandlerUtil.py
+++ b/VMEncryption/main/Utils/HandlerUtil.py
@@ -90,6 +90,7 @@ class HandlerUtility:
         self.disk_util = None
         self.find_last_nonquery_operation = False
         self.config_archive_folder = '/var/lib/azure_disk_encryption_archive'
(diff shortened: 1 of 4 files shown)


VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -426,15 +501,16 @@ def archive_old_configs(self):
...
+        if src:
+            dest = os.path.join(self.config_archive_folder, 'lnq.settings')
+            if src != dest: 
+                shutil.copy2(src,dest)
 
-                    shutil.copy2(src, dest)
+        # TODO: consider saving [n].settings files in the archive folder 
+        # under versioned subfolders in addition to keeping lnq.settings 
 
     def restore_old_configs(self):
```
vermashi: Do we call this method anywhere anymore?
If we don't we can just remove this line..

ejarvi: removed

PR #872, opened 2019-08-05
LAD: move setup from enable to install step. Enable step can be calle…

canfikret:
…d consecutively resulting in two setups happening in the same minute. Got a CRI with rpm lock issue


canfikret:
Correct. According to extension developer guide, setup needs to be done at install step.

PR #874, opened 2019-08-08
Reject settings if update is called while a daemon is running

vimish:
1. Reject settings if fatal exception occurs while a daemon is running.
2. Preserve waagent log from pivot root environment

### DIFF
--- a/VMEncryption/main/Common.py
+++ b/VMEncryption/main/Common.py
@@ -20,12 +20,15 @@
 class CommonVariables:
     utils_path_name = 'Utils'
     extension_name = 'AzureDiskEncryptionForLinux'
-    extension_version = '1.1.0.43'
+    extension_version = '1.1.0.44'
     extension_type = extension_name
     extension_media_link = 'https://amextpaas.blob.core.windows.net/prod/' + extension_name + '-' + str(extension_version) + '.zip'
     extension_label = 'Azure Disk Encryption For Linux VMSS'
     extension_description = extension_label
     extension_shim_filename = "extension_shim.sh"
+    test_extension_publisher = 'Microsoft.Azure.Security.Edp.'
+    # Change based on test pipeline used
+    test_extension_name = 'AzureDiskEncryptionForLinux'
 
     """
     wire protocol message format
--- a/VMEncryption/main/Utils/HandlerUtil.py
+++ b/VMEncryption/main/Utils/HandlerUtil.py
@@ -72,6 +72,7 @@ from Utils.WAAgentUtil import waagent
 from waagent import LoggerInit
 import logging
 import logging.handlers
+from ProcessLock import ProcessLock
 
 DateTimeFormat = "%Y-%m-%dT%H:%M:%SZ"
 
@@ -91,6 +92,8 @@ class HandlerUtility:
         self.find_last_nonquery_operation = False
         self.config_archive_folder = '/var/lib/azure_disk_encryption_archive'
         self._context = HandlerContext(self._short_name)
+        self.config_folder = '/var/lib/azure_disk_encryption_config'
+        self.status_lock_file_path = '/var/lib/azure_disk_encryption_config/status_lock_file.lck'
 
     def _get_log_prefix(self):
         return '[%s-%s]' % (self._context._name, self._context._version)
@@ -104,6 +107,8 @@ class HandlerUtility:
                 try:
                     if file.endswith('.settings'):
                         cur_seq_no = int(os.path.basename(file).split('.')[0])
+                        if os.path.exists(join(config_folder, str(cur_seq_no) + '.settings.rejected')):
+                            continue
                         if freshest_time == None:
                             freshest_time = os.path.getmtime(join(config_folder, file))
                             seq_no = cur_seq_no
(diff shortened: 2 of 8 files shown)


VMEncryption/main/oscrypto/rhel_72/encryptstates/PatchBootSystemState.py
```diff
@@ -75,9 +75,29 @@ def enter(self):
             self.command_executor.ExecuteInBash('for i in dev proc sys boot; do mount --move /oldroot/$i /$i; done')
 
             extension_full_name = 'Microsoft.Azure.Security.' + CommonVariables.extension_name
+            extension_versioned_name = 'Microsoft.Azure.Security.' + CommonVariables.extension_name + '-' + CommonVariables.extension_version
+            test_extension_full_name = CommonVariables.test_extension_publisher + CommonVariables.test_extension_name
+            test_extension_versioned_name = CommonVariables.test_extension_publisher + CommonVariables.test_extension_name + '-' + CommonVariables.extension_version
```
ejarvi: are these test hooks safe to include in production

VMEncryption/main/Common.py
```diff
@@ -20,12 +20,15 @@
...
-    extension_version = '1.1.0.40'
+    # Change based on test pipeline used
+    test_extension_name = 'AzureDiskEncryptionForLinuxTest1'
+    extension_version = '1.1.0.41'
     extension_type = extension_name
     extension_media_link = 'https://amextpaas.blob.core.windows.net/prod/' + extension_name + '-' + str(extension_version) + '.zip'
     extension_label = 'Azure Disk Encryption For Linux VMSS'
     extension_description = extension_label
     extension_shim_filename = "extension_shim.sh"
+    test_extension_publisher = 'Microsoft.Azure.Security.Edp.'
```
ejarvi: Are test_extension_name and test_extension_publisher leftovers, or intentionally included?

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -403,72 +408,108 @@ def redo_current_status(self):
                               stat[0]["status"]["code"],
                               stat[0]["status"]["formattedMessage"]["message"])
 
+    def reject_settings(self):
```
ejarvi: a small comment on why this method was introduced and in what scenarios it is invoked would be helpful

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -403,72 +408,108 @@ def redo_current_status(self):
...
+                latest_seq = str(self.get_latest_seq())
+            else:
+                self.log("sequence number could not be derived from settings files, using 0.status")
+                latest_seq = "0"
+
+            self._context._status_file = os.path.join(self._context._status_dir, latest_seq + '.status')
+            if message is None:
+                message = ""
+                
+            message = filter(lambda c: c in string.printable, message)
```
ejarvi: Is this valid lambda syntax on Python 2.6?

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -403,72 +408,108 @@ def redo_current_status(self):
                               stat[0]["status"]["code"],
                               stat[0]["status"]["formattedMessage"]["message"])
 
+    def reject_settings(self):
+        try:
+            previous_seq = str(self.get_latest_seq() - 1)
+            previous_status_file = os.path.join(self._context._status_dir, previous_seq + '.status')
+            
+            stat_rept = waagent.GetFileContents(previous_status_file)
```
ejarvi: If a previous status file is not found, it seems this would throw an exception which would be handled but then exit without writing a status file at the current sequence.  If the extension exits without writing a status file corresponding to the current sequence number, I think guest agent will report a QOS impacting failure for that particular operation.

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -403,72 +408,108 @@ def redo_current_status(self):
...
+                stat[0]["status"]["substatus"] = substat
+                
+                if "VMRestartPending" in encryption_status:
+                    stat[0]["status"]["formattedMessage"]["message"] = "OS disk successfully encrypted, please reboot the VM"
+                    
+            stat_rept = json.dumps(stat)
+            # rename all other status files, or the WALA would report the wrong
+            # # status file.
+            # # because the wala choose the status file with the highest sequence
+            # # number to report.
```
ejarvi: I could be wrong on this, but I thought the design of WALA was to choose the sequence number of the operation it has just executed, not necessarily the file with the highest number.   WALA drops an _n_.settings file to the config folder at start of the operation process launch, and then looks for an _n_.status file in the status folder when operation process exits.

VMEncryption/main/handle.py
```diff
@@ -668,6 +660,16 @@ def enable():
                           code=(CommonVariables.configuration_error),
                           message=msg)
 
+    except Exception as e:
+        msg = "Unexpected Error during enable. Error message {0}\n Stacktarce: {1}".format(str(e), traceback.format_exc())
```
ejarvi: Stacktarce -> Stacktrace

VMEncryption/main/handle.py
```diff
@@ -705,6 +707,10 @@ def handle_encryption(public_settings, encryption_status, disk_util, bek_util, e
 
     if extension_parameter.config_file_exists() and extension_parameter.config_changed():
         logger.log("Config has changed, updating encryption settings")
+        # If a daemon is already running reject and exit an update encryption settings request
+        if is_daemon_running():
+            logger.log("An operation already running. Cannot accept an update settings.")
```
ejarvi: an update settings **request**

VMEncryption/main/oscrypto/rhel_72/encryptstates/PatchBootSystemState.py
```diff
@@ -75,9 +75,29 @@ def enter(self):
...
             self.command_executor.Execute('cp -ax' +
                                           ' /var/log/azure/{0}'.format(extension_full_name) +
                                           ' /oldroot/var/log/azure/{0}.Stripdown'.format(extension_full_name))
+            self.command_executor.ExecuteInBash('cp -ax' +
+                                          ' /var/lib/waagent/{0}/config/*.settings.rejected'.format(extension_versioned_name) +
+                                          ' /oldroot/var/lib/waagent/{0}/config'.format(extension_versioned_name))
+            self.command_executor.ExecuteInBash('cp -ax' +
+                                          ' /var/lib/waagent/{0}/status/*.status.rejected'.format(extension_versioned_name) +
+                                          ' /oldroot/var/lib/waagent/{0}/status'.format(extension_versioned_name))
+            self.command_executor.Execute('cp -ax' +
```
ejarvi: remove or comment out test hooks?

VMEncryption/main/Utils/HandlerUtil.py
ejarvi: consider adding a unit test for this function

VMEncryption/main/oscrypto/rhel_72/encryptstates/PatchBootSystemState.py
vimish: Yes Eric. My intention was to save the log files from the pivot root env in the test extension run. As of now these log files are not retained and can sometime make debugging difficult in case of failures. In production extension we will see a failure in logs saying the files cannot be found but it will not fail the encryption operation. I will enable suppress logging option for test extension log copy command so that this operation is not visible in production extension log.

VMEncryption/main/Common.py
vimish: Yes it is intentional. When any of us is deploying a test extension, he should change the value here depending on the test pipeline used. This will ensure extension logs from pivot root env are preserved. Let me know if it sound ok.

VMEncryption/main/Utils/HandlerUtil.py
vimish: Will do.

VMEncryption/main/Utils/HandlerUtil.py
vimish: This syntax is already part of current extension but I will double check it.

VMEncryption/main/Utils/HandlerUtil.py
vimish: I will change the logic here to handle the scenario better.

VMEncryption/main/Utils/HandlerUtil.py
vimish: If an operation for n.settings is executing and an n+1.settings is received in between then the guest agent only reports n+1.status. Hence, any status update from n.settings also needs to go to n+1.status.

VMEncryption/main/handle.py
vimish: Will do.

VMEncryption/main/handle.py
vimish: Will do.

VMEncryption/main/oscrypto/rhel_72/encryptstates/PatchBootSystemState.py
vimish: Reason explained in previous comments. Let me know if it is fine.

VMEncryption/main/Utils/HandlerUtil.py
vimish: I will check the feasibility of adding a unit test here.

VMEncryption/main/Utils/HandlerUtil.py
ejarvi: OK, that makes sense to me now.  Could you add this explanation to the code comment?

VMEncryption/main/Utils/HandlerUtil.py
ejarvi: feel free to close, if this syntax is already in use then it's a moot point

VMEncryption/main/oscrypto/rhel_72/encryptstates/PatchBootSystemState.py
ejarvi: I understand the reasoning now as per previous comments.  It would be preferred to suppress the test-only portion in production to the extent possible.

VMEncryption/main/Utils/HandlerUtil.py
vimish: Done

vimish: I will add unit test for this in another PR.

VMEncryption/main/Utils/HandlerUtil.py
vimish: Done.

VMEncryption/main/Utils/HandlerUtil.py
vimish: Added comment

VMEncryption/main/oscrypto/rhel_72/encryptstates/PatchBootSystemState.py
vimish: Add suppress logging for test env

PR #880, opened 2019-08-13
Fix time2

ChrisCoe:
I want to add a log message if there is no instrumentation key given. I also edited a log message to suggest that mdsd may not be running. For some reason, the commit about the time change is here as well. I was not able to take it out. I'm not sure why that diff is there if Azure/azure-linux-extensions already has that change.

@simathih


ChrisCoe:
This change is not needed as iKey context is now removed from the extension.

PR #885, opened 2019-08-20
ADE DP - Add CentOS 7.7 and RHEL 7.7 

ejarvi:
Add support for CentOS 7.7 and RHEL 7.7 to ADE dual pass extension


vermashi:
Once it has been sanity tested, LGTM
________________________________

PR #887, opened 2019-08-20
ADE DP - fix unsupported file system warning for dual pass extension

ejarvi:
Fix unsupported file system warning message in dual pass extension.  

(Brings DP into parity with existing SP warning message logic)


ejarvi:
sanity tests passed using AzureDiskEncryptionForLinuxTest3

PR #890, opened 2019-08-26
Account for LUKS header size more than 2MB

vimish:
Resize filesystem by 10 MB. Luks format will take only the required space by header.


vimish:
The workitem is on halt. Will create a different PR later.

PR #894, opened 2019-08-29
install/enable and uninstall fixes

simathih:


OmsAgent/omsagent.py
```diff
@@ -514,8 +531,16 @@ def enable():
...
+                try:
+                    open(extension_marker_path, 'w+').close()
+                    hutil_log_info('Created extension marker file ' \
+                               '{0}'.format(extension_marker_path))
+                except IOError as e:
+                    hutil_log_error('Error creating {0} with error: ' \
+                                '{1}'.format(extension_marker_path, e))
+                    # we are having some kind of permissions issue creating the marker file
+                    output = "Couldn't create marker file"
+                    exit_code = 52 # since it is a missing dependency
```
hestolz: **(nit)** I don't know if this is a "missing dependency" scenario; is there a more appropriate exit code?

simathih: it kind of is, we aren't able to create a marker file which means we don't have write access to the filesystem. Write access to filesystem is a requirement

PR #895, opened 2019-09-03
VMBackup extension changes for 1909

vityagi:


VMBackup/main/backuplogger.py
```diff
@@ -46,8 +50,9 @@ def log(self, msg, local=False, level='Info'):
             else:
                 log_msg = "{0}  {1}  {2} \n".format(str(datetime.datetime.now()) , level , msg)
                 self.log_to_con(log_msg)
-            if self.enforced_local_flag_value != None:
-                local = self.enforced_local_flag_value
+            local = True
+            if self.enforced_local_flag_value == False:
+                local = False
             if(local):
```
bpramod: if this variable doesn't serve any purpose, we can directly use enforced_local_flag_value  here.
I assume 'local' variable is not removed as to not break somewhere

VMBackup/main/backuplogger.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, hutil):
 
     def enforce_local_flag(self, enforced_local):
         self.enforced_local_flag_value = enforced_local
+        if (self.enforced_local_flag_value == False):
+            self.msg = self.msg + "================== Logs during Freeze ==============" + "\n"
+        elif (self.enforced_local_flag_value == True):
+            self.commit_to_local()
```
bpramod: add below log or something similar  here
            self.msg = self.msg + "================== End of Logs during Freeze ==============" + "\n"

VMBackup/main/backuplogger.py
```diff
@@ -60,25 +65,33 @@ def log_to_con(self, msg):
                 C.write(message.encode('ascii','ignore'))
         except IOError as e:
             pass
+        except Exception as e:
+            log_msg = "Excetion in log_to_con_py3"
```
bpramod: NIT: Typo Exception

VMBackup/main/backuplogger.py
```diff
@@ -60,25 +65,33 @@ def log_to_con(self, msg):
...
+            time = datetime.datetime.now().strftime(u'%Y/%m/%d %H:%M:%S.%f')
+            log_msg = u"{0}  {1}  {2} \n".format(time , level , msg)
+            log_msg= str(log_msg.encode('ascii', "backslashreplace"), 
+                         encoding="ascii")
             with open(self.con_path, "w") as C :
                 C.write(log_msg)
         except IOError:
             pass
+        except Exception as e:
+            log_msg = "Excetion in log_to_con_py3"
```
bpramod: NIT: Same Typo here

VMBackup/main/backuplogger.py
```diff
@@ -60,25 +65,33 @@ def log_to_con(self, msg):
                 C.write(message.encode('ascii','ignore'))
         except IOError as e:
             pass
+        except Exception as e:
+            log_msg = "Excetion in log_to_con_py3"
+            C.write(log_msg)
```
bpramod: have you tested this flow. "C" seems to be scoped in the with block.
It seems shouldn't work.

VMBackup/main/backuplogger.py
```diff
@@ -60,25 +65,33 @@ def log_to_con(self, msg):
...
+            log_msg = u"{0}  {1}  {2} \n".format(time , level , msg)
+            log_msg= str(log_msg.encode('ascii', "backslashreplace"), 
+                         encoding="ascii")
             with open(self.con_path, "w") as C :
                 C.write(log_msg)
         except IOError:
             pass
+        except Exception as e:
+            log_msg = "Excetion in log_to_con_py3"
+            C.write(log_msg)
```
bpramod: have you tested this flow. "C" seems to be scoped in the with block.
It seems shouldn't work.

VMBackup/main/Utils/DiskUtil.py
```diff
@@ -293,21 +297,30 @@ def get_mount_points(self):
                                 if(fstypeEnd >=0):
                                     fs_type = line[fstypeStart+1:fstypeEnd]
                             # If there is a duplicate, keep only the first instance
-                            if(mount_point not in mount_points):
-                                self.logger.log("mount command mount :" + str(mount_point) + ": and fstype :"+ str(fs_type) + ":", True) 
+                            if ((mount_point not in mount_points) and ((mount_point,deviceName,fs_type) not in mount_points_info)):
```
bpramod: Suggestion : as mount_points, mount_points_info seems same. Checking twice is not needed. i.e. check can be changed as below 
if ((mount_point,deviceName,fs_type) not in mount_points_info):

VMBackup/main/Utils/DiskUtil.py
```diff
@@ -327,9 +340,12 @@ def get_mount_file_systems(self):
                                 fstypeEnd = line.find(" ", fstypeStart+1)
                                 if(fstypeEnd >=0):
                                     fs_type = line[fstypeStart+1:fstypeEnd]
-                            # If there is a duplicate, keep only the first instance
-                    if (file_system,fs_type,mount_point) not in file_systems_info:
+                    # If there is a duplicate, keep only the first instance
+                    if ((mount_point not in mount_points) and ((file_system,fs_type,mount_point) not in file_systems_info)):
```
bpramod: Suggestion :Same here, older check should suffice.

VMBackup/main/fsfreezer.py
```diff
@@ -183,7 +190,7 @@ def thaw_safe(self):
                 else:
                     break
             self.logger.enforce_local_flag(True)
-            self.logger.log("Binary output after process end: ", True)
+            self.logger.log("============== Binary output traces start ================= ", True)
```
bpramod: this logging of binary output is being done when freeze is failed . Can you refactor it into separate method and add these tags there.

VMBackup/main/fsfreezer.py
```diff
@@ -156,15 +157,21 @@ def freeze_safe(self,timeout):
                             self.logger.log(line.rstrip(), True)
                         else:
                             break
-                error_msg="freeze failed for some mount"
-                freeze_result.errors.append(error_msg)
-                self.logger.log(error_msg, True, 'Error')
+                if (sig_handle == 0):
```
bpramod: Suggestion : please add log of "sig_handle" here as well as we are seeing some mis-matches
i.e. inside "if(sig_handle != 1):"

VMBackup/main/fsfreezer.py
```diff
@@ -156,15 +157,21 @@ def freeze_safe(self,timeout):
...
+                    timedout = True
```
bpramod: if freeze fails and ends up in these flows, i dont think we are making a thaw call.
Then where are we setting "enforce_local_flag" to true. Can you please confirm
This might be existing code and we are getting the logs means, things are not even ending in this flow. 
I can only see it getting set in Thaw

VMBackup/main/mounts.py
```diff
@@ -33,46 +33,61 @@ def __init__(self, name, type, fstype, mount_point):
...
+        for mount_point_info in mount_points_info:
+            mountPoint = mount_point_info[0]
+            deviceNameParts = mount_point_info[1].split("/")
+            uniqueName = str(mountPoint) + "_" + str(deviceNameParts[len(deviceNameParts)-1])
+            fsType = mount_point_info[2]
+            if((mountPoint in lsblk_mount_points) and (mountPoint not in added_mount_point_names)):
+                lsblk_mounts_index = 0
+                try:
+                    lsblk_mounts_index = lsblk_unique_names.index(uniqueName)
+                except ValueError as e:
```
bpramod: Just for my understanding, mount output being subset of lsblk output this exception shouldn't ideally arise..right ?

VMBackup/main/Utils/DiskUtil.py
vityagi: if((mount_point,deviceName,fs_type) not in mount_points_info)
Above check alone will not be sufficient for the case where multiple devices were mounted on the same mount-point. Each combination will be different and all will be added to the list, which we have to avoid. Infact, rather (mount_point not in mount_points) alone is sufficient. So, keeping only that.

VMBackup/main/Utils/DiskUtil.py
vityagi: As commented above keeping only (mount_point not in mount_points)

VMBackup/main/backuplogger.py
vityagi: Good suggestion. Done.

VMBackup/main/backuplogger.py
vityagi: Done the change to directly use enforced_local_flag_value.
Yes, local is not removed to not break.

VMBackup/main/backuplogger.py
vityagi: removed this

VMBackup/main/backuplogger.py
vityagi: Good catch. I had not tested this removing this from here.

VMBackup/main/backuplogger.py
vityagi: corrected

VMBackup/main/backuplogger.py
vityagi: removed from here.

VMBackup/main/fsfreezer.py
vityagi: Added

VMBackup/main/fsfreezer.py
vityagi: refactored into a method.

VMBackup/main/fsfreezer.py
vityagi: at the end it is anyways commited to local (end of enable)

VMBackup/main/mounts.py
vityagi: ideally not, it is a safety check for any such condition where uniquenames have some difference.

bpramod:
Looks Good

PR #902, opened 2019-09-12
ADE SP handle case of no settings file

ejarvi:
* handle case of no settings files in default parse config 
* handle case of no settings file in nonquery logic
* add unit test to simulate disable when no settings file present
* update version number


ejarvi:
reviewed with Vishal, test extension validation passed

PR #903, opened 2019-09-13
reducing the poll times

bpramod:


VMBackup/main/PluginHost.py
```diff
@@ -199,8 +201,8 @@ def pre_script(self):
             curr = curr + 1
 
         flag = True
-        for i in range(0,((self.timeoutInSeconds)/5)+2): #waiting 10 more seconds to escape race condition between Host and script timing out
-            time.sleep(5)
+        for i in range(0, int(self.timeoutInSeconds/self.pollTime) + 2): #waiting 10 more seconds to escape race condition between Host and script timing out
```
vityagi: make sure this syntax works in all like pyrthon 2.6
int(self.timeoutInSeconds/self.pollTime)

VMBackup/main/freezesnapshotter.py
```diff
@@ -294,9 +294,10 @@ def takeSnapshotFromFirstGuestThenHost(self):
 
         run_result, run_status, blob_snapshot_info_array, all_failed, all_snapshots_failed, unable_to_sleep, is_inconsistent  = self.takeSnapshotFromGuest()
 
-        time.sleep(60) #sleeping for 60 seconds so that previous binary execution completes
-
         if(all_snapshots_failed):
+            #to make sure binary is thawed
+            self.logger.log('[takeSnapshotFromFirstGuestThenHost] : Thawing again post the guest snapshotting failure')
```
vityagi: check if try-catch is needed here

VMBackup/main/PluginHost.py
bpramod: both are made into int right...what exactly should i test ?

VMBackup/main/ScriptRunner.py
```diff
@@ -102,6 +102,8 @@ def get_config(self):
                 self.postScriptNoOfRetries = configData['postScriptNoOfRetries']
             if 'fsFreezeEnabled' in configDataKeys:
                 self.fsFreeze_on = configData['fsFreezeEnabled']
+            if 'ScriptsExecutionPollTimeSeconds' in configDataKeys and int(configData['ScriptsExecutionPollTimeSeconds']) >= 1 and int(configData['ScriptsExecutionPollTimeSeconds']) <=5:
+                self.pollSleepTime = int(configData['ScriptsExecutionPollTimeSeconds'])
```
vityagi: add try-except ValueError to handle non-int case

bpramod: It is already in bigger try-catch.

VMBackup/main/freezesnapshotter.py
```diff
@@ -294,9 +294,14 @@ def takeSnapshotFromFirstGuestThenHost(self):
...
 
-        time.sleep(60) #sleeping for 60 seconds so that previous binary execution completes
-
         if(all_snapshots_failed):
+            try:
+                #to make sure binary is thawed
+                self.logger.log('[takeSnapshotFromFirstGuestThenHost] : Thawing again post the guest snapshotting failure')
+                self.freezer.thaw_safe()
+            except Exception as e:
+                self.logger.log('[takeSnapshotFromFirstGuestThenHost] : Exception in Thaw - ' + str(e))
```
vityagi: Trace in the way we trace exceptions with stack trace.
+ ': %s, stack trace: %s' % (str(err), traceback.format_exc())

bpramod: Good Catch!

PR #905, opened 2019-09-17
ADE DP improve error handling when no settings file present

ejarvi:
- equivalent to change recently merged to single pass extension
- adds error handling in the case of no settings file present when parsing context
- adds a unit test around disable to cover this scenario since a failure in disable can block update


ejarvi:
changes deployed to AzureDiskEncryptionForLinuxTest3, running tests

PR #920, opened 2019-10-17
Fix typo

johanburati:
Fixed typo,  *Permisisons* should be *Permissions*, also removed double space.


johanburati:
This fix issue  https://github.com/Azure/azure-linux-extensions/issues/921

PR #924, opened 2019-10-18
LAD: Reinstall OMI if it is failing to start as recovery action

canfikret:
We got a few CRIs where OMI service does not start due to a possible corruption. LAD detects that OMI is not available; it attempts to restart the service. If still a problem, the component is re-installed.

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -311,7 +311,7 @@ def main(command):
                 dependencies_err, dependencies_msg = setup_dependencies_and_mdsd(configurator)
                 if dependencies_err != 0:
                     g_lad_log_helper.report_mdsd_dependency_setup_failure(waagent_ext_event_type, dependencies_msg)
-                    hutil.do_status_report(g_ext_op_type, "error", '-1', "Enablecd  failed")
+                    hutil.do_status_report(g_ext_op_type, "error", '-1', "Enabled failed")
                     return
 
             if g_dist_config.use_systemd():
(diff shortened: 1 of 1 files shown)


Diagnostic/diagnostic.py
```diff
@@ -311,7 +311,7 @@ def main(command):
                 dependencies_err, dependencies_msg = setup_dependencies_and_mdsd(configurator)
                 if dependencies_err != 0:
                     g_lad_log_helper.report_mdsd_dependency_setup_failure(waagent_ext_event_type, dependencies_msg)
-                    hutil.do_status_report(g_ext_op_type, "error", '-1', "Enablecd  failed")
+                    hutil.do_status_report(g_ext_op_type, "error", '-1', "Enabled failed")
```
AaronRM: Enabled=>Enable?

Diagnostic/diagnostic.py
```diff
@@ -644,7 +644,21 @@ def restart_omi_if_crashed(omi_installed, mdsd):
                         "Restarting OMI and sending SIGHUP to mdsd after 5 seconds.")
```
AaronRM: Change message to say 10 seconds.

Diagnostic/diagnostic.py
canfikret: yep :-)

Diagnostic/diagnostic.py
canfikret: right.

PR #929, opened 2019-10-22
Migrate crypt mount to crypttab[DualPass]

vimish:


VMEncryption/main/Common.py
```diff
@@ -217,10 +218,60 @@ def __init__(self):
...
+
+    def __eq__(self, other):
+        """
+        Override method for "==" operation, useful for making CryptItem comparison a little logically consistent
+        For example a luks_slot value of "-1" and "None" are logically equivalent, so this method, treats them the same
+        This is done by "consolidating" both values to "None".
+        """
+        if not isinstance(other, CryptItem):
+            return NotImplemented
+
```
jax-microsoft: Should we check:  if (self is other) : return true

PR #938, opened 2019-11-14
Extension Packaging for Azure Monitor Linux Agent

nidhanda:
This code is built from the OMSLinuxAgent extension code base, so a lot of Waagent and Util functions have been reused. 

Creating this PR right now because it is a big change and thus might take more time for review and since i will be OOF starting next Tuesday.

The current version of the extension that is deployed to EDP Canary is 1.3 . Due to the PIR subscription quota limit, I haven't been able to test the update feature yet since no new agent extension can be deployed. Will create a new PR if a fix is needed for the update feature. Waiting on 1.4 to be deployed.


abenbachir:
Can we do the versioning with the patch number too like:
<Major>.<Minor>.<Patch>

AzureMonitorAgent/manifest.xml
```diff
@@ -0,0 +1,18 @@
...
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
+  <ProviderNameSpace>Microsoft.Azure.Monitor</ProviderNameSpace>
+  <Type>AzureMonitorLinuxAgent</Type>
+  <Version>1.4</Version>
+  <Label>Microsoft Azure Monitoring Agent for Linux</Label>
+  <HostingResources>VmRole</HostingResources>
+  <MediaLink></MediaLink>
+  <Description>Microsoft Azure Monitoring Agent for Linux</Description>
+  <IsInternalExtension>true</IsInternalExtension>
+  <Eula>https://msazure.visualstudio.com/DefaultCollection/One/_git/Compute-Runtime-Tux?path=%2FLICENSE</Eula>
```
simathih: is this accessible outside of corp without MSFT creds?

AzureMonitorAgent/agent.py
```diff
@@ -0,0 +1,864 @@
...
+    Ensure that this VM distro and version are supported.
+    Install the Azure Monitor Linux Agent package, using retries.
+    Note: install operation times out from WAAgent at 15 minutes, so do not
+    wait longer.
+    """
+    find_package_manager("Install")
+    exit_if_vm_not_supported('Install')
+
+    # public_settings, protected_settings = get_settings()
+    # if public_settings is None:
```
simathih: pls remove commented code

AzureMonitorAgent/agent.py
```diff
@@ -0,0 +1,864 @@
...
+
+def find_package_manager(operation):
+    """
+    Checks if the dist is debian based or centos based and assigns the package manager accordingly
+    """
+    global PackageManager
+    global BundleFileName
+    dist, ver = find_vm_distro(operation)
+
+    dpkg_set = {"oracle", "debian", "ubuntu", "suse"}
```
simathih: shouldn't we also check if the packagemanger is available in that distro?

AzureMonitorAgent/agent.py
```diff
@@ -0,0 +1,864 @@
...
+    rpm_set = {"redhat", "centos", "red hat"}
+    for dpkg_dist in dpkg_set:
+        if dist.lower().startswith(dpkg_dist):
```
simathih: is startswith enough? also, break the for loop once the match is found

AzureMonitorAgent/agent.py
```diff
@@ -0,0 +1,864 @@
...
+            PackageManager = "dpkg"
+            BundleFileName = BundleFileNameDeb
+    for rpm_dist in rpm_set:
+        if dist.lower().startswith(rpm_dist):
+            PackageManager = "rpm"
```
simathih: as above, break as soon as a match is found

AzureMonitorAgent/agent.py
```diff
@@ -0,0 +1,864 @@
...
+    exit_if_vm_not_supported('Install')
+
+    # public_settings, protected_settings = get_settings()
+    # if public_settings is None:
+    #     raise ParameterMissingException('Public configuration must be ' \
+    #                                     'provided')
+    package_directory = os.path.join(os.getcwd(), PackagesDirectory)
+    bundle_path = os.path.join(package_directory, BundleFileName)
+    os.chmod(bundle_path, 100)
+    print (PackageManager, "and", BundleFileName)
```
simathih: " and " with spaces

simathih: unless print automatically adds them

AzureMonitorAgent/agent.py
```diff
@@ -0,0 +1,864 @@
...
+    hutil_log_info('Running command "{0}"'.format(OneAgentInstallCommand))
+
+    # Retry, since install can fail due to concurrent package operations
+    exit_code, output = run_command_with_retries_output(OneAgentInstallCommand, retries = 15,
+                                         retry_check = retry_if_dpkg_locked,
+                                         final_check = final_check_if_dpkg_locked)
+
+    try:
+        if os.path.isfile("/etc/default/mdsd"):
+            with open("/etc/default/mdsd", "a") as f:
```
simathih: can you pls verify that it adds this as expected on a test depl?

AzureMonitorAgent/agent.py
nidhanda: I was keeping them if in case we decide to send some settings data through the waagent itself instead of MCS. So kept it  in case we need to modify.

AzureMonitorAgent/agent.py
nidhanda: Yes I can see additions in both rhel and ubuntu test distros

AzureMonitorAgent/agent.py
nidhanda: Wouldn't they by default always have a package manager?

AzureMonitorAgent/agent.py
nidhanda: Yes startswith is enough to test the azure distros, This is the way we currently use in omsagent installation as well. So they are well tested.

AzureMonitorAgent/agent.py
nidhanda: Got it. Will add the break.

AzureMonitorAgent/manifest.xml
nidhanda: I don't assume so, but where else can we pull the License from? Should I add that licence into this repo?

simathih: ok, can you point it to this doc for now ? https://docs.microsoft.com/en-us/azure/azure-monitor/learn/quick-collect-linux-computer . I am not sure if we should release the other one rt now

AzureMonitorAgent/agent.py
simathih: if you check omsagent shell bundle it actually checks for the existence of the binary...we can leave it this way for now

AzureMonitorAgent/manifest.xml
nidhanda: Will do

Issue #891, opened 2019-08-28
LAD 3.0.123 extension installation on Oracle 7 image from CIS fails


nayakshweta:
LAD 3.0.123 extension installation on Oracle 7 image from CIS fails with the error :

2019/08/20 08:50:12 ERROR:CalledProcessError.  Command result was INFO:  Configuring OMS agent service LAD ...

2019/08/20 08:50:12 [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.123] Return 1:INFO:  Configuring OMS agent service LAD ...

2019/08/20 08:50:12

2019/08/20 08:50:12 ERROR:[Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.123] Failed to perform extension operation Install with error:global name 'waagent_ext_event_type' is not defined, Traceback (most recent call last):

2019/08/20 08:50:12 ERROR:  File "/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.123/diagnostic.py", line 300, in main

2019/08/20 08:50:12 ERROR:    g_lad_log_helper.report_mdsd_dependency_setup_failure(waagent_ext_event_type, dependencies_msg)

2019/08/20 08:50:12 ERROR:NameError: global name 'waagent_ext_event_type' is not defined

2019/08/20 08:50:12 ERROR:

2019/08/20 08:50:12 [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.123] Install,error,0,Extension operation Install failed:global name 'waagent_ext_event_type' is not defined

 

 

2019/08/20 08:50:13 ERROR:CalledProcessError.  Command result was /bin/sh: /usr/local/lad/bin/mdsd: No such file or directory

2019/08/20 08:50:13 [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.123] Return 127:/bin/sh: /usr/local/lad/bin/mdsd: No such file or directory

2019/08/20 08:50:13

2019/08/20 08:50:13 [Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.123] Problem(s) detected in generated mdsd configuration. Can't enable, although this install/enable operation is reported as successful so the VM can complete successful startup. Linux Diagnostic Extension will exit. Config validation message: /bin/sh: /usr/local/lad/bin/mdsd: No such file or directory

2019/08/20 08:50:13

[...]

PR #939, opened 2019-11-14
properly initialize var

hestolz:
#891 and ICM 158387067


shawsers:
any update on this issue?

hestolz:
@axelgMS not the case. first of all this commit is not deployed yet, secondly it is actually a fix for that error you are seeing ("global name is not defined"). That being said, in reality this bug path is only encountered when LAD was going to fail to install anyways. So you need to check elsewhere (probably earlier) in the log file. This bug is hit when LAD was failing to install anyways and was going to log an error, and that logging statement has the bug.

PR #941, opened 2019-11-19
Using Purge to uninstall with dpkg and adding checks for MCS vars

nidhanda:
By default, the -r/--remove flag in dpkg doesn't remove the config files (for ex. /etc/default/mdsd ) And we want the file to be replaced with the new installation. 
Thus adding the -P (--purge) flag to completely remove the configs related to mdsd so that the new version can create a fresh copy

Also adding checks before adding the MCS export vars statements (In case of Debian systems, if --install operation is ran multiple times, it will keep appending to the same file, Doesn't happen in RedHat systems as rpm prevents from installing the same package)

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -205,11 +205,15 @@ def install():
 
     try:
         if os.path.isfile("/etc/default/mdsd"):
-            with open("/etc/default/mdsd", "a") as f:
-                f.write("\n")
-                f.write("export ENABLE_MCS=true\n")
-                f.write("export MCS_ENDPOINT=mcs.azure.com\n")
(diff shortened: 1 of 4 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -205,11 +205,15 @@ def install():
 
     try:
         if os.path.isfile("/etc/default/mdsd"):
-            with open("/etc/default/mdsd", "a") as f:
-                f.write("\n")
-                f.write("export ENABLE_MCS=true\n")
-                f.write("export MCS_ENDPOINT=mcs.azure.com\n")
-                f.write("export AZURE_ENDPOINT=https://management.azure.com/\n")
+            with open("/etc/default/mdsd", "r+") as f:
```
simathih: a+ to write at end of the file

Issue #942, opened 2019-11-21
os.chmod: SyntaxError: invalid token


johanburati:
## ERROR

```
[ExtensionOperationError] Non-zero exit code: 1, /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OMSAgentForLinux-1.12.15/omsagent.py -install
[stdout]
2019/11/20 23:19:07.321999 INFO ExtHandler Added PID 103624 to cgroup /sys/fs/cgroup/cpu/walinuxagent.extensions/Microsoft.EnterpriseCloud.Monitoring.OMSAgentForLinux_1.12.15
2019/11/20 23:19:07.324020 INFO ExtHandler Added PID 103624 to cgroup /sys/fs/cgroup/memory/walinuxagent.extensions/Microsoft.EnterpriseCloud.Monitoring.OMSAgentForLinux_1.12.15


[stderr]
  File "/var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OMSAgentForLinux-1.12.15/omsagent.py", line 501
    os.chmod(etc_final_path, 0750)
                                ^
SyntaxError: invalid token
```

johanburati:
Both *python2* and *python3* are installed on the system.

I have submitted a [fix ](https://github.com/Azure/azure-linux-extensions/pull/943) for this.

simathih:
Fix is part of Feb release

PR #943, opened 2019-11-21
Fix for issue #942

johanburati:
I have tested `os.chmod("/tmp/test", 0o750)` with *python2* and *python3* and this syntax works with both version, therefore removing the `if..then` that distiguish between the two versions.

This fixes issue https://github.com/Azure/azure-linux-extensions/issues/942

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -495,24 +495,15 @@ def enable():
             os.chown(etc_final_path, uid, gid)
 
             # octal numbers are represented differently in python 3
-            if sys.version_info >= (3,):
-                os.chmod(etc_final_path, 0o750)
-            else:    
-                os.chmod(etc_final_path, 0750)
+            os.chmod(etc_final_path, 0o750)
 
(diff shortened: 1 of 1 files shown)


simathih:
I am not sure if this is safe. There are definitely some older versions of python wher 0o didn't work which is why we had to go with this check. I can pull up those incidents up. Check the history of the code there. We have to handle all scenarios; are you sure it works in Python 2.6 and 2.7? IIRC, the behavior changed in some version of 2. So, all client before that will fail.

johanburati:
> This won't work for all versions in Python 2.

Python need to be  older than 2.6,
I checked on older supported distro and could not find any with a version of Python older than 2.6:
```
CentOS 6.6 Final
Python 2.6.6
Before 'os.chmod(file, 0o755)': 644
After  'os.chmod(file, 0o755)': 755
..
Ubuntu 12.04 precise
Python 2.7.3
Before 'os.chmod(file, 0o755)': 644
After  'os.chmod(file, 0o755)': 755
```

hestolz:
Fix made #986, ETA for release is February.

Issue #896, opened 2019-09-05
Add NS-BSD compatibility for VMBackup Extension


remip2:
I'm trying to add compatibility with NS-BSD OS for the VMBackup extension. The current extension  (9151) fails during the installation and the backup fails as well.

I prepared the patch in a fork: https://github.com/remip2/azure-linux-extensions/tree/nsbsd-os-support

Before submitting a pull-request I want to test the code and verify the extension install correctly and allows backups. How can I manually install my patched extension ?

soumyapattnaik:
i have added the code review comments.

vrdmr:
#966 Merged. Closing the issue.

PR #966, opened 2019-12-13
NSBSD os support for VMBackup extension

remip2:
This patch adds support for NSBSD OS for the Backup extension.

I installed manually the patched extension and backups now end with success.

Issue #896

### DIFF
--- a/VMBackup/main/Utils/HandlerUtil.py
+++ b/VMBackup/main/Utils/HandlerUtil.py
@@ -204,7 +204,11 @@ class HandlerUtility:
                 f.close()
                 waagent.SetFileContents(f.name,config['runtimeSettings'][0]['handlerSettings']['protectedSettings'])
                 cleartxt = None
-                cleartxt = waagent.RunGetOutput(self.patching.openssl_path + " base64 -d -A -in " + f.name + " | " + self.patching.openssl_path + " smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
+                if 'NS-BSD' in platform.system():
+                    # base64 tool is not available with NSBSD, use openssl
+                    cleartxt = waagent.RunGetOutput(self.patching.openssl_path + " base64 -d -A -in " + f.name + " | " + self.patching.openssl_path + " smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
+                else:
+                    cleartxt = waagent.RunGetOutput(self.patching.base64_path + " -d " + f.name + " | " + self.patching.openssl_path + " smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
                 jctxt = {}
                 try:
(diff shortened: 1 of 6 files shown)


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -204,7 +204,7 @@ def _parse_config(self, ctxt):
                 f.close()
                 waagent.SetFileContents(f.name,config['runtimeSettings'][0]['handlerSettings']['protectedSettings'])
                 cleartxt = None
-                cleartxt = waagent.RunGetOutput(self.patching.base64_path + " -d " + f.name + " | " + self.patching.openssl_path + " smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
+                cleartxt = waagent.RunGetOutput(self.patching.openssl_path + " base64 -d -A -in " + f.name + " | " + self.patching.openssl_path + " smime  -inform DER -decrypt -recip " + cert + "  -inkey " + pkey)[1]
```
soumyapattnaik: please have this code specific to NS-BSD. We dont want to break it for other supported OS

VMBackup/main/patch/__init__.py
```diff
@@ -19,6 +19,7 @@
 import os
 import re
 import platform
+import traceback
```
soumyapattnaik: this import is unwanted

VMBackup/main/handle.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
soumyapattnaik: please revert this to bash shell. For freeBSD we are suggesting customers to install bash. We have plans to move this to bourne shell in few days. It has to be tested with all the supported distros.

VMBackup/main/patch/__init__.py
remip2: the lib is used line 64, it's something i came across during debugging

VMBackup/main/handle.sh
remip2: Unfortunately, NSBSD is not FreeBSD, Bash is not available and I can't add it. If you already plan to move to sh, I can wait and merge. At first look there is no bash specific syntax in the script.

VMBackup/main/Utils/HandlerUtil.py
remip2: ok, a fix has been pushed

VMBackup/main/handle.sh
soumyapattnaik: got it. I will merge the changes with bourne shell, as we plan to support bourne shell in next release

remip2:
@soumyapattnaik thanks for the merge, do you have an idea when this version of vmbackup will be deployed live on Azure ? I tested today and this is still Microsoft.Azure.RecoveryServices__VMSnapshotLinux__1.0.9157.0 without the patch which is downloaded.

PR #975, opened 2020-01-10
ADE SP fix Ubuntu 14.04 OS disk encryption logic

ejarvi:
Update the SplitRootPartitionState logic to more reliably unmount the old root device prior to entering subsequent cryptsetup commands that will fail if the device is still in use.


ejarvi:
@vermashi - I do not see your comments in the PR, but the current set of changes incorporate the revisions suggested in our earlier review, tests passing against AzureDiskEncryptionForLinuxTest2, expediting merge.

PR #980, opened 2020-01-15
ADE DP fix U14 unmount oldroot sequence in SplitRootPartitionState

ejarvi:
Fixes an issue in SplitRootPartitionState that was not properly unmounting oldroot prior to advancing to next state, breaking cryptsetup commands for OS disk encryption when device was still in use.


ejarvi:
@vermashi  @vimish - this is the dual pass equivalent of the fix already merged to single pass

ejarvi:
@vermashi @vimish - please review

PR #981, opened 2020-01-15
Fixed MCS env var update in /etc/default/mdsd 

nidhanda:
Added checks for existing values of the variables and replace them with new values instead of appending at the end.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -35,6 +35,7 @@ import urllib2
 import shutil
 import crypt
 import xml.dom.minidom
+import re
 from distutils.version import LooseVersion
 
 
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -203,25 +204,41 @@ def install():
...
-                    f.write("export AZURE_ENDPOINT=https://management.azure.com/\n")
-                if "ADD_REGION_TO_MCS_ENDPOINT" not in data:
-                    f.write("export ADD_REGION_TO_MCS_ENDPOINT=false\n")
-                if "OMS_TLD" not in data:
-                    f.write("export OMS_TLD=int2.microsoftatlanta-int.com\n")
-                if "customResourceId" not in data:
-                    f.write("export customResourceId=/subscriptions/42e7aed6-f510-46a2-8597-a5fe2e15478b/resourcegroups/amcs-test/providers/Microsoft.OperationalInsights/workspaces/amcs-pretend-linuxVM\n")
+                for var in default_configs.keys():
+                    pattern = "(#*)export " + str(var) + "=.*$"
+                    r_obj = re.compile(pattern, flags=re.MULTILINE)
```
simathih: how did you verify this?

nidhanda: I copied over that file to a vm in canary that had ama installed and then used that agent.py to uninstall/install/update. While changing the mdsd file and checking if it was updated correctly

PR #984, opened 2020-01-17
ADE SP - add python-six for CentOS 6.8 compatibility

ejarvi:
adds python-six for CentOS 6.8 compatibility, required by transitions


ejarvi:
@vimish @vermashi - please review

PR #986, opened 2020-01-18
fix octal number and ascii encoding install issues

simathih:
- also version changes for 1.12.25


hestolz:
@simathih these chmod changes need to be mirrored in `omsagent_ff.py` and `omsagent_mc.py`

PR #989, opened 2020-01-22
add first party support to AMA

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -204,14 +208,75 @@ def install():
...
-        "customResourceId" : "/subscriptions/42e7aed6-f510-46a2-8597-a5fe2e15478b/resourcegroups/amcs-test/providers/Microsoft.OperationalInsights/workspaces/amcs-pretend-linuxVM"
+        "ADD_REGION_TO_MCS_ENDPOINT" : "true",
+        #"OMS_TLD" : "int2.microsoftatlanta-int.com",
+        #"customResourceId" : "/subscriptions/42e7aed6-f510-46a2-8597-a5fe2e15478b/resourcegroups/amcs-test/providers/Microsoft.OperationalInsights/workspaces/amcs-pretend-linuxVM",
+        #"ENABLE_MCS" : "true",
     }
+
+    # decide the mode
+    if protected_settings is None:
+        default_configs["ENABLE_MCS"] = "true"
```
canfikret: false?

AzureMonitorAgent/agent.py
```diff
@@ -204,14 +208,75 @@ def install():
...
+            default_configs["MONITORING_CONFIG_VERSION"] = MONITORING_CONFIG_VERSION
+            default_configs["MONITORING_GCS_CERT_CERTFILE"] = "/etc/mdsd.d/gcscert.pem"
+            default_configs["MONITORING_GCS_CERT_KEYFILE"] = "/etc/mdsd.d/gcskey.pem"
+
+            # write the certificate and key to disk
+            fh = open("/etc/mdsd.d/gcscert.pem", "wb")
+            fh.write(MONITORING_GCS_CERT_CERTFILE)
+            fh.close()
+
+            fh = open("/etc/mdsd.d/gcskey.pem", "wb")
```
canfikret: set mode and owner.

AzureMonitorAgent/agent.py
```diff
@@ -204,14 +208,75 @@ def install():
...
+        MONITORING_GCS_REGION = ""
+        if protected_settings.has_key("monitoringGCSRegion"):
+            MONITORING_GCS_REGION = protected_settings.get("monitoringGCSRegion")
+
+        MONITORING_CONFIG_VERSION = ""
+        if protected_settings.has_key("configVersion"):
+            MONITORING_CONFIG_VERSION = protected_settings.get("configVersion")
+
+        if MONITORING_GCS_CERT_CERTFILE is None or MONITORING_GCS_CERT_KEYFILE is None or MONITORING_GCS_ENVIRONMENT is "" or MONITORING_GCS_NAMESPACE is "" or MONITORING_GCS_ACCOUNT is "" or MONITORING_GCS_REGION is "" or MONITORING_CONFIG_VERSION is "":
+            default_configs["ENABLE_MCS"] = "true"
```
canfikret: Why not having an explicit switch to enable MCS? A misconifg in GCS does not imply MCS is enabled.

AzureMonitorAgent/agent.py
simathih: no, if no protected settings are passed, the default mode is 3rd party mode with AMCS

AzureMonitorAgent/agent.py
simathih: MCS is enabled by default, but I now removed the code to enable it when GCS is misconfigured

Issue #992, opened 2020-01-23
[OmsAgentForLinux] Unsupported operating system: RHEL/CentOS  8


johanburati:
**RHEL/Centos 8** is still not supported by **OmsAgentForLinux** (hard coded)

### Version
```
# cat /etc/redhat-release
Red Hat Enterprise Linux release 8.0 (Ootpa)
```
### Error
```
[ExtensionOperationError] Non-zero exit code: 51, /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.12.17/omsagent.py -install
[stdout]
2020/01/23 05:53:24 [Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.12.17] Install,failed,51,Unsupported operating system: Red Hat Enterprise Linux 8.0
```

### Version
```
# cat /etc/redhat-release
CentOS Linux release 8.0.1905 (Core)
```
### Error
```
[ExtensionOperationError] Non-zero exit code: 51, /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.12.17/omsagent.py -install
[stdout]
2020/01/23 05:42:35 [Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.12.17] Install,failed,51,Unsupported operating system: CentOS Linux 8.0.1905
```

johanburati:
Same issue with *Oracle Linux 8*

## Oracle Linux 8
### Version
```
# cat /etc/redhat-release
Red Hat Enterprise Linux release 8.0 (Ootpa)
```
## Error
```
[ExtensionOperationError] Non-zero exit code: 51, /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.12.17/omsagent.py -install
[stdout]
2020/01/23 06:14:24 [Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.12.17] Install,failed,51,Unsupported operating system: Oracle Linux Server 8.0
```

johanburati:
I have submitted  PR https://github.com/Azure/azure-linux-extensions/pull/993 to allow the **OmsAgentForLinux** extension to install on **RHEL/CentOS/Oracle Linux 8**

hestolz:
By design. These distributions are not yet supported by the agent. We will verify them in the future; RHEL8 is targeted for this February. Others will follow TBD.

[...]

PR #993, opened 2020-01-23
Add Redhat/Centos 8 to OmsAgent

johanburati:
Fix for https://github.com/Azure/azure-linux-extensions/issues/992


hestolz:
@johanburati Thanks for your contribution—however we hard code this for a reason. We must first test the agent on new distributions before we can allow it to pass this check pre-install. Verification work/ fixes for RHEL 8 are in the pipeline, slated for February. Following that we will verify CentOS and Oracle, ETA TBD.

DimitrovMiro:
@hestolz do we have an ETA for supporting OMS with RHEL 8.0 . We are march now

PR #996, opened 2020-01-28
Stop shipping fluent-plugin-mdsd.gem with LAD (already packaged in OMS)

abenbachir:


Diagnostic/Utils/omsagent_util.py
```diff
@@ -58,12 +56,6 @@ def setup_omsagent_for_lad(run_command):
         if cmd_exit_code != 0:
             return 2, 'setup_omsagent_for_lad(): LAD workspace onboarding failed. Output: {0}'.format(cmd_output)
 
-    # 3. Install fluentd out_mdsd plugin (uninstall existing ones first)
```
AaronRM: Please remove the other comments in this file that discuss installing the out_mdsd plugin.

abenbachir: done

PR #999, opened 2020-01-28
Fixed default mdsd variable addition

nidhanda:


AzureMonitorAgent/agent.py
```diff
@@ -290,22 +290,28 @@ def install():
...
+                        export_dict[var] = "export " + var + "=" + default_configs[var] + "\n"
+                        if var in line:
+                            line = export_dict[var]
+                            vars_set.add(var)
+                            break
+                    new_data += line
+            
+            for var in default_configs.keys():
+                if var not in vars_set:
+                    new_data += export_dict[var]
```
simathih: this will be empty for new tokens not already in file in some edge cases due to break statement above

simathih: I meant the export_dict[var]

nidhanda: export_dict[var] will have an entry for every single var that is in default_configs. Any new tokens that are to be added will already be part of default_configs right? I am not sure if i understand the edge case where this will be empty. Could you give an example?

PR #1005, opened 2020-02-05
ADE: Configuring CDPx build in github-private.visualstudio.com/azure

ivanp-ms:
- Configuring ADE-SinglePass-Dev branch for CDPx builds in github-private.visualstudio.com.

- CDPx requires .pipelines folder under the source root folder with YAML files describing [mulit-phase](https://onebranch.visualstudio.com/Pipeline/_wiki/wikis/Pipeline.wiki/291/Multi-Phase-and-Multi-Configuration-Builds) build steps. 

- Multiple YAML files are all specific to builds tagged with "ade". This allows having other projects their specific CDPx pipelines. YAML files are inclusive so other projects should not be creating YAML files for all tags since that could affect ADE specific YAML files.

- CDPx drops the ZIP file and manifest.xml file in VSTS. Currently there is not support for integration with CloudVault but I'm following up with OneBranch team on the plan for that.

- The change moves the VM Extension version out of the code over to version.txt file. CDPx supports [automatic build versioning](https://onebranch.visualstudio.com/Pipeline/_wiki/wikis/Pipeline.wiki/325/Versioning) without making any changes in the checked in code. 

- version.txt file gets updated at build time. This allows using Jenkins pipeline for the time being until we fully switch to CDPx.


VMEncryption/main/BuildVersion.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
+#
+# Azure Disk Encryption For Linux extension
+#
+# Copyright 2016 Microsoft Corporation
```
ejarvi: 2020

VMEncryption/main/oscrypto/91ade/__init__.py
```diff
@@ -0,0 +1,14 @@
+#
+# Copyright 2015 Microsoft Corporation
```
ejarvi: 2020

PR #1010, opened 2020-02-11
[LAD] Add flag to disable storage account requirement

### LINKED ISSUE #944: Install LAD without storage account sink
Is there a way to install the Linux diagnostics extension without providing storage account credentials? We are only interested in the Event Hub sink.

Without the storage account credentials, the extension fails to start.

snehanagendra:
This flag is for syslog table only. It is useful when shipping syslog via eventhubpublisher, and don't need/want to duplication in Azure storage table

Fixes #944 

Signed-off-by: Sneha Nagendra <snnagen@microsoft.com>

### DIFF
--- a/Diagnostic/lad_config_all.py
+++ b/Diagnostic/lad_config_all.py
@@ -432,7 +432,7 @@ class LadConfigAll:
             fileLogs_setting = self._ext_settings.get_fileLogs_setting()
             lad_logging_config_helper = LadLoggingConfig(syslogEvents_setting, fileLogs_setting, self._sink_configs,
                                                          self._pkey_path, self._cert_path, self._encrypt_secret)
-            mdsd_syslog_config = lad_logging_config_helper.get_mdsd_syslog_config()
+            mdsd_syslog_config = lad_logging_config_helper.get_mdsd_syslog_config(self._ext_settings.read_protected_config('disableStorageAccount') == True)
             mdsd_filelog_config = lad_logging_config_helper.get_mdsd_filelog_config()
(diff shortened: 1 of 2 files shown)


Diagnostic/lad_config_all.py
```diff
@@ -432,7 +432,7 @@ def generate_all_configs(self):
             fileLogs_setting = self._ext_settings.get_fileLogs_setting()
             lad_logging_config_helper = LadLoggingConfig(syslogEvents_setting, fileLogs_setting, self._sink_configs,
                                                          self._pkey_path, self._cert_path, self._encrypt_secret)
-            mdsd_syslog_config = lad_logging_config_helper.get_mdsd_syslog_config()
+            mdsd_syslog_config = lad_logging_config_helper.get_mdsd_syslog_config(self._ext_settings.read_protected_config('disableStorageAccount') == True)
```
canfikret: where is disableStorageAccount to be defined? In public json?

snehanagendra: under protected settings. like so - 

{
	"protectedSettings": {
		"storageAccountName": "__PlaceholderStorageAccountName__",
		"storageAccountSasToken": "__PlaceholderSASToken__",
		**"disableStorageAccount": true,**
		"sinksConfig": {
			"sink": [
				{
					"name": "IdleEventHubSink",
					"type": "EventHub",
					"sasURL": "<...>"
				}
			]
		}
	}
}

l3ender:
How is this flag used from configuration? We are interested in using LAD 4.0 only for Event Hub.

Thank you!

PR #1012, opened 2020-02-11
OMS Python 2/3

hestolz:
We need to support both Python 2.6+ and 3 (2.6+ is required by Guest Agent; I have additionally checked and all our supported distros ship with 2.6 minimum). I made the changes following [this guide](https://sebastianraschka.com/Articles/2014_python_2_3_key_diff.html#integer-division) and using the [futurize tool](https://python-future.org/automatic_conversion.html). 

Besides running the tool I did manual checks over the code, informed by the linked guide
- string representations: ensured no concats or compares between `bytes` and `str` type, trickiest area was the workspace key check using base64 encode/decode, where for py3 we have to decode the returned `bytes` before comparison (base64 docs [2](https://docs.python.org/2/library/base64.html), [3](https://docs.python.org/3/library/base64.html))
- integer division (chose to use `//` as its behavior is consistent between 2 and 3 and will round down for safety)
- `except ... as`, `raise` syntax, `print()` syntax
- unorderable types

Following that I created a shim following the design of the [Custom Script](https://github.com/Azure/azure-linux-extensions/blob/master/CustomScript/shim.sh) and VMAccess ([1](https://github.com/Azure/azure-linux-extensions/pull/649), [2](https://github.com/Azure/azure-linux-extensions/pull/636)) extensions. The shim will explicitly call the versioned `python2` or `python3` commands, reason being that RHEL family of distributions 8+ will:
 1. not include python by default
 2. not alias `python` to the version installed by the user (it can be done but is explicitly discouraged)

Additionally, there's no reason for us to maintain nearly-identical scripts for FF/MC. We can have a single file and comment/uncomment the correct onboarding command (will be documented in release onenote).


OmsAgent/watcherutil.py
```diff
@@ -115,36 +117,22 @@ def reset_marker(self):
...
         :param hutil_log: Normal logging function (e.g., hutil.log). This is not a stream.
-        :param log_to_console: Indicates whether to log any issues to /dev/console or not.
         """
         self._hutil_error = hutil_error
         self._hutil_log = hutil_log
-        self._log_to_console = log_to_console
         self._consecutive_error_count = 0
         self._consecutive_restarts_due_to_error = 0
 
-    def _do_log_to_console_if_enabled(self, message):
```
hestolz: Unused ...

OmsAgent/watcherutil.py
```diff
@@ -229,14 +217,14 @@ def upload_telemetry(self):
                             self.write_waagent_event(event)
                             self._hutil_log("Successfully processed telemetry status file: "+sf)
 
-                        except Exception as e:
```
canfikret: Does object **e** contain anything useful?

OmsAgent/omsagent.py
```diff
@@ -17,6 +17,9 @@
 # limitations under the License.
 
 from __future__ import print_function
+from future import standard_library
```
canfikret: Is future module included in the base python?

hestolz: it is part of the standard library in 2.6+ and 3, but I don't think all distros have it ... I'll have to install it in the shim.

OmsAgent/watcherutil.py
hestolz: in any case we are calling `traceback.format_exc` after which should have the relevant info

OmsAgent/HandlerManifest.json
```diff
@@ -3,11 +3,11 @@
     "name":  "OmsAgentForLinux",
     "version": "1.12.25",
     "handlerManifest": {
-      "installCommand": "omsagent.py -install",
-      "uninstallCommand": "omsagent.py -uninstall",
-      "updateCommand": "omsagent.py -update",
-      "enableCommand": "omsagent.py -enable",
-      "disableCommand": "omsagent.py -disable",
+      "installCommand": "shim.sh -install",
```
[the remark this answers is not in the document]
hestolz: so I thought about that ... I felt that if you have `omsagent.py` and `omsagent.sh` it may seem initially confusing where the entrypoint is (without looking at this file), plus then their names are almost the same ... `shim.sh` more clearly delineates the purpose of this file. 

A compromise could be `omsagent_shim.sh` ?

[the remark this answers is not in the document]
canfikret: yes, good one. For example, if a customer sends us ps aux output; we can tell better...

OmsAgent/omsagent_shim.sh
```diff
@@ -0,0 +1,59 @@
...
+    if [ $? != 0 ]; then
+        apt-get update
+        apt-get install -y python-future
+    fi
+elif [ -f "/etc/redhat-release" ]; then # RHEL, CentOS, Oracle
+    rpm -qi python-future
+    if [ $? != 0 ]; then
+        yum install -y python-future
+    fi
+elif [ -f "/etc/os-release" ]; then # Possibly SLES, openSUSE
```
hestolz: can't parse `/etc/os-release` across the board because when I checked, not present on centos6
can't `-f /etc/SuSE-release` because it has been removed in SLES15 ...

OmsAgent/omsagent_shim.sh
```diff
@@ -0,0 +1,59 @@
...
+else
+    echo "Found `${PYTHON} --version`"
+fi
+
+# Install python-future dependency required for omsagent.py
+if [ -f "/etc/debian_version" ]; then # Ubuntu, Debian
+    dpkg-query -l python-future | grep ^ii
+    if [ $? != 0 ]; then
+        apt-get update
+        apt-get install -y python-future
```
hestolz: @canfikret necessary to check result of these install commands?

OmsAgent/omsagent.py
```diff
@@ -68,12 +71,15 @@
 ExtensionStateSubdirectory = 'state'
 
 # Commands
```
hestolz: @nidhanda @abenbachir can I get a quick review of just this part; rest is signed off by Fikret

nidhanda: Looks good to me.

PR #1022, opened 2020-02-18
Linux Bug fixes: Calling Handle.py only when command is enable, Having default values when settings are read from vmbackup.conf

suana-ms:


VMBackup/main/handle.py
```diff
@@ -195,6 +200,11 @@ def freeze_snapshot(timeout):
         if(hutil.get_value_from_configfile('seqsnapshot') == None):
             hutil.set_value_to_configfile('seqsnapshot', '0')
         seqsnapshotflag = hutil.get_value_from_configfile('seqsnapshot')
+        try:
```
vityagi: seqsnapshot is also read in file guestsnapshotter.py, make sure to look for all occurences of get_value_from_configfile

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -155,6 +155,11 @@ def log(self, message,level='Info'):
 
     def log_with_no_try_except(self, message, level='Info'):
         WriteLog = self.get_value_from_configfile('WriteLog')
```
vityagi: It will be cleaner to write wrappers around get_value_from_configfile()  with default-values as input and try-catch inside within them :
get_**str**_value_from_configfile(config, default_str_value)
get_**int**_value_from_configfile(config, default_int_value)

VMBackup/main/handle.sh
```diff
@@ -11,6 +11,12 @@ logfolder=$(echo $postsubstr | cut -b 1-$resultstrlen)
 logfile=$logfolder'/shell.log'
 
 rc=3
+
+if [ $1 != "enable"]
+then
+	echo "The command is not enable, exiting" >> $logfile
```
vityagi: trace date and command ($1) as well

VMBackup/main/handle.sh
```diff
@@ -11,6 +11,12 @@ logfolder=$(echo $postsubstr | cut -b 1-$resultstrlen)
...
+	exit $rc
```
vityagi: exitcode should be 0 here

VMBackup/main/handle.py
```diff
@@ -195,6 +200,11 @@ def freeze_snapshot(timeout):
         if(hutil.get_value_from_configfile('seqsnapshot') == None):
             hutil.set_value_to_configfile('seqsnapshot', '0')
         seqsnapshotflag = hutil.get_value_from_configfile('seqsnapshot')
```
bpramod: Same againg. seqsnapshot is not being used here. As a cleanup you can remove this logic.

VMBackup/main/handle.sh
bpramod: in current code flow, we are writing a status file for non enable calls as well. it will become a business logic change now.
Have you tested this scenario ?
@soumyapattnaik / @vityagi - can you please confirm ?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -335,8 +331,50 @@ def get_value_from_configfile(self, key):
...
+        value = default
+        configfile = '/etc/azure/vmbackup.conf'
+        try :
+            if os.path.exists(configfile):
+                config = ConfigParsers.ConfigParser()
+                config.read(configfile)
+                if config.has_option('SnapshotThread',key):
+                    value = config.get('SnapshotThread',key)
+        except Exception as e:
+            pass
```
vityagi: call get_value_from_configfile() instead of rewriting similar (above) code

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -335,8 +331,50 @@ def get_value_from_configfile(self, key):
...
```
vityagi: same, call call get_value_from_configfile()

VMBackup/main/handle.py
```diff
@@ -355,7 +341,7 @@ def daemon():
 
         if(para_parser.taskId is not None and para_parser.taskId != ""):
             backup_logger.log('taskId: ' + str(para_parser.taskId), True)
-            exit_if_same_taskId(para_parser.taskId) 
+            #exit_if_same_taskId(para_parser.taskId)
```
vityagi: remember to uncomment before merge

VMBackup/main/guestsnapshotter.py
```diff
@@ -340,25 +328,9 @@ def snapshotall_seq(self, paras, freezer, thaw_done, g_fsfreeze_on):
...
-                    self.logger.log("Config File doesn't have the key :" + key)
-        except Exception as e:
-            errorMsg = " Unable to ed config file.key is "+ key +"with error: %s, stack trace: %s" % (str(e), traceback.format_exc())
-            self.logger.log(errorMsg)
-        return value
-
     def snapshotall(self, paras, freezer, g_fsfreeze_on):
         thaw_done = False
-        if (self.get_value_from_configfile('seqsnapshot') == '1' or self.get_value_from_configfile('seqsnapshot') == '2' or (len(paras.blobs) <= 4)):
+        if (self.hutil.get_intvalue_from_configfile('seqsnapshot',0) == '1' or self.hutil.get_intvalue_from_configfile('seqsnapshot',0) == '2' or (len(paras.blobs) <= 4)):
```
vityagi: shouldn't 1 be used instead of '1'

VMBackup/main/handle.sh
vityagi: I don't think we write status for non-enable commands. But, please check and confirm @suana-ms

bpramod: @vityagi 
https://github.com/Azure/azure-linux-extensions/blob/8eafcba3f785abcd3e94a2b11a154e4a6cc1f49a/VMBackup/main/handle.py#L96

https://github.com/Azure/azure-linux-extensions/blob/8eafcba3f785abcd3e94a2b11a154e4a6cc1f49a/VMBackup/main/Utils/HandlerUtil.py#L688

**do_status_report** is the method where we write to status. So we are writing to the status file.

We might overwrite the status when backup runs as sequence number will be same

VMBackup/main/handle.py
```diff
@@ -192,10 +192,6 @@ def convert_time(utcTicks):
 def freeze_snapshot(timeout):
     try:
         global hutil,backup_logger,run_result,run_status,error_msg,freezer,freeze_result,para_parser,snapshot_info_array,g_fsfreeze_on
-        if(hutil.get_value_from_configfile('seqsnapshot') == None):
-            hutil.set_value_to_configfile('seqsnapshot', '0')
```
bpramod: dont remove **set_value_to_configfile**

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -332,6 +333,37 @@ def get_value_from_configfile(self, key):
             pass
 
         return value
+
+    def get_strvalue_from_configfile(self, key, default):
+        global backup_logger
```
bpramod: remove this

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -332,6 +333,37 @@ def get_value_from_configfile(self, key):
...
+        try :
+            value_str = str(value)
+        except ValueError :
+            self.log('Not able to parse the read value as string, falling back to default value', True, 'Warning')
+            value = default
+
+        return value
+
+    def get_intvalue_from_configfile(self, key, default):
+        global backup_logger
```
bpramod: here too

VMBackup/main/handle.sh
```diff
@@ -11,6 +11,13 @@ logfolder=$(echo $postsubstr | cut -b 1-$resultstrlen)
 logfile=$logfolder'/shell.log'
 
 rc=3
+arc=0
+
+if [ $1 != "enable"]
+then
+	echo "`date`- The command is $1, exiting" >> $logfile
+	exit arc
```
soumyapattnaik: exit $arc as you are refering to variable

VMBackup/HandlerManifest.json
```diff
@@ -10,6 +10,6 @@
             "updateCommand": "main/handle.sh update"
         }, 
         "name": "MyBackupTestLinuxInt", 
-        "version": "1.0.9120.0"
+        "version": "1.0.9121.0"
```
soumyapattnaik: not this file. revert this back
update prodmanifest.xml

PR #1029, opened 2020-02-22
Checking to prevent error messages from accidental onboarding to multiple workspace

nidhanda:
1) Added stopOnMultipleConnections flag to the HandlerManifest.json
2) Added checks for multiple connections in enable() and prepare_update() methods as well as fixed this check in install() method. 

Testing Strategy - 
1) Tested the enable() method's code snippet as that doesnt involve waagent downloading the latest extension zip so copying over the changes and running AZ CLI command to install to a new workspace worked as a test strategy for this case. 
2) Tested the install() method's code snippet by manually uninstalling from the VM and mock-creating a new workspace guid directory in /etc/opt/microsoft/omsagent directory and then manually running install to trigger the check.  
3) Will need to deploy to Canary to test for update method test


hestolz:
FYI looks like these are possible [remediation steps](https://bwit.blog/fix-please-set-stoponmultipleconnections-to-false/) (for the 'hackers')

OmsAgent/HandlerManifest.json
```diff
@@ -10,7 +10,9 @@
       "disableCommand": "omsagent_shim.sh -disable",
       "rebootAfterInstall": false,
       "reportHeartbeat": false,
-      "updateMode": "UpdateWithInstall"
+      "updateMode": "UpdateWithInstall",
+      "continueOnUpdateFailure": "true",
```
hestolz: I don't see this anywhere in the repo, where is it referenced?

nidhanda: We never really created the PR for adding this to the Fairfax github code (forgot to add it during the whole Fairfax release mitigation). This feature is part of the newer waagent versions - https://microsoft.sharepoint.com/:w:/t/ComputeVM/EX85ZKjTWqNOoN7P5rZBRKgBRZ5cXR-TyTxlWttCi4ICOg?e=HU3Uhm 
This change is currently part of the 1.12.17 Fairfax extension's HandlerManifest.json as that is what allowed us to ignore the failed disable status and continue with newer version installation.

hestolz: Ok so it is a flag just for use by guest agent

nidhanda: Yes, correct.

nidhanda: Removed the "continueOnUpdateFailure": "true", flag from this PR. Will create a new PR for that change since it is out of scope for the general topic of this PR

PR #1032, opened 2020-02-26
ADE: Extension name configuration and override.

ivanp-ms:
Getting extension name from a dedicated text file configurable at build time. Added YAML definitions for Test1, Test2, and Test3 extensions. Refactored helper functions for getting configuration from text files.


VMEncryption/cdpx_build.sh
```diff
@@ -18,5 +18,15 @@ echo "ADE source folder: " + $PWD
 # BUGBUG: Keeping the static version until we figure out the versioning strategy (PBI 6218633)
 # echo CDP_FILE_VERSION_NUMERIC_NOLEADINGZEROS > main/version.txt
 
+# Extension name overried if specified in the YAML file
```
ejarvi: typo: override

VMEncryption/main/ConfigFromFile.py
```diff
@@ -18,11 +18,17 @@
...
     if not os.path.exists(file_path):
         raise IOError('Unexpected: version.txt file cannot be found at [%s]' % file_path)
     
     with open(file_path) as text_file:
-        return text_file.read()
\ No newline at end of file
+        return text_file.read().strip()
+
+def get_build_version():
+    return get_config_value_from_file('version.txt')
```
vimish: Will it be possible to have one file with both version and name?

VMEncryption/main/ConfigFromFile.py
```diff
@@ -18,11 +18,17 @@
...
 import os
 
-def get_build_version_from_file():
-    # Note, version.txt is expected to be within the same directory as this file
-    file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'version.txt')
+def get_config_value_from_file(file_name):
+    # Note,the config file is expected to be within the same directory as this file
+    file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), file_name)
     if not os.path.exists(file_path):
         raise IOError('Unexpected: version.txt file cannot be found at [%s]' % file_path)
```
vimish: The error message should use file_name instead of hardcoded version.txt

VMEncryption/main/ConfigFromFile.py
ivanp-ms: I was thinking the same but with any file format, the build code would need to understand it to be able to modify. If we keep adding config values, we should probably move them to a JSON file or so.

ejarvi: Yes, consolidating any values being introduced by the build system into a single config file (rather than multiple) would be nice.

ivanp-ms: Agree. Let me see how that would work with JSON.

ejarvi: ConfigParser may also be an option - https://docs.python.org/2/library/configparser.html

ivanp-ms: Switched to a common JSON file.

VMEncryption/test/test_common.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+import os
+import main.Common
+import json
+
+class TestCommonVariables(unittest.TestCase):
+    
+    def test_common_variables(self):
+        self.assertEqual(main.Common.CommonVariables.extension_name, "AzureDiskEncryptionForLinux")
+        self.assertEqual(main.Common.CommonVariables.extension_version, "1.1.0.47")
```
vimish: Instead of hardcoding values, it would be good if values can be directly read from file using File I/O and compared to values in Common.py. Otherwise we will need to manually upgrade the tests on every version upgrade. I am fine if you want to this change in another PR.

ivanp-ms: Good catch - thank you! I forgot to remove those hardcoded values. PR updated.

PR #1033, opened 2020-02-28
First checking for azure attestation linux extension

shkhisti:


AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+    # Determine the operation being executed
+    operation = None
+    try:
+        option = sys.argv[1]
+        if re.match('^([-/]*)(disable)', option):
+            operation = 'Disable'
+        elif re.match('^([-/]*)(uninstall)', option):
+            operation = 'Uninstall'
+        elif re.match('^([-/]*)(install)', option):
+            operation = 'Install'
```
bhupeshpant19jan: Why can't we use Argparse for parsing args here? They are more robust and more elegant.

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+  
+
+def install():
+    """
+    Ensure that this VM distro and version are supported.    
+    """
+    
+    # We want to check if extension supports current distro if not 
+    # then we want to exit with exit code 51 which is special error code 
+    # that waagent uderstands as unsuuported OS/distro.
```
bhupeshpant19jan: typo on "un-supported"

AzureAttestationExtension/manifest.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version='1.0' encoding='utf-8' ?>
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
+  <ProviderNameSpace>Microsoft.Azure.TestAttestation</ProviderNameSpace>
```
bhupeshpant19jan: why the provider namespace is named as TestAttestation, is this manifest file for test extension? I am assuming one for the actual extension as well?

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+
+from threading import Thread
+
+try:
+    from Utils.WAAgentUtil import waagent
+    import Utils.HandlerUtil as HUtil    
+    
+except Exception as e:
+    # These utils have checks around the use of them; this is not an exit case
+    print('Importing utils failed with error: {0}'.format(e))
```
bhupeshpant19jan: Incase WAAgentUtil module couldn't be imported we should log and throw or exit from here. At some places, waagent is used without checking and as the first instance, at init_waagent_logger, it will throw the undefined symbols. (Also a misleading error)

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+# Configuration
+HUtilObject = None
+SettingsSequenceNumber = None
+HandlerEnvironment = None
+SettingsDict = None
+
+# Default configuration settings
+MaaEndpoint = "https://microsoft.attestation.net/",
+AscReportingEndpoint = "https://management.azure.com/",
+UseCustomeToken = "false"
```
bhupeshpant19jan: typo: useCustomToken

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+
+# Default configuration settings
+MaaEndpoint = "https://microsoft.attestation.net/",
+AscReportingEndpoint = "https://management.azure.com/",
+UseCustomeToken = "false"        
+    
+# Change permission of log path - if we fail, that is not an exit case
+# like any other extension attestation extension is started as a root so we should be good here.
+try:
+    ext_log_path = '/var/log/azure/'
```
bhupeshpant19jan: Is this a standard log path? SHouldn't it be somewhere under tmp directory?
Any idea what should be for Windows?

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+    """
+    Initialize a HandlerUtil object for this operation.
+    If the required modules have not been imported, this will return None.
+    """
+    hutil = None
+    if ('Utils.WAAgentUtil' in sys.modules
+            and 'Utils.HandlerUtil' in sys.modules):
+        try:
+            
+            logFileName = 'extension.log'
```
bhupeshpant19jan: extension.log is a very generic name. Can't we have something like GuestAttestationExtension.log or AttestationExtension.log?

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+                  'Stacktrace: {2}'.format(operation, e,
+                                           traceback.format_exc())
+     
+    # Finish up and log messages
+    log_and_exit(operation, exit_code, message)        
+  
+
+def install():
+    """
+    Ensure that this VM distro and version are supported.
```
bhupeshpant19jan: shouldn't install also require the binaries to be copied at a specified destination?
My understanding is as a part of deployment zip file will be downloaded and unzipped to some temp dir. Once we install the extension the binaries(Guest attestation that Vinay is working on) will be copied to /var/lib/ or some specific directories. 
Please correct me if my understanding is not correct.

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+    else:
+        return HUtilObject._context._log_dir
+
+def start_attestation_client():
+    """
+    Start attestation process that performs periodic monitoring of attestion process.
+    :return: None
+
+    """
+    stop_process('attestation_client')
```
bhupeshpant19jan: Why are we trying to stop the attestation_client? We could check if it is already running we can simply exit from there?

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+        with open(pids_filepath, "r") as f:
+            for pids in f.readlines():
+                kill_cmd = "kill " + pids
+                run_command_and_log(kill_cmd)
+                run_command_and_log("rm "+pids_filepath)
+
+def stop_attestation_client_process():    
+    stop_process('attestation_extension')
+
+    return 0, "Successfully stopped attestation proces"
```
bhupeshpant19jan: I couldn't find the definition of the stop_Process, hence not sure about the behavior of error case handling. I still feel we are returning 0 without checking the return code of stop_process.

AzureAttestationExtension/agent.py
shkhisti: good suggestion. Bit newbie with python

AzureAttestationExtension/agent.py
shkhisti: thanks, fixed

AzureAttestationExtension/manifest.xml
shkhisti: discussed offline

AzureAttestationExtension/agent.py
shkhisti: I will change other places where we are using waagent without checking.

AzureAttestationExtension/agent.py
shkhisti: thanks will make the change

AzureAttestationExtension/agent.py
shkhisti: will fix it.

AzureAttestationExtension/agent.py
bhupeshpant19jan: As per our offline discussion, this path should come from the config.json file, provided by the agent.

AzureAttestationExtension/agent.py
bhupeshpant19jan: I have referred other python extension code and they are also written in the same way so this approach seems pretty much standard so I would suggest using it as it is.

AzureAttestationExtension/agent.py
```diff
@@ -0,0 +1,708 @@
...
+        HUtilObject = parse_context(operation)
+        exit_code, output = operations[operation]()
+
+        # Exit code 1 indicates a general problem that doesn't have a more
+        # specific error code; it often indicates a missing dependency. Although
+        # we don't have it right now, we should handle it properly.
+        if exit_code is 1 and operation == 'Install':
+            message = 'Install failed with exit code 1. Please check that ' \
+                      'dependencies are installed. For details, check logs ' \
+                      'in /var/log/azure/Microsoft.Azure.Attestation' \
```
bhupeshpant19jan: **/var/log/azure/** THis path should come from config file installed by agent.

AzureAttestationExtension/agent.py
bhupeshpant19jan: Also small typo in logging: **proces**

PR #1040, opened 2020-03-06
[ADE SP] Fix Resource Disk only encryption behavior

vermashi:


VMEncryption/main/handle.py
```diff
@@ -632,6 +632,19 @@ def enable():
             # Migrate to early unlock if using crypt mount
             if crypt_mount_config_util.should_use_azure_crypt_mount():
                 crypt_mount_config_util.migrate_crypt_items()
+        elif ResourceDiskUtil.RD_MAPPER_NAME in [ci.mapper_name for ci in crypt_mount_config_util.get_crypt_items()]:
```
vimish: Should we check Resource disk is the only crypt item? To be safe from scenarios where passphrase is missing for some other reason (like customer accidentally deleting it).

vermashi: If they did something like that it will fail later anyways. I feel we should still at least mount the RD. Plus, adding the check is dangerous if they have some devices not managed by ADE in the crypttab.

PR #1043, opened 2020-03-09
Use rpm for Oracle Linux

simathih:


hestolz:
@simathih shouldn't you make this change for suse as well? AFAIK zypper works with RPMs

nidhanda:
> @simathih shouldn't you make this change for suse as well? AFAIK zypper works with RPMs

I'll confirm it on Suse and create a new PR.

PR #1044, opened 2020-03-10
Changes for bug fixes in recent change while  reading vmbackupconf

suana-ms:


VMBackup/main/handle.sh
```diff
@@ -13,10 +13,11 @@ logfile=$logfolder'/shell.log'
 rc=3
 arc=0
 
-if [ $1 != "enable"]
+if [ [ $1 != "enable"  ] && [ $1 != "deamon" ] ]
```
vityagi: I think it should be "daemon". Was the spelling correct at test (installation) path and got changed while doing the changes to repo local enlistment file ? Usually I follow below steps to avoid such mistakes :
1. Create a Linux VM, enable backup on it and enlist the extension repo also on the same VM.
2. Do required changes in the files of the installed extension and test
3. Copy/overwrite the changed files to repo. This avoids doing same changes manually again which always has chances of mistakes.

suana-ms: yes, Vikas. It should have been daemon. Both my enlisted copy and the test machine had the incorrect "deamon" only. But the backups were succeeding because there was a small syntax error but the shell script ignored that syntax error and continued with handle.py invocation. I have fixed and tested the "install " scenario now and it works as expected.

PR #1048, opened 2020-03-13
OMS Whitelist RHEL8

hestolz:


OmsAgent/omsagent.py
```diff
@@ -760,9 +760,9 @@ def is_vm_supported_for_extension():
     The supported distros of the OMSAgent-for-Linux are allowed to utilize
     this VM extension. All other distros will get error code 51
     """
-    supported_dists = {'redhat' : ['6', '7'], # CentOS
```
hestolz: I am not sure why this was commented CentOS. Using our docker containers, I checked CentOS, Oracle, and RHEL each for 6, 7, and 8. None of them had a non-corresponding distro name. It is possible images from certain distro maintainers will return different distro names through the `platform.dist()` / `platform.linux_distribution()` interface. But I think it is better to err on the side of ensuring RHEL8 works with the small risk that some unusual CentOS/Oracle 8 images will accidentally not be stopped by this distro detection logic.

PR #1055, opened 2020-03-19
Disable cmd error check during signing + add more logging

abenbachir:


OmsAgent/omsagent.py
```diff
@@ -150,7 +150,7 @@ def verifyShellBundleSigningAndChecksum():
         raise Exception("Unable to find the dscgpgkey.asc file at " + dscGPGKeyFilePath)
 
     importGPGKeyCommand = "sh ImportGPGkey.sh " + dscGPGKeyFilePath
-    exit_code, output = run_command_with_retries_output(importGPGKeyCommand, retries = 0, retry_check = retry_skip)
+    exit_code, output = run_command_with_retries_output(importGPGKeyCommand, retries = 0, retry_check = retry_skip, check_error = False)
```
hestolz: should you be checking `exit_code`

hestolz:
I don't think you should be setting `check_error = False`. If you do so, `waagent.RunGetOutput` (the underlying function underneath several layers in `run_command_with_retries_output`) will always return 0. So none of the checks you are doing for `exit_code` will be valid anymore.

abenbachir:
> I don't think you should be setting `check_error = False`. If you do so, `waagent.RunGetOutput` (the underlying function underneath several layers in `run_command_with_retries_output`) will always return 0. So none of the checks you are doing for `exit_code` will be valid anymore.

Not true, we will still get the exit code:
```
    try:                                     
        output=subprocess.check_output(cmd,stderr=subprocess.STDOUT,shell=True)
    except subprocess.CalledProcessError,e :
        if chk_err :
            Error('CalledProcessError.  Error Code is ' + str(e.returncode)  )
            Error('CalledProcessError.  Command string was ' + e.cmd  )
            Error('CalledProcessError.  Command result was ' + (e.output[:-1]).decode('latin-1'))
        return e.returncode,e.output.decode('latin-1') # <---------------------------------  Here
    return 0,output.decode('latin-1')

```

OmsAgent/omsagent.py
hestolz: should check `exit_code` here?

abenbachir: We don't care about the exit_code  at that place because I do more check down below.

PR #1056, opened 2020-03-20
enable logrotate in linux ectension

vishak-ms:


VMBackup/main/handle.py
```diff
@@ -534,6 +534,7 @@ def update():
 def enable():
     global backup_logger,hutil,error_msg,para_parser,patch_class_name,orig_distro
     try:
+        do_log_rotate('./main/config/extension.logrotate')
```
soumyapattnaik: use absolute path by using something like os.path.join(os.getcwd(), "main/config/extension.logrotate")

VMBackup/main/handle.py
```diff
@@ -564,6 +565,17 @@ def enable():
         error_msg = 'Failed to call the daemon'
         exit_with_commit_log(temp_status, temp_result,error_msg, para_parser)
 
+def set_file_permission(file_path, permission):
+    os.system('chmod ' + permission + " " + file_path)
```
soumyapattnaik: please move this to HandlerUtil

VMBackup/main/handle.py
```diff
@@ -564,6 +565,17 @@ def enable():
...
+
+def convert_dos_to_unix(file_path):
+    os.system('dos2unix ' + file_path)
+
+def do_log_rotate(config_file_path):
```
soumyapattnaik: infact move all the method to Utils/HandlerUtil

VMBackup/main/handle.py
soumyapattnaik: setting permission at every enable is not recommended. Change permission only if it is not 0644

VMBackup/main/handle.py
```diff
@@ -564,6 +565,17 @@ def enable():
         error_msg = 'Failed to call the daemon'
         exit_with_commit_log(temp_status, temp_result,error_msg, para_parser)
 
+def set_file_permission(file_path, permission):
+    os.system('chmod ' + permission + " " + file_path)
+
+def convert_dos_to_unix(file_path):
+    os.system('dos2unix ' + file_path)
```
soumyapattnaik: even dont do it all the time. check if dos2unix is needed then perform the step.

VMBackup/main/handle.py
```diff
@@ -564,6 +565,17 @@ def enable():
...
+
+def do_log_rotate(config_file_path):
+    set_file_permission(config_file_path, "0644")
+    convert_dos_to_unix(config_file_path)
+    os.system('logrotate ' + config_file_path)
```
soumyapattnaik: there is no exception handling. what if logrotate keeps failing. There is no logging on what was generated on successful completion of logrotate.

VMBackup/main/config/extension.logrotate
```diff
@@ -0,0 +1,23 @@
...
+    # the compressed files would not go beyond ~2 MB per day.
+    size 20M
+    rotate 50
+
+    # Add date as extension when rotating logs
+    dateext
+
+    # Formatting the date extension to YYYY-MM-DD-SSSSSSS format. logrotate does not provide hours, mins and seconds
+    # option. Adding the %s (system clock epoch time) to differentiate rotated log files within the same day.
+    dateformat -%Y-%m-%d-%s
```
soumyapattnaik: format is UTC or machine time?

VMBackup/main/config/extension.logrotate
```diff
@@ -0,0 +1,23 @@
+/var/log/azure/Microsoft.Azure.RecoveryServices.VMSnapshotLinux/extension.log {
```
soumyapattnaik: add handling for extension running with older agent, where log file is created inside extension extension version folder.

VMBackup/main/config/extension.logrotate
```diff
@@ -0,0 +1,23 @@
...
+    # Old versions of log files are compressed with gzip by default.
+    compress
```
soumyapattnaik: if gzip util is missing in the VM then what is the expected handling?

vrdmr:
This should be taken care of logrotate configuration added in WALinuxAgent, starting 2.2.46 - https://github.com/Azure/WALinuxAgent/blob/develop/config/waagent-extn.logrotate. If you add this logrotate, it would conflict with this logrotate configuration and can cause issues.

/cc @pgombar

VMBackup/main/config/extension.logrotate
vishak-ms: gzip is unix system cmd. So it should be present. Also logrotate is being used by other agents, so we should be safe. In worst case it won't do logrotate and silently fail without exceptions.

VMBackup/main/config/extension.logrotate
vishak-ms: machine time

soumyapattnaik:
no more needed

PR #1057, opened 2020-03-20
Bug fix DistInfo() should return 'Default' instead of 'Abstract'

harvek:
In DistInfo(), when an exception occurs (e.g. if the os-release
file is missing), ['Abstract','1.0'] is returned.
In that case GetMyPatching() tries to instantiate AbstractPatching().
This will fail because the import for AbstractPatching is
missing.

DistInfo() should return ['Default','1.0'].


harvek:
ping

PR #1058, opened 2020-03-20
Add python-future as a submodule

hestolz:
- Original approach of using package managers to install python-future is not viable (package is not hosted in standard RPM repos), neither is using pip to install future (same reason, installing even pip is not straightforward/reliable).
- Instead, let's add python-future as a submodule. When we create the extension zip with `packaging.sh` the source files will be copied to `./ext/future`, which we add to PYTHONPATH.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -17,16 +17,21 @@
 # limitations under the License.
 
 from __future__ import print_function
-from future import standard_library
-standard_library.install_aliases()
-from builtins import str
+import sys
+
+# future imports have no effect on python 3 (verified in official docs)
+# importing from source causes import errors on python 3, lets skip import
+if sys.version_info[0] < 3:
+    from future import standard_library
+    standard_library.install_aliases()
+    from builtins import str
+
 import os
 import os.path
 import signal
 import pwd
 import grp
 import re
-import sys
 import traceback
 import time
 import platform
--- a/OmsAgent/omsagent_shim.sh
+++ b/OmsAgent/omsagent_shim.sh
@@ -1,24 +1,28 @@
 #!/usr/bin/env bash
 
-# The entry point for the OMS extension through which the correct python version (if any) is used to invoke omsagent.py. 
+# The entry point for the OMS extension through which the correct python version (if any) is used to invoke omsagent.py.
 # We default to python2 and always invoke with the versioned python command to accomodate the RHEL 8+ python strategy.
 # Control arguments passed to the shim are redirected to omsagent.py without validation.
 
 COMMAND="./omsagent.py"
 PYTHON=""
+FUTURE_PATH=""
 ARG="$@"
 
 function find_python() {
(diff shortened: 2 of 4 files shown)


ejarvi:
@nkuchta ping

OmsAgent/omsagent_shim.sh
```diff
@@ -39,42 +28,5 @@ else
...
-    run_command "yum install -y python-future" $ACTION
-elif [ -x "$(command -v zypper)" ]; then
-    run_command "zypper --non-interactive install python-future" $ACTION
-else
-    echo "Unsupported or indeterminable operating system" >&2
-    exit 51
-fi
-
-${PYTHON} ${COMMAND} ${ARG}
+PYTHONPATH=${PWD}/ext/future:${PYTHONPATH} ${PYTHON} ${COMMAND} ${ARG}
```
vrdmr: Along with `PYTHONPATH`, do you need to also set up `PATH`?

OmsAgent/omsagent.py
```diff
@@ -17,16 +17,21 @@
 # limitations under the License.
 
 from __future__ import print_function
-from future import standard_library
-standard_library.install_aliases()
-from builtins import str
+import sys
+
+# future imports have no effect on python 3 (verified in official docs)
```
hestolz: > The imports have no effect on Python 3. On Python 2, they shadow the corresponding builtins, which normally have different semantics on Python 3 versus 2, to provide their Python 3 semantics. [(source)](https://pypi.org/project/future/)

OmsAgent/omsagent_shim.sh
hestolz: Doesn't seem necessary from my testing

OmsAgent/packaging.sh
```diff
@@ -43,7 +45,7 @@ fi
 
 echo "Packaging extension $PACKAGE_NAME to $output_path"
 excluded_files="omsagent.version packaging.sh apply_version.sh update_version.sh"
-zip -r $output_path/$PACKAGE_NAME * -x $excluded_files "./Fairfax/*" "./Mooncake/*" "./test/*" "./extension-test/*" "./references"
```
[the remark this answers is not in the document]
hestolz: I removed those actually, now we will only maintain omsagent.py, just need to uncomment the correct line here https://github.com/Azure/azure-linux-extensions/blob/master/OmsAgent/omsagent.py#L81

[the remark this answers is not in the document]
nidhanda: The Fairfax/ and MC/ have always been redundant. They contained renamed copies of the omsagent.py, manifest.xml and handlermanifest.json. However the contents differed only with the line Henry mentioned above which contains the azure endpoint for each cloud . So now, thanks to this change, we can simply uncomment the respective endpoints and create the zips.

hestolz:
@nkuchta review is blocking on you, need to add a git submodule. thanks

PR #1062, opened 2020-03-27
Get python-future with pip

hestolz:
An alternative to the submodule approach in #1058 
- Pro: more "supported" way of getting a new python module, via pip; doesn't require as many steps during extension packaging; avoids hacky python version dependent path/imports setting
- Con: may be more prone to failure due to pip download failure, also apparently there can be some issues depending on whether chose to install pip itself or packages USING pip with or without the `--user` flag (issue threads [1](https://github.com/pypa/pip/issues/1668), [2](https://github.com/pypa/pip/issues/7209) ...)


HussainHacker162 (commented):
pkg

PR #1067, opened 2020-03-31
PR for workload telemetry

soumyapattnaik:


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -784,6 +786,22 @@ def UriHasSpecialCharacters(self, blobs):
 
         return uriHasSpecialCharacters
 
+    def get_workload_running(self):
+        workloads = []
```
bpramod, line 790: add everything in try-catch

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -784,6 +786,22 @@ def UriHasSpecialCharacters(self, blobs):
...
+        dblist= ["mysqld","postgres","oracle","cassandra",",mongo"]
+        if os.path.isdir("/proc"):
+            pids = [pid for pid in os.listdir('/proc') if pid.isdigit()]
+            for pid in pids:
+                pname = open(os.path.join('/proc', pid, 'cmdline'), 'rb').read()
+                for db in dblist :
+                    if db in pname and db not in workloads :
```
bpramod, line 797: compare the string by ignoring the case

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -784,6 +786,22 @@ def UriHasSpecialCharacters(self, blobs):
 
         return uriHasSpecialCharacters
 
+    def get_workload_running(self):
```
bpramod, line 789: why are we not using PF command ?

VMBackup/main/backuplogger.py
```diff
@@ -32,9 +32,14 @@ def __init__(self, hutil):
         self.enforced_local_flag_value = True
         self.hutil = hutil
         self.prev_log = ''
+        self.logging_off = False
 
     def enforce_local_flag(self, enforced_local):
-        if (self.enforced_local_flag_value != False and enforced_local == False):
+        if (self.hutil.get_intvalue_from_configfile('LoggingOff', 0) == 1):
```
bpramod, line 38: can we rename as disable logging?

bpramod, line 38: why are we using this ? as we already disable logging in freeze
where do we unset this again?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -784,6 +786,22 @@ def UriHasSpecialCharacters(self, blobs):
...
+        dblist= ["mysqld","postgres","oracle","cassandra",",mongo"]
+        if os.path.isdir("/proc"):
+            pids = [pid for pid in os.listdir('/proc') if pid.isdigit()]
+            for pid in pids:
+                pname = open(os.path.join('/proc', pid, 'cmdline'), 'rb').read()
+                for db in dblist :
+                    if db in pname and db not in workloads :
+                        self.log("workload running found with command : " + str(pname))
+                        workloads.append(db)
+                    else :
```
bpramod, line 800: is "else: pass" needed?
does it add value ?

VMBackup/main/backuplogger.py
```diff
@@ -43,6 +48,8 @@ def enforce_local_flag(self, enforced_local):
 
     """description of class"""
     def log(self, msg, local=False, level='Info'):
+        if(self.enforced_local_flag_value == False and self.logging_off == True):
```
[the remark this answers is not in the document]
soumyapattnaik: yes but if that is the issue then the backup should always fail and not intermittently. right?

VMBackup/main/Utils/HandlerUtil.py
soumyapattnaik: this is a better approach than using ps commands. also this is being used in guest agent code. so its tried and tested.

VMBackup/main/backuplogger.py
soumyapattnaik: no. already provided private to a customer

soumyapattnaik: logging_off will are only doing during fsfreeze phase i.e not to put logs in memory

VMBackup/main/Utils/HandlerUtil.py
soumyapattnaik: done

VMBackup/main/backuplogger.py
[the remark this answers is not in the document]
vityagi: We can't say surely. It might be that during some runs a resources is locked by some process etc., which is needed to write to console (while in other runs when it succeed, it might not be locked). Similar to memory as it is succeeding also sometimes.
I am just saying it looks to be worth trying because there are below kind of sequence also of traces in customer logs, which does not explain how memory getting freed up in between of freeze (in this run Freeze continued upto **00:35:45**)
2020/01/20 00:38:21 2020-01-20 **00:34:45.047363**  Info  freezed 
2020/01/20 00:38:21 2020-01-20 **00:35:13.795149**  Info  ****** 4. Freeze Completed (Signal=1 received) 
2020/01/20 00:38:21 2020-01-20 00:37:09.995405  Info  some child process terminated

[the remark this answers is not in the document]
vityagi: I am not suggesting that logging_off change should not be done. I think it will be useful anyways. But, asking to also check if console change works for this particular customer.

VMBackup/main/safefreeze/src/safefreeze.c
```diff
@@ -103,15 +105,15 @@ int main(int argc, char *argv[])
 
         if ((fileSystemDescriptors[i] = open(mountPoint, O_RDONLY)) < 0)
```
soumyapattnaik, line 106: @arunjvs do we need to change the O_RDONLY to O_NONBLOCK

VMBackup/main/backuplogger.py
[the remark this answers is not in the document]
soumyapattnaik: ok.. will share the same private to the customer.

VMBackup/main/safefreeze/src/safefreeze.c
```diff
@@ -103,15 +105,15 @@ int main(int argc, char *argv[])
...
         {
-            logger("Failed to open: %s\n", mountPoint);
+            logger("Failed to open: %s with error: %d and error message: %s\n", mountPoint, fileSystemDescriptors[i], strerror(errno));
```
arunjvs, line 108: BUG. errno is a global variable that must be captured `*immediately*` after the failing system call/library call. Subsequent calls like printf/time/etc. that you are calling in logger will overwrite it. Please fix it in all places below.

See the NOTES section in the `man errno` page. Quoting here some parts

```
NOTES
       A common mistake is to do

           if (somecall() == -1) {
               printf("somecall() failed\n");
               if (errno == ...) { ... }
           }

       where  errno  no  longer  needs  to  have the value it had upon return from somecall() (i.e., it may have been
       changed by the printf(3)).  If the value of errno should be preserved across a library call, it must be saved:

           if (somecall() == -1) {
               int errsv = errno;
               printf("somecall() failed\n");
               if (errsv == ...) { ... }
           }
```

VMBackup/main/safefreeze/src/safefreeze.c
soumyapattnaik: there is no specific case but in ioctl man page i saw this being recommended.

PR #1071, opened 2020-04-02
Support for newer distros by adding default distro; gitignore updated

vrdmr:
With newer distros coming up in Azure, some of the extension scenarios are failing because of the failing GetMyDistro() call, which gives a bad extensions experience to those VMs. I add a new DefaultDistro scenario which could help in preventing such scenarios and try to present a better user experience  in such cases.

Also update gitignore to prevent venv, pycharm and mac os files to be committed.

### DIFF
--- a/Common/WALinuxAgent-2.0.16/waagent
+++ b/Common/WALinuxAgent-2.0.16/waagent
@@ -743,8 +743,7 @@ class AbstractDistro(object):
 
 class DefaultDistro(AbstractDistro):
     """
-    Asianux Distro concrete class
-    Put Asianux specific behavior here...
+    Default Distro concrete class: This class serves as a default OS behavior class.
     """
 
     def startDHCP(self):
--- a/Common/WALinuxAgent-2.0.16/waagent
+++ b/Common/WALinuxAgent-2.0.16/waagent
@@ -741,6 +741,30 @@ class AbstractDistro(object):
         pass
 
 
+class DefaultDistro(AbstractDistro):
+    """
+    Asianux Distro concrete class
+    Put Asianux specific behavior here...
+    """
+
+    def startDHCP(self):
+        """
+        Following the pattern used in WALinuxAgent for Default distro: This method is not implemented in the default
+        case.
+        """
+        pass
+
+    def stopDHCP(self):
+        """
+        Following the pattern used in WALinuxAgent for Default distro: This method is not implemented in the default
+        case.
+        """
+        pass
+
+    def __init__(self):
+        super(DefaultDistro, self).__init__()
+
+
 ############################################################
 #   GentooDistro
 ############################################################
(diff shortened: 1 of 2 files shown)


Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -741,6 +741,30 @@ class AbstractDistro(object):
         pass
 
 
+class DefaultDistro(AbstractDistro):
+    """
+    Asianux Distro concrete class
```
jscalev, line 746: Shouldn't it be "default Distro"?

ejarvi:
From the context of Azure Disk Encryption team review - The list of distros supported by guest agent is already a superset of those distros currently supported by the Azure Disk Encryption extension which does its own checking.  The list of distros supported by ADE is maintained here::  https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption-overview#supported-operating-systems

hestolz:
Likewise for [OMS](https://github.com/Azure/azure-linux-extensions/blob/master/OmsAgent/omsagent.py#L769) and [LAD](https://github.com/Azure/azure-linux-extensions/blob/master/Diagnostic/DistroSpecific.py#L318)

Common/WALinuxAgent-2.0.16/waagent
vrdmr: Sorry - Yup. Will fix this.

Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -6493,9 +6517,14 @@ def GetMyDistro(dist_class_name=''):
     else:
         Distro = dist_class_name
     if dist_class_name not in globals():
-        print(Distro + ' is not a supported distribution.')
-        return None
-    return globals()[dist_class_name]()  # the distro class inside this module.
+        msg = Distro + ' is not a supported distribution. Reverting to DefaultDistro to support scenarios in ' \
+                       'unknown/unsupported distribution.'
+        print(msg)
```
pgombar, line 6522: Remove?

vrdmr:
@ejarvi @hestolz Thanks for confirming.
@Azure/azure-agent-extensions @Azure/walinuxagent Can you please take a look at this PR and approve?

Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -741,6 +741,29 @@ class AbstractDistro(object):
         pass
 
 
+class DefaultDistro(AbstractDistro):
+    """
+    Default Distro concrete class: This class serves as a default OS behavior class.
+    """
+
+    def startDHCP(self):
```
D1v38om83r, line 749: How much functionality do we lose by having these functions be noop?

vrdmr: I validated - `startDHCP` and `stopDHCP` are not used by any extension directly, and the implementation is only used in the `Agent.Run`, but we don't ever call that (that is the way to run the old agent).

PR #1077, opened 2020-04-14
Avoiding leaked snapshots 

suana-ms:
Currently, this new error code doesn't come up as extensionErrorCode in BCmbackupStats. As I got to know, for it to show as ExtensionerrorCode in BCMbackup stats, this error Code with the same enum number value has to be present in the Status Codes in windows extension, that error code is already there. So I assume no other changes are required in the linux extension for the same. Please let me know if it is otherwise.

### DIFF
--- a/VMBackup/main/handle.py
+++ b/VMBackup/main/handle.py
@@ -341,6 +341,7 @@ def daemon():
 
             if total_span_in_seconds > MAX_TIMESPAN :
                 error_msg = "CRP timeout limit has reached, will not take snapshot."
+                errMsg = error_msg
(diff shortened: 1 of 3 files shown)


VMBackup/main/handle.py
```diff
@@ -334,11 +334,19 @@ def daemon():
             utcNow = datetime.datetime.utcnow()
             backup_logger.log('command start time is ' + str(commandStartTime) + " and utcNow is " + str(utcNow), True)
```
soumyapattnaik, line 335: check if command start time is coming from BCM or CRP. If BCM, check if with every retry are we changing this command start time.

suana-ms: this is sent by crp in case of managed vm. In case of unmanaged VM where this field is set by BCM. and only for managed VM it takes > 150 mins to invoke guest agent.

PR #1080, opened 2020-04-21
Fix data disk not getting mounted after disabling encryption.

vimish:


VMEncryption/main/DiskUtil.py
```diff
@@ -319,6 +319,10 @@ def umount(self, path):
         return self.command_executor.Execute(umount_cmd)
 
     def mount_all(self):
+        # Reload systemd daemon to get lastest changes from fstab
+        # pidof systemd seems unreliable on Ubuntu hence directly invoking systemctl
```
ejarvi, line 323: actual: "pidof"
expected: "pid of"

VMEncryption/main/DiskUtil.py
```diff
@@ -319,6 +319,10 @@ def umount(self, path):
...
+        # This command will just fail without side effects on system without systemd
+        self.command_executor.Execute('systemctl daemon-reload')
```
ejarvi, line 325: if a failure is expected in some cases, consider disabling logging so it doesn't look like an unexpected error in the logs on those environments (I think this is a flag that can be sent)

VMEncryption/main/DiskUtil.py
vimish: I was trying to use command **pidof systemd** to determine if systemd is in use. But it turns out that this command was not very reliable on Ubuntu. Hence, just wrote the command and reason not to use it here.

VMEncryption/main/DiskUtil.py
vimish: My intention not to suppress logging was to get evidence in the log that this command was actually executed. I agree that it might cause confusion on platforms that do not use systemd. 
I will add explicit log message to tell that we tried to execute the command and suppress logging for the actual execution.

VMEncryption/main/DiskUtil.py
ejarvi: 👍

PR #1090, opened 2020-05-01
Add support for NVMe disks.

vimish:
1. NVMe disks will only be supported with encrypt format all. They will be skipped for standard encryption
2. Their dev path will be UUID based.
3. Stale NVMe disks entry will be cleaned from fstab/crypttab (NVMe disks are recreated after stop/start)

### DIFF
--- a/VMEncryption/main/BekUtil.py
+++ b/VMEncryption/main/BekUtil.py
@@ -19,6 +19,7 @@
 from Common import TestHooks
 import base64
 import os.path
+import shutil
 
 """
 add retry-logic to the network api call.
@@ -82,6 +83,8 @@ class BekUtil(object):
 
             for file in os.listdir(self.bek_filesystem_mount_point):
                 if bek_filename in file:
+                    # Make sure LinuxPassPhraseFileName exist for Resource and Nvme disks 
+                    shutil.copy2(os.path.join(self.bek_filesystem_mount_point, file), os.path.join(self.bek_filesystem_mount_point, bek_filename))
                     return os.path.join(self.bek_filesystem_mount_point, file)
 
         except Exception as e:
--- a/VMEncryption/main/CryptMountConfigUtil.py
+++ b/VMEncryption/main/CryptMountConfigUtil.py
@@ -199,6 +199,38 @@ class CryptMountConfigUtil(object):
                 self.disk_util.umount(temp_mount_point)
                 self.disk_util.luks_close(crypt_item.mapper_name)
 
+    def clear_stale_nvme_disks(self):
+        fstab_items = []
+        for crypt_item in self.get_crypt_items():
+            if crypt_item.mapper_name.startswith(CommonVariables.nvme_disk_identifier):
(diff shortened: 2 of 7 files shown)


VMEncryption/main/CryptMountConfigUtil.py
```diff
@@ -199,6 +199,38 @@ def consolidate_azure_crypt_mount(self, passphrase_file):
...
+            if crypt_item.mapper_name.startswith(CommonVariables.nvme_disk_identifier):
+                if not os.path.exists(crypt_item.dev_path):
+                    fstab_items.append(CommonVariables.dev_mapper_root+crypt_item.mapper_name)
+                    self.remove_crypt_item(crypt_item)
+        
+        if len(fstab_items) == 0:
+            self.logger.log("No stale nvme disk to clear")
+            return
+        
+        shutil.copy2('/etc/fstab', '/etc/fstab.backup.' + str(str(uuid.uuid4())))
```
ejarvi, line 214: str twice?  str(str(...))

VMEncryption/main/handle.py
```diff
@@ -190,7 +190,8 @@ def stamp_disks_with_settings(items_to_encrypt, encryption_config):
         disk_util=disk_util,
         crypt_mount_config_util=crypt_mount_config_util)
 
-    settings.post_to_wireserver(data)
+    if len(data.get('Disks')) > 0:
```
ejarvi, line 193: should some error or warning be logged if if it isn't > 0 ?

VMEncryption/main/handle.py
```diff
@@ -1596,9 +1607,14 @@ def find_all_devices_to_encrypt(encryption_marker, disk_util, bek_util, volume_t
             if device_item.mount_point is None or device_item.mount_point == "":
                 # Don't encrypt partitions that are not even mounted
                 continue
-            if os.path.join('/dev/', device_item.name) not in dev_path_reference_table:
+            if os.path.join('/dev/', device_item.name) not in dev_path_reference_table and not device_item.name.startswith(CommonVariables.nvme_disk_identifier):
```
ejarvi, line 1610: risk seems small - but if user happens to have a non-nvme mapper name that also happens to start with "nvme" for some reason - this could misidentify, maybe something to call out in some documentation footnotes

VMEncryption/main/BekUtil.py
```diff
@@ -82,6 +83,8 @@ def get_bek_passphrase_file(self, encryption_config):
 
             for file in os.listdir(self.bek_filesystem_mount_point):
                 if bek_filename in file:
+                    # Make sure LinuxPassPhraseFileName exist for Resource and Nvme disks
```
vermashi, line 86: Should avoid self copy or unnecessary overwriting:
maybe add:
if bekfilename != file and not os.file.exists(bekfilename):

jeffhoward-revint:
The official documentation still states the NVMe disks are not supported with ADE on Linux.  Any reason why this was never merged into the the official release of the Linux extention?

PR #1097, opened 2020-05-06
vmaccess updates

D1v38om83r:
fixed code injection issue
fixed crash when ovf-env.xml is missing
removed references to platform.distro(), made code compatible with python 3.8
removed dependency on waagent file (added an alternative to HandlerUtil.py file that also doesn't need to import waagent file)

QOL improvements
extracted modules from waagent files
added unit tests for some modules


Utils/constants.py
```diff
@@ -0,0 +1,46 @@
...
+
+LBProbeResponder=y                      # Respond to load balancer probes if requested by Azure.
+
+Logs.Verbose=n                          # Enable verbose logs
+
+OS.RootDeviceScsiTimeout=300            # Root device timeout in seconds.
+OS.OpensslPath=None                     # If "None", the system default version is used.
+"""
+
+waagent_config_path = "/etc/waagent.conf"
```
narrieta, line 32: extensions shouldn't rely on waagent.conf, we should not expose it here so that we do not encourage extensions to take dependencies on it

D1v38om83r: move it inside vmaccess.py instead? We need it to read values like Provisioning.PasswordCryptSaltLength

narrieta: ok, that sounds good (making it local to vmaccess)

it is weird we have this dependency... does vmaccess depend on any other settings? (I'll add comments in the agent's code pointing out this dependency)

Utils/constants.py
```diff
@@ -0,0 +1,46 @@
...
+Openssl = "openssl"
+os_release = "/etc/os-release"
+
+
+class WALAEventOperation:
+    HeartBeat = "HeartBeat"
+    Provision = "Provision"
+    Install = "Install"
+    UnIsntall = "UnInstall"
```
pgombar, line 41: typo in operation name

Utils/constants.py
larohra: > move it inside vmaccess.py instead? We need it to read values like Provisioning.PasswordCryptSaltLength

Suggestion, why not only copy the flags that VMAccess needs instead of copying the whole thing. This would remove unnecessary dependencies.

narrieta: @larohra - the problem is that since this was exposed it is now not possible to know if anybody is depending on this. Say (just as an example) somebody created a custom image and set the conf to a specific length, which should be used both by the agent during provisioning and by vmaccess in subsequent password updates.

VMAccess/vmaccess.py
```diff
@@ -302,25 +322,44 @@ def _set_sshd_config(config, name, val):
...
+            return "debian_default"
+        if re.match("fedora", os_name, re.IGNORECASE):
+            return "fedora_default"
+        if re.match("red\s?hat", os_name, re.IGNORECASE):
+            return "redhat_default"
+        if re.match("suse", os_name, re.IGNORECASE):
+            return "SuSE_default"
+        if re.match("ubuntu", os_name, re.IGNORECASE):
+            return "ubuntu_default"
+        return "default"
```
pgombar, line 348: Curious how this list is compiled, is VMAccess not supported on other distros?
For example, the [`osutil` factory](https://github.com/Azure/WALinuxAgent/blob/develop/azurelinuxagent/common/osutil/factory.py) in WALinuxAgent has more distros not mentioned here like `arch`, `kali`, etc.

D1v38om83r: Any for distro not mentioned here, we don't ship a default ssh configuration file. We then default to the default-default ssh configuration

Utils/openssutils.py
```diff
@@ -0,0 +1,56 @@
...
+            c = 0
+            index = 7
+    return s
+
+
+def openssl_publickey_to_ssh(file):
+    """
+    Return base-64 encoded key appropriate for ssh.
+    """
+    from pyasn1.codec.der import decoder as der_decoder
```
D1v38om83r, line 39: Python 2.7 on Ubuntu 18 by default doesn't ship with this module. What to do?

Utils/constants.py
D1v38om83r: fixed

Utils/constants.py
D1v38om83r: I can only see 2 usages
"Provisioning.PasswordCryptId" and 
"Provisioning.PasswordCryptSaltLength"

Utils/openssutils.py
D1v38om83r: This also means that when the user didn't provide an ssh-rsa public key, the existing behavior is to try to extract a public key from the certificate present under /var/lib/waagent and add it to authorized_keys under $HOME/.ssh

The old behavior was to catch the exception

ext_utils.add_extension_event(name=hutil.get_name(),
                                          op=constants.WALAEventOperation.Enable,
                                          is_success=False,
                                          message="(02100)Failed to reset ssh key.")

and carry on, but the extension status itself would show succeeded.

Utils/constants.py
D1v38om83r: fixed in [commit](https://github.com/Azure/azure-linux-extensions/pull/1097/commits/89eabc1d5be906fefaa534d1bdcad423297a237e)

Utils/openssutils.py
D1v38om83r: Fixed old behavior in new [commit](https://github.com/Azure/azure-linux-extensions/pull/1097/commits/20aff780d34ba7e71c672692ddc18095366705dd)

but we still cannot convert non ssh-rsa public keys on python2.7

Utils/distroutils.py
```diff
@@ -0,0 +1,521 @@
...
+    def __init__(self, config):
+        """
+        Generic Attributes go here.  These are based on 'majority rules'.
+        This __init__() may be called or overriden by the child.
+        """
+        self.selinux = None
+        self.service_cmd = '/usr/sbin/service'
+        self.ssh_service_restart_option = 'restart'
+        self.ssh_service_name = 'ssh'
+        self.disro_name = 'default'
```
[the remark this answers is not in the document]
D1v38om83r: fixed

Utils/distroutils.py
```diff
@@ -0,0 +1,499 @@
+import os
```
boumenot, line 1: Am I correct in that you wrote this file from scratch?  Impressive.  There is a lot of code here to handle the various cases.  WALinuxAgent does most (all?) of what you want, and supports more distros and distro versions.  If possible, consider using that code instead.  Or, did you already do that and created this file from pieces of the agent?

How are you testing this code?

Utils/logger.py
```diff
@@ -0,0 +1,230 @@
...
+                    if sys.version_info[0] == 3:
+                        message = ''.join(list(message)).encode('ascii', 'ignore').decode("ascii", "ignore")
+                    elif sys.version_info[0] == 2:
+                        message = message.encode('ascii', 'ignore')
+
+                    F.write(message + "\n")
+            except IOError as e:
+                pass
+
+    def write_to_console(self, message):
```
boumenot, line 52: write_to_file and write_to_console are mostly the same method with two different parameters.

VMAccess/vmaccess.py
```diff
@@ -37,10 +38,58 @@
 OutputSplitter = ';'
 SshdConfigPath = '/etc/ssh/sshd_config'
 
+# overwrite the default logger
+logger.global_shared_context_logger = logger.Logger('/var/log/waagent.log', '/dev/stdout')
```
boumenot, line 42: Do you want to log to waagent.log?

VMAccess/vmaccess.py
```diff
@@ -157,24 +208,33 @@ def _remove_user_account(user_name, hutil):
...
-    ovf_xml = waagent.GetFileContents('/var/lib/waagent/ovf-env.xml')
-    ovf_env = waagent.OvfEnv().Parse(ovf_xml)
+    ovf_xml = None
+    ovf_env = None
+    try:
+        ovf_xml = ext_utils.get_file_contents('/var/lib/waagent/ovf-env.xml')
+        ovf_env = ovf_utils.OvfEnv.parse(ovf_xml, Configuration)
+    except (OSError, ValueError, KeyError, AttributeError):
+        pass
+    if ovf_xml is None or ovf_env is None:
```
boumenot, line 234: It should be sufficient to check if ovf_env is None.

VMAccess/manifest.xml
```diff
@@ -2,7 +2,7 @@
 <ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
   <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
   <Type>VMAccessForLinux</Type>
-  <Version>1.5.4</Version>
+  <Version>1.5.5</Version>
```
boumenot, line 5: This is a big change to release as a patch.  I don't know the current breakdown for extensions with autoUpgradeMinorVersion, but you can release as 1.6 now and verify it works then re-release as 1.5.5 if necessary.

VMAccess/vmaccess.py
```diff
@@ -37,10 +38,58 @@
...
+        return self.values.get(key)
+
+    def yes(self, key):
+        config_value = self.get(key)
+        if config_value is not None and config_value.lower().startswith("y"):
+            return True
+        else:
+            return False
+
+    def no(self, key):
```
boumenot, line 79: `None` is a valid value, and no() matches on it incorrectly.

VMAccess/vmaccess.py
```diff
@@ -194,106 +254,104 @@ def _set_user_account_pub_key(protect_settings, hutil):
...
+                    pub_path = ovf_env.prepare_dir(pub_path, MyDistro)
                     final_cert_txt = cert_txt
-                    if(not cert_txt.endswith("\n")):
-                        final_cert_txt = final_cert_txt+"\n"
-                    waagent.AppendFileContents(pub_path, final_cert_txt)
-                    waagent.MyDistro.setSelinuxContext(pub_path,
-                        'unconfined_u:object_r:ssh_home_t:s0')
-                    waagent.ChangeOwner(pub_path, user_name)
-                    waagent.AddExtensionEvent(name=hutil.get_name(), op="scenario", isSuccess=True, message="create-user")
+                    if not cert_txt.endswith("\n"):
```
boumenot, line 288: NIT: Save an if statement `final_cert_txt = cert_txt.rstrip() + "\n"`

Utils/distroutils.py
D1v38om83r: Most of the code is borrowed from WALinuxAgent, with a few tweaks like class initialization, refactoring, and ensuring python 3.6 and 3.8 compatibility, and making sure the subprocess is used with shell=False.

The other distros you mentioned inherited the functions that VMAccess uses form the AbstractDistro class. So there was no need to implement them here.

VMAccess/vmaccess.py
D1v38om83r: HandlerUtil later changes the logging file to extension specific. So I have named the object global_shared_context_logger.

VMAccess/manifest.xml
D1v38om83r: That is a great idea. I'll keep that in mind.

VMAccess/vmaccess.py
D1v38om83r: Its copied from waagent. I think I should preserve old behavior.

Utils/distroutils.py
```diff
@@ -0,0 +1,499 @@
...
+            return FedoraDistro(config)
+        if re.match("red\s?hat", os_name, re.IGNORECASE):
+            # Red Hat
+            return RedhatDistro(config)
+        if re.match("coreos", os_name, re.IGNORECASE):
+            # CoreOs
+            return CoreOSDistro(config)
+        if re.match("freebsd", os_name, re.IGNORECASE):
+            # FreeBSD
+            return FreeBSDDistro(config)
```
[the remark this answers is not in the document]
D1v38om83r: Wasn't sure if the previous check was enough.

Utils/distroutils.py
```diff
@@ -0,0 +1,499 @@
...
+import Utils.logger as logger
+import Utils.extensionutils as ext_utils
+import Utils.constants as constants
+
+
+def get_my_distro(config):
+    if 'FreeBSD' in platform.system():
+        return FreeBSDDistro(config)
+    os_name = ext_utils.get_line_starting_with("NAME", constants.os_release)
+    if os_name is not None:
```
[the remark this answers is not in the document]
D1v38om83r: AbstractDistro should take care of them. The waagent file would also use AbstractDistro from them

Utils/distroutils.py
```diff
@@ -0,0 +1,499 @@
...
+
+class RedhatDistro(GenericDistro):
+    """
+    Redhat Distro concrete class
+    Put Redhat specific behavior here...
+    """
+    def __init__(self, config):
+        super(RedhatDistro, self).__init__(config)
+        self.service_cmd = '/sbin/service'
+        self.ssh_service_restart_option = 'condrestart'
```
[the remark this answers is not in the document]
D1v38om83r: Refactored code form waagent file. Its existing behavior that is not a part for this PR.

Utils/distroutils.py
```diff
@@ -0,0 +1,499 @@
...
+        if re.match("freebsd", os_name, re.IGNORECASE):
+            # FreeBSD
+            return FreeBSDDistro(config)
+    return GenericDistro(config)
+
+
+# noinspection PyMethodMayBeStatic
+class GenericDistro(object):
+    """
+    GenericiDstro defines a skeleton neccesary for a concrete Distro class.
```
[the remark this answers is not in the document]
D1v38om83r: fixed

VMAccess/vmaccess.py
D1v38om83r: Being extra cautions doesn't hurt this case.

PR #1109, opened 2020-05-13
Support auto-config mode

abenbachir:
When customer install AMA with auto_config flag we don't set the LA and default because most 1p customer will use mdsdmgrctl to deploy their tenants.

But i still don't know if there is a scenario where we need to setup default mdsd with Auto-Config env variable. @canfikret any input on this ?

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -232,8 +232,11 @@ def install():
         #"customResourceId" : "/subscriptions/42e7aed6-f510-46a2-8597-a5fe2e15478b/resourcegroups/amcs-test/providers/Microsoft.OperationalInsights/workspaces/amcs-pretend-linuxVM",        
     }
 
-    # decide the mode
-    if protected_settings is None or len(protected_settings) is 0:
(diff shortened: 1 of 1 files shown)


nidhanda (dismissed):

AzureMonitorAgent/agent.py
```diff
@@ -369,7 +372,13 @@ def enable():
     """
     exit_if_vm_not_supported('Enable')
 
-    OneAgentEnableCommand = "systemctl start mdsd"
+    public_settings, protected_settings = get_settings()
+
+    if public_settings.get("AUTO_CONFIG") == "true":
+        OneAgentEnableCommand = "systemctl start mdsdmgr"
+    else:
+        OneAgentEnableCommand = "systemctl start mdsd"
```
canfikret, line 380: For the default tenant (/etc/default/mdsd), you need mdsd service running.

Has Nikhil confirmed to move /etc/default/mdsd to /etc/mdsd.d/tenants?

canfikret (dismissed):

simathih (approved):

PR #1112, opened 2020-05-18
Added compatibility for python 2.6 subprocess.check_output

D1v38om83r:
Fix for centos 2.6

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -367,21 +367,26 @@ def _set_sshd_config(config, name, val):
 
 
 def _get_default_ssh_config_filename():
-    os_name = ext_utils.get_line_starting_with("NAME", constants.os_release)
+    if os.path.isfile(constants.os_release):
+        os_name = ext_utils.get_line_starting_with("NAME", constants.os_release)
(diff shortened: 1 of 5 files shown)


VMAccess/vmaccess.py
```diff
@@ -367,21 +367,26 @@ def _set_sshd_config(config, name, val):
...
+        os_name = ext_utils.get_line_starting_with("NAME", constants.os_release)
+    elif os.path.isfile(constants.system_release):
+        os_name = ext_utils.get_file_contents(constants.system_release)
+    else:
+        return "default"
     if os_name is not None:
         # the default ssh config files are present in
         # /var/lib/waagent/Microsoft.OSTCExtensions.VMAccessForLinux-<version>/resources/
-        if re.match("centos", os_name, re.IGNORECASE):
+        if re.search("centos", os_name, re.IGNORECASE):
```
larohra, line 379: Suggestion: Shouldn't this also be a part of the DistroUtils instead?

D1v38om83r: This is for figuring out which default ssh_config to use, which is vmaccess specific code.

PR #1117, opened 2020-05-29
RHEL 7.8 support and QoS Fixes

vimish:
Attempt create directory only if it doesn't exist
Attempt mount an encrypted disk only if it is not already mounted
Handle redo last status if status file doesn't exist

### DIFF
--- a/VMEncryption/main/DiskUtil.py
+++ b/VMEncryption/main/DiskUtil.py
@@ -81,6 +81,9 @@ class DiskUtil(object):
         return self.command_executor.Execute(mkfs_cmd)
 
     def make_sure_path_exists(self, path):
+        if os.path.exists(path):
+            self.logger.log("Path {0} already exists.".format(path))
+            return 0
         mkdir_cmd = self.distro_patcher.mkdir_path + ' -p ' + path
         self.logger.log("make sure path exists, executing: {0}".format(mkdir_cmd))
         return self.command_executor.Execute(mkdir_cmd)
@@ -302,6 +305,12 @@ class DiskUtil(object):
 
     def mount_crypt_item(self, crypt_item, passphrase):
         self.logger.log("trying to mount the crypt item:" + str(crypt_item))
+        if crypt_item.mapper_name == CommonVariables.osmapper_name:
+            self.logger.log("Skipping OS disk.")
+            return
+        if self.is_device_mounted(crypt_item.mapper_name):
+            self.logger.log("Device {0} is already mounted".format(crypt_item.mapper_name))
+            return
         self.logger.log(msg=('First trying to auto mount for the item'))
         mount_filesystem_result = self.mount_auto(os.path.join(CommonVariables.dev_mapper_root, crypt_item.mapper_name))
         if str(crypt_item.mount_point) != 'None' and mount_filesystem_result != CommonVariables.process_success:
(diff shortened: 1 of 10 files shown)


VMEncryption/main/DiskUtil.py
```diff
@@ -302,6 +305,11 @@ def mount_filesystem(self, dev_path, mount_point, file_system=None):
 
     def mount_crypt_item(self, crypt_item, passphrase):
         self.logger.log("trying to mount the crypt item:" + str(crypt_item))
+        if crypt_item.mapper_name == CommonVariables.osmapper_name:
```
vermashi, line 308: log a quick message here?

VMEncryption/main/oscrypto/rhel_72_lvm/encryptstates/PatchBootSystemState.py
```diff
@@ -59,6 +59,9 @@ def enter(self):
 
         self.unmount_lvm_volumes()
         
+        # RHEL 7.8 does not seem to activate VG on close and open. 
+        # Hence, explicitly activating it.
+        self.command_executor.Execute('vgchange -a y rootvg')
```
ejarvi, line 64: Is this the first time vgchange has been introduced?  If so we should make sure that it and the syntax used is valid across all supported RHEL + CentOS distros (E2E testing should cover).

VMEncryption/test/test_utils.py
```diff
@@ -8,6 +8,7 @@ def __init__(self, name, version, kernel):
         self.distro_info[1] = version
         self.kernel_version = kernel
         self.mount_path = 'mount'
+        self.mkdir_path = 'mkdir'
```
ejarvi, line 11: Just a comment - As unit tests are moved under tox so we can test both python2 + python3 simultaneously, and with the goal of being able to run all tests within a container as part of build (not running in a pre-prepped Azure VM with a running guest agent or specific file system structure) - we'll need to mock everything and run as low privilege user.  Running mount and mkdir within the unit test itself push the boundaries of a unit test.  Not sure it has to be changed here, but is a a consideration that is going to be coming up moving forward.

VMEncryption/main/DiskUtil.py
vimish: Will do.

VMEncryption/main/oscrypto/rhel_72_lvm/encryptstates/PatchBootSystemState.py
vimish: I believe vgchange is present for a while now. So far we didn't have to use it as luksClose and luksOpen seems to have been activating the volume group. This does not seem to be the case with RHEL 7.8. Hence, have to explicitly invoke vgchange to activate it. I tried the command on already activated VG and it didn't fail. Hence, thought it should be safe for older distros. But, I will definitely confirm it by running it through our AllDistros test.

VMEncryption/test/test_utils.py
vimish: These are just dummy values. mount and mkdir will not actually be executed during unittest. It has be added as code expects these property to be present in DistroPatcher.
Sample here: https://github.com/Azure/azure-linux-extensions/blob/ade-singlepass-dev/VMEncryption/main/DiskUtil.py#L84
The actual execution of mount and mkdir is mocked in unittest.

VMEncryption/main/Utils/HandlerUtil.py
```diff
@@ -418,6 +418,13 @@ def redo_last_status(self):
             previous_seq = str(latest_sequence_num - 1)
             previous_status_file = os.path.join(self._context._status_dir, previous_seq + '.status')
 
+            if not os.path.exists(previous_status_file):
```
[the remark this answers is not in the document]
vimish: Not sure about the scenarios here. The guest agent telemetry just showed redo last status is failing because status file was not found. The scenario you described could also be a reason. Should I change the implementation here to exit with success instead of failure?

VMEncryption/main/DiskUtil.py
vimish: Done.

PR #1129, opened 2020-06-11
Move Resource Disk Encryption to main process.

vimish:


VMEncryption/main/handle.py
```diff
@@ -882,12 +882,40 @@ def enable_encryption():
...
+                        else:
+                            logger.log("Resource disk encrypted successfully.")
+
                 encryption_marker = mark_encryption(command=extension_parameter.command,
                                                     volume_type=extension_parameter.VolumeType,
                                                     disk_format_query=extension_parameter.DiskFormatQuery)
                 start_daemon('EnableEncryption')
 
     except Exception as e:
+        hutil.save_seq()
```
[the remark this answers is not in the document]
vimish: It kind of ensures if an operation failed it won't be retried on every guest agent restart. It helps throwing same error multiple times. If an enable failed then user will have explicitly send another enable command to retry the operation.

VMEncryption/test/test_resource_disk_util.py
```diff
@@ -211,3 +211,24 @@ def test_automount(self, mock_try_remount, mock_encrypt_format_mount, mock_is_pl
         mock_encrypt_format_mount.return_value = False
         self.assertEqual(self.resource_disk.automount(), False)
         mock_encrypt_format_mount.assert_called_once()
+
+    @mock.patch('main.ResourceDiskUtil.ResourceDiskUtil._is_crypt_mounted')
+    @mock.patch('main.ResourceDiskUtil.ResourceDiskUtil.encrypt_format_mount')
+    def test_encrypt_resource_disk(self, mock_efm, mock_icm):
```
ejarvi, line 217: Just a heads up these tests may need to be updated in the future so they can run independently without reliance on presence of disks, folders, or current user privileges.  For example if they result in a call to "mount" which requires root privileges.

VMEncryption/main/patch/redhatPatching.py
```diff
@@ -95,7 +95,11 @@ def install_cryptsetup(self):
             packages.remove('cryptsetup')
             
         if self.command_executor.Execute("rpm -q " + " ".join(packages)):
-            self.command_executor.Execute("yum install -y " + " ".join(packages))
+            return_code = self.command_executor.Execute("yum install -y " + " ".join(packages), timeout=30)
+            if return_code == -9:
+                msg = "Command: yum install timed out. Make sure yum is configured correctly and there are no network problems."
```
ejarvi, line 100: RHEL BYOS images must be registered before yum can be used.  Not sure if that warrants a mention here, it would probably fail with a different exit code in that case.

vimish: This exit code only captures timeout. As of now we do not capture other failures during package installation. I will try and address the broader issue of failing early if dependent packages are not installed in a different PR.

VMEncryption/test/test_resource_disk_util.py
vimish: Thanks for the heads up Eric. This particular test does not run any privileged command. I just mocked return values of two functions.

PR #1135, opened 2020-06-21
AMA metrics support

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -395,14 +415,192 @@ def update():
...
+def start_metrics_process():
+    """
+    Start telemetry process that performs periodic monitoring activities
+    :return: None
+
+    """
+    stop_metrics_process()
 
+    package_directory = os.path.join(os.getcwd(), PackagesDirectory)
+    telegraf_path = os.path.join(package_directory, TelegrafBinName)
```
nidhanda, line 465: Telegraf will also be part of the latest mdsd deb/rpm and it will copy it to /usr/sbin/telgraf as it does with mdsd.

AzureMonitorAgent/agent.py
```diff
@@ -395,14 +415,192 @@ def update():
...
+
+                        json_data = json.loads(data)  
+                        
+                        telegraf_config, telegraf_namespaces = telhandler.handle_config(
+                            json_data, 
+                            "udp://127.0.0.1:" + metrics_constants.ama_metrics_extension_udp_port, 
+                            "unix:///var/run/mdsd/default_influx.socket",
+                            False)
+                        
+                        me_handler.setup_me(False)
```
nidhanda, line 556: Nit - Add is_lad=False  just to keep uniformity with other such calls

AzureMonitorAgent/agent.py
simathih: as discussed, its not there in the buidl yet, so we will remove this once its done

PR #1136, opened 2020-06-22
Billing changes for only os disk backup scenario

srbansMS:
Adding the used space of only mount point '/' in case only os disk backup is enabled


VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -16,12 +16,24 @@
...
         self.logger=logger
         self.file_systems_info = []
         self.non_physical_file_systems = ['fuse', 'nfs', 'cifs', 'overlay', 'aufs', 'lustre', 'secfs2', 'zfs', 'btrfs', 'iso']
         self.known_fs = ['ext3', 'ext4', 'jfs', 'xfs', 'reiserfs', 'devtmpfs', 'tmpfs', 'rootfs', 'fuse', 'nfs', 'cifs', 'overlay', 'aufs', 'lustre', 'secfs2', 'zfs', 'btrfs', 'iso']
+        self.isOnlyOSDiskBackupEnabled = False
+        try:
+            if(para_parser.customSettings != None and para_parser.customSettings != ''):
+                self.logger.log('customSettings : ' + str(para_parser.customSettings))
+                customSettings = json.loads(para_parser.customSettings)
+                if("isOnlyOSDiskBackupEnabled" in customSettings.keys()):
```
soumyapattnaik, line 30: see if key is present can be used

srbansMS: Couldn't find a better way to check if key is present without iterating the list. There are some 5 custom settings so I think we can keep it this way or can you please let me know the method to use.

PR #1150, opened 2020-07-02
Generate msi token only when using AMCS

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -508,41 +508,41 @@ def metrics_watcher(hutil_error, hutil_log):
...
-                else:
-                    hutil_error(log_messages)
-
             if os.path.isfile(MdsdCounterJsonPath):
                 f = open(MdsdCounterJsonPath, "r")
                 data = f.read()
                 
-                if (data != ''):
-                    crc = hashlib.sha256(data).hexdigest()
+                if (data != ''):
```
nidhanda, line 515: We should add this check to Enable method as well where we call Start_metrics() .

simathih: the problem is you don't know when the AMCS config will be added,  but you need the metrics watcher up

PR #1155, opened 2020-07-08
AMA arc support

nidhanda:
Adding support for using the hybrid meta data service for Arc for getting metadata as well as MSI token in AMA


LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -148,10 +168,29 @@ def generate_MSI_token():
...
         data = None
         while retries <= max_retries:
-            msiauthurl = "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://ingestion.monitor.azure.com/"
-            req = urllib2.Request(msiauthurl, headers={'Metadata':'true', 'Content-Type':'application/json'})
-            res = urllib2.urlopen(req)
-            data = json.loads(res.read())
+            arc_endpoint = get_arc_endpoint()
+            try:
+                msiauthurl = arc_endpoint + "/metadata/identity/oauth2/token?api-version=2019-11-01&resource=https://management.azure.com/"
+                req = urllib2.Request(msiauthurl, headers={'Metadata':'true'})
```
simathih, line 174: If it’s expected to fail without token, then why make this call without token

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -49,6 +49,27 @@ def is_systemd():
     check_systemd = os.system("pidof systemd 1>/dev/null 2>&1")
     return check_systemd == 0
 
+def is_arc_installed():
```
simathih, line 52: This and functions below are duplicated

nidhanda: I wrote separate functions for both telegraf and ME files. Would you rather us write them once and import one file into another to use? I feel like that would introduce more complication?

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
nidhanda: This is by design from the arc team. The tokens don't exist by default, we only get them after the first call fails, then we use that token to again make the call with auth.

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
simathih: Can some of these common to a common library?

nidhanda: Added them to a common utils lib

PR #1172, opened 2020-07-24
Adding support for non systemd distros and fixing python set bug

nidhanda:
1) VMs with CentOS 6, RHEL 6, Oracle 6 don't have systemctl and thus are unable to run mdsd service. Added support to run mdsd binary as a process and storing the pid in a file to later kill the process during disable. 

2) Switched to set([...]) instead of {}  as py 2.6 doesnt support the later simpler way of set instantiation.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -208,15 +208,7 @@ def install():
     """
     find_package_manager("Install")
     exit_if_vm_not_supported('Install')
-
-    public_settings, protected_settings = get_settings()
-    
-    # if public_settings is None:
-    #     raise ParameterMissingException('Public configuration must be ' \
-    #                                     'provided')
-    #public_settings.get('workspaceId')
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -391,43 +349,58 @@ def enable():
         else:
             log_and_exit("install", MissingorInvalidParameterErrorCode, "Could not find the file - /etc/default/mdsd" )        
     except:
-        log_and_exit("install", MissingorInvalidParameterErrorCode, "Failed to add MCS Environment Variables in /etc/default/mdsd" ) 
+        log_and_exit("install", MissingorInvalidParameterErrorCode, "Failed to add MCS Environment Variables in /etc/default/mdsd" )        
+    return exit_code, output
+
+def check_kill_process(pstring):
+    for line in os.popen("ps ax | grep " + pstring + " | grep -v grep"):
```
simathih: -ax?

PR #1174, opened 2020-07-25
oracle workload fixes

soumyapattnaik:


bpramod (approved):

VMBackup/main/workloadPatch/WorkloadPatch.py
```diff
@@ -179,14 +185,15 @@ def postMaster(self):
         
         if 'mysql' in self.name.lower():
             self.logger.log("WorkloadPatch: Create connection string for post master")
-            postscript = os.path.join(os.getcwd(), "main/workloadPatch/scripts/postMysqlMaster.sql")
+            postscript = os.path.join(os.getcwd(), "main/workloadPatch/"+self.scriptpath+"/postMysqlMaster.sql")
```
suana-ms, line 188: Can we have a single function to return locations for pre/Post scripts for both the databases? Looks like these are being used at multiple places. We can standardize the Pre/Post script names , something like : "post" + DBName + "Constant".

PR #1175, opened 2020-07-27
`sshd_config` Backup Permission Preservation

progmem:
The following PR proposes a change to how backups of `sshd_config` are performed. Specifically, the following is being done:

* `copy2` is being leveraged in order to preserve original ownership and file metadata.
* An additional `chown` is being performed to make sure that the original owner and group are preserved.

The method `_backup_sshd_config` is called in a couple of cases: resetting the SSH configuration, or changing any user's password. In particular, this was being triggered while updating a user via the Azure CLI (`az vm user update ...`). The way current backups are being performed with `copyfile` does preserve the _contents_ of the configuration, however it doesn't preserve any specific permissions made to the configuration. In the cases of some OS configurations (e.g. CentOS 7) where the configuration is "for `root`'s eyes only" (permission set `600`), these configuration backups lose this permission set, and given that `/etc/ssh` can be accessed by non-root users this causes the configuration backups to be accessible by non-root users as well.

### DIFF
--- a/VMAccess/manifest.xml
+++ b/VMAccess/manifest.xml
@@ -2,7 +2,7 @@
 <ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
   <ProviderNameSpace>Microsoft.OSTCExtensions</ProviderNameSpace>
   <Type>VMAccessForLinux</Type>
-  <Version>1.5.8</Version>
+  <Version>1.5.9</Version>
   <Label>Microsoft Azure VM Access Extension for Linux Virtual Machines</Label>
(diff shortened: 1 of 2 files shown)


D1v38om83r:
LGTM

D1v38om83r:
Please also edit line 5 in https://github.com/Azure/azure-linux-extensions/blob/master/VMAccess/manifest.xml to  **<Version>1.5.9</Version>**

progmem:
@D1v38om83r Version bumped, should be good to go now!

D1v38om83r:
Approved

PR #1187, opened 2020-08-12
Add network connectivity info to exit code 53

aswatt:
When omsadmin.sh returns 8 (curl error in onboarding), error could be due to either incorrect workspace key or an issue connecting to the `<wsid>.oms.opinsights.azure.com` URL. Adding this info to output.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -1206,7 +1206,7 @@ def run_command_and_log(cmd, check_error = True, log_cmd = True):
                 # 53 is the exit code for configuration errors
                 # https://github.com/Azure/azure-marketplace/wiki/Extension-Build-Notes-Best-Practices#error-codes-and-messages-output-to-stderr
                 exit_code = 53
-                output = "Installation failed due to incorrect workspace key. Please check if the workspace key is correct. For details, check logs in /var/log/azure/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux/extension.log"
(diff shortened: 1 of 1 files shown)


aswatt:
Joint with this PR in OMS-Agent-for-Linux: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1234

OmsAgent/omsagent.py
```diff
@@ -1206,7 +1206,7 @@ def run_command_and_log(cmd, check_error = True, log_cmd = True):
                 # 53 is the exit code for configuration errors
                 # https://github.com/Azure/azure-marketplace/wiki/Extension-Build-Notes-Best-Practices#error-codes-and-messages-output-to-stderr
                 exit_code = 53
-                output = "Installation failed due to incorrect workspace key. Please check if the workspace key is correct. For details, check logs in /var/log/azure/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux/extension.log"
+                output = "Installation failed due to either incorrect workspace key or networking error. Please verify the connectivity to the url and the correctness of the workspace key. For details, check logs in /var/log/azure/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux/extension.log"
```
hestolz, line 1209: It won't be clear to what URL the customer should be checking connectivity for, this needs to be specified. If we don't know exactly what the URL will be in extension (e.g. this code path could be hit for multiple instances of "check the correctness of the workspace ID" each with a different problem URL), we should ensure shell bundle specifies the URL in question and parse that out here in `omsagent.py`.

simathih: alternative, shell bundle could output it to STDOUT and the above message will include it automatically

simathih (approved):

PR #1191, opened 2020-08-18
[OneZip] Detect cloud region and apply domain dynamically at Install time

abenbachir:


OmsAgent/omsagent.py
```diff
@@ -692,6 +735,51 @@ def get_vmresourceid_from_metadata():
...
+            for region in regions:
+                regionCompacted = region.replace(" ", "")
+                if regionCompacted.lower() == location.lower():
+                    hutil_log_info("Detecting cloud type '%s' by using the current location '%s' retrieved from IMDS. The domain '%s' will be used." % (cloudName, location, domain))
+                    return domain
+
+        defaultDomain = CloudRegionMap[DefaultCloudName]['Domain']
+        hutil_log_info("Warning: We couldn't find a cloud type associated with the current location '%s', fallback to '%s' domain '%s'" % (location, DefaultCloudName, defaultDomain))
+        return defaultDomain
+    except Exception as e:
```
hestolz, line 778: In this case we should also fall back to `DefaultCloudName`

abenbachir: that what I did

hestolz: I'm talking about the exception. If an exception is thrown, we just return None.

OmsAgent/omsagent.py
```diff
@@ -692,6 +735,51 @@ def get_vmresourceid_from_metadata():
...
+def get_azure_cloud_domain():
+    try:
+        location = get_azure_region_from_imds()
+        # look for the cloud type associated with current location
+        for cloudName, data in CloudRegionMap.items():
+            domain = data['Domain']
+            regions = data['Regions']
+            for region in regions:
+                regionCompacted = region.replace(" ", "")
+                if regionCompacted.lower() == location.lower():
```
hestolz, line 771: `get_azure_region_from_imds()` can return `None`, in which case `location.lower()` will throw exception. We catch it, but better to handle `None` to start with IMO ...

PR #1193, opened 2020-09-01
Adding support for AMA on Arc

nidhanda:
Adding extension support for AMA on Arc. Will be running a separate arc daemon to keep track of the valid key/token files for making the MSI call to Arc's hybrid service. 

**Work remaining -** Adding retry logic for arc watcher and making it more resilient.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -48,6 +48,7 @@ from threading import Thread
 import telegraf_utils.telegraf_config_handler as telhandler
 import metrics_ext_utils.metrics_constants as metrics_constants
 import metrics_ext_utils.metrics_ext_handler as me_handler
+import metrics_ext_utils.metrics_common_utils as metrics_utils
 
 try:
     from Utils.WAAgentUtil import waagent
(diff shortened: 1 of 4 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -668,16 +680,122 @@ def metrics():
...
+    """
+    This is needed to override mdsd's syslog permissions restriction which prevents mdsd 
+    from reading temporary key files that are needed to make https calls to get an MSI token for arc during onboarding to download amcs config
+    This method spins up a process that will continuously keep refreshing that particular file path with valid keys
+    So that whenever mdsd needs to refresh it's MSI token, it is able to find correct keys there to make the https calls
+    """
+    # check every 25 seconds
+    sleepTime =  25
+
+    # sleep before starting the monitoring.
```
simathih, line 760: verify this

simathih (approved):

AzureMonitorAgent/agent.py
nidhanda: Is this for that whole verification on RHEL or just this specific code block?

simathih: both :)

nidhanda: Tested both on RHEL. Works as expected

PR #1198, opened 2020-09-09
Modified python call order and added support for python executable not in env.

abhishekgoyal-microsoft:
Modified python call order and added support for python executable not in env. Also added support for fallback to previous python versions if script run fails.


VMBackup/main/handle.sh
```diff
@@ -30,59 +30,44 @@ then
...
-    echo "`date`- python2.6 path exists" >> $logfile
-    /usr/bin/python2.6 main/handle.py -$1
-    rc=$?
-elif [ -f "/usr/bin/python" ]
-then
-    echo "`date`- python path exists" >> $logfile
-    /usr/bin/python main/handle.py -$1
-    rc=$?
-elif [ -f "`which python`" ]
+pythonVersionList="python3.6 python3.5 python3.4 python3.3 python3 python2.7 python2.6 python2 python"
```
soumyapattnaik, line 33: add 3.8 and 3.7 as well

VMBackup/main/handle.sh
```diff
@@ -30,59 +30,44 @@ then
...
+
+for i in ${pythonVersionList};
```
soumyapattnaik, line 35: use meaningful variable

VMBackup/main/handle.sh
```diff
@@ -30,59 +30,44 @@ then
...
-    /usr/bin/env python main/handle.py -$1
-    rc=$?
-else
-    echo "`date`- python version unknown" >> $logfile
+	if [ -f "`which python`" ]
+	then
+		echo "`date`- python path exists" >> $logfile
+		/usr/bin/env python main/handle.py -$1
+		rc=$?
+	elif [ -f "${pythonPath}" ]
```
soumyapattnaik, line 61: add check if which python fails

soumyapattnaik (approved):

PR #1199, opened 2020-09-09
Made VMAccessLinux more robust to the absence of ovf-env.xml

D1v38om83r:


Utils/ovfutils.py
```diff
@@ -88,58 +88,64 @@ def parse(xml_text, configuration, is_deprovision=False):
         Parse xml tree, retrieving user and ssh key information.
         Return self.
         """
-        ofv_env = OvfEnv()
+        ovf_env = OvfEnv()
+        if xml_text is None:
+            return None
         logger.log_if_verbose(re.sub("<UserPassword>.*?<", "<UserPassword>*<", xml_text))
```
nkuchta, line 94: Does this work if the xml string has a namespace in the element? e.g., "<ns1:UserPassword>..."

D1v38om83r: No it wouldn't work for that. Updated the logic in a new commit

nkuchta (approved):

PR #1201, opened 2020-09-16
Update IMDS Endpoint to work for Arc machines

aswatt:
Currently OMS will attempt to query the IMDS endpoint on Arc machines, timing out after 2+ minutes. This change checks if the machine is an Arc machine, and then if so, instead queries the Arc IMDS endpoint.


nidhanda (dismissed):

hestolz:
@aswatt has this been tested on an Arc VM?

hestolz (approved):

PR #1225, opened 2020-10-14
Adding Resource disk validation for encryptFormatAll

ShridharBhatGokarna:
This PR includes:

Verification of Resource disk existence prior enabling encryption on resource disk when using EncryptFormatAll. The encryptFormatAll operation used to fail as there was no check to verify whether resource disk exists before the encryption operation.

### DIFF
--- a/VMEncryption/main/ResourceDiskUtil.py
+++ b/VMEncryption/main/ResourceDiskUtil.py
@@ -61,6 +61,11 @@ class ResourceDiskUtil(object):
         cmd = 'cryptsetup isLuks ' + self.RD_DEV_PATH
         return (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success
 
+    def _resource_disk_exists(self):
+        """ true if udev name for resource disk exists """
+        cmd = 'test -b ' + self.RD_BASE_DEV_PATH
+        return (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success
+
     def _resource_disk_partition_exists(self):
         """ true if udev name for resource disk partition exists """
         cmd = 'test -b ' + self.RD_DEV_PATH
--- a/VMEncryption/main/ResourceDiskUtil.py
+++ b/VMEncryption/main/ResourceDiskUtil.py
@@ -64,7 +64,11 @@ class ResourceDiskUtil(object):
     def _resource_disk_exists(self):
         """ true if udev name for resource disk exists """
         cmd = 'test -b ' + self.RD_BASE_DEV_PATH
-        return (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success
+        if (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success:
+            self.logger.log("Resource disk exists.")
+            return True
+        self.logger.log("Resource disk does not exist.")
+        return False
 
     def _resource_disk_partition_exists(self):
         """ true if udev name for resource disk partition exists """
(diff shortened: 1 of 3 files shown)


vermashi (commented):
Add a unit test. It should be easy to mock the output of "_resource_disk_exists()" and make sure that encyrption is skipped

VMEncryption/main/handle.py
```diff
@@ -918,16 +918,19 @@ def enable_encryption():
...
                         resource_disk_util = ResourceDiskUtil(logger, disk_util, crypt_mount_config_util, passphrase_file, get_public_settings(), DistroPatcher.distro_info)
-                        rd_encrypted = resource_disk_util.encrypt_resource_disk()
-                        if not rd_encrypted:
-                            hutil.save_seq()
-                            hutil.do_exit(exit_code=CommonVariables.configuration_error,
-                                          operation='EnableEncryption',
-                                          status=CommonVariables.extension_error_status,
-                                          code=str(CommonVariables.configuration_error),
-                                          message='Failed to encrypt resource disk. Please make sure no process is using it.')
+                        if resource_disk_util._resource_disk_exists():
```
vimish, line 921: Resource disk encryption is also called at this location:
https://github.com/Azure/azure-linux-extensions/blob/ade-singlepass-dev/VMEncryption/main/handle.py#L921
Similar check needs to be added there.

Also, I will suggest to move the check to encrypt_format_mount method (https://github.com/Azure/azure-linux-extensions/blob/ade-singlepass-dev/VMEncryption/main/ResourceDiskUtil.py#L282)
The first check here should be if Resource disk exists or not. If it does not exist return True.

vimish (changes requested):

VMEncryption/main/handle.py
ejarvi: +1 to Vimish's comments

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -61,6 +61,11 @@ def _is_luks_device(self):
         cmd = 'cryptsetup isLuks ' + self.RD_DEV_PATH
         return (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success
 
+    def _resource_disk_exists(self):
```
ejarvi, line 64: unit tests for this would be helpful

ejarvi (changes requested):

VMEncryption/main/handle.py
vermashi: In that case I'd say move the logging lines to the ResourceDiskUtil as well

vermashi: moreover, you might need to add the "resource_disk_exists" check to resource disk util's automount method.

vermashi: Finally moving it all to RDUtil will make writing unit tests a lot easier as that file has a test-file already set up for it

ShridharBhatGokarna: > moreover, you might need to add the "resource_disk_exists" check to resource disk util's automount method.

The automount method when invoked without resource disk will return false and won't throw exception. Hence I did not add the check here as the existing code suffices this.

vermashi: > The automount method when invoked without resource disk will return false and won't throw exception. Hence I did not add the check here as the existing code suffices this.

Cool. Could you add a test case for automount to cover that code path?

vimish: Check for this line in handle.py **rd_encrypted = resource_disk_util.encrypt_resource_disk()**
This will be called when encrypt format all is set while invoking the extension for the first time

vimish: Sorry...please disregard my comment. I got confused with automount error.

But I still want to either have this check for automount also or move it to encrypt_format_mount.
Even though we don't throw any exception in automount failure we will still go ahead and modify waagent configuration before failing.

VMEncryption/main/ResourceDiskUtil.py
ShridharBhatGokarna: Unit tests have been added

VMEncryption/main/handle.py
ShridharBhatGokarna: > Finally moving it all to RDUtil will make writing unit tests a lot easier as that file has a test-file already set up for it

I agree with you on this. Changes have been made for this.

VMEncryption/main/ResourceDiskUtil.py
ShridharBhatGokarna: Unit tests have been added

VMEncryption/main/handle.py
ShridharBhatGokarna: > moreover, you might need to add the "resource_disk_exists" check to resource disk util's automount method.

this check has been added at the beginning of encrypt_format_mount() since this function calls the prepare function for resource disk

VMEncryption/main/handle.py
```diff
@@ -927,8 +927,7 @@ def enable_encryption():
                                           code=str(CommonVariables.configuration_error),
                                           message='Failed to encrypt resource disk. Please make sure no process is using it.')
                         else:
-                            logger.log("Resource disk encrypted successfully.")
-
+                            logger.log("Resource disk is not present or encrypted successfully.")
```
vermashi, line 930: Ambiguous sentence.
Change to "Resource Disk is either absent or encrypted successfully."

vermashi (approved):

PR #1249, opened 2020-11-06
AMA python2/3 interoperability 

nidhanda:
1) Followed the steps listed by Henry in his OMS py2/3 PR - #1012 . Initially followed the [guide ](https://sebastianraschka.com/Articles/2014_python_2_3_key_diff.html#integer-division) and using the [futurize tool](https://python-future.org/automatic_conversion.html).  
2) Manually fixed quite a few issues that couldn't be ported with automation. 
3) Re-used OMSAgent's shim file for interoping with py2/3. Currently we are not installing python if it isnt there. 
4) Added python-future as a submodule and using it in the shim file during agent.py's execution
5) Also added the continueOnUpdateFailure flag in Handlermanifest.json file to make the extension installation process more stable (that flag allows waagent to override faulty installations when they're stuck in update-disable loop and can't be updated or removed)
6) Added the subprocess monkey-patch #1114 from OMSAgent's fix
7) Replaced HutilObject.log with hutil_log_info method that has additional None check 
8) Added future submodule in extension packaging logic

Creating the PR now to make the review process easy/streamlined

Things left to do - 
- Test on all the distros in our support matrix and add them in the code as supported once individually verified. 
- Test with diff py versions on RHEL/Ubuntu

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -17,11 +17,16 @@
 # limitations under the License.
 
 from __future__ import print_function
-from __future__ import division
-from future import standard_library
-standard_library.install_aliases()
-from builtins import str
-from past.utils import old_div
+import sys
+# future imports have no effect on python 3 (verified in official docs)
+# importing from source causes import errors on python 3, lets skip import
+if sys.version_info[0] < 3:
+    from __future__ import division
+    from future import standard_library
+    standard_library.install_aliases()
+    from builtins import str
+    from past.utils import old_div
+
 import os
 import os.path
 import datetime
@@ -31,7 +36,6 @@ import grp
 import re
 import filecmp
 import stat
-import sys
 import traceback
 import time
 import platform
@@ -197,7 +201,7 @@ def get_free_space_mb(dirname):
     Get the free space in MB in the directory path.
     """
     st = os.statvfs(dirname)
-    return old_div(old_div(st.f_bavail * st.f_frsize, 1024), 1024)
+    return (st.f_bavail * st.f_frsize) // (1024 * 1024)
 
 
 def is_systemd():
@@ -518,7 +522,7 @@ def start_metrics_process():
     
     #start telemetry watcher
     oneagent_filepath = os.path.join(os.getcwd(),'agent.py')
-    args = ['python', oneagent_filepath, '-metrics']
+    args = ['python{0}'.format(sys.version_info[0]), oneagent_filepath, '-metrics']
     log = open(os.path.join(os.getcwd(), 'daemon.log'), 'w')
     HUtilObject.log('start watcher process '+str(args))
     subprocess.Popen(args, stdout=log, stderr=log)
@@ -714,7 +718,7 @@ def start_arc_process():
     
     #start telemetry watcher
     oneagent_filepath = os.path.join(os.getcwd(),'agent.py')
-    args = ['python', oneagent_filepath, '-arc']
+    args = ['python{0}'.format(sys.version_info[0]), oneagent_filepath, '-arc']
     log = open(os.path.join(os.getcwd(), 'daemon.log'), 'w')
     HUtilObject.log('start watcher process '+str(args))
     subprocess.Popen(args, stdout=log, stderr=log)
@@ -901,9 +905,9 @@ def is_vm_supported_for_extension(operation):
     The supported distros of the AzureMonitorLinuxAgent are allowed to utilize
     this VM extension. All other distros will get error code 51
     """
-    supported_dists = {'redhat' : ['6', '7'], # CentOS
+    supported_dists = {'redhat' : ['6', '7', '8'], # Rhel
                        'centos' : ['6', '7'], # CentOS
-                       'red hat' : ['6', '7'], # Oracle, RHEL
+                       'red hat' : ['6', '7', '8'], # Oracle, RHEL
                        'oracle' : ['6', '7'], # Oracle
                        'debian' : ['8', '9'], # Debian
                        'ubuntu' : ['14.04', '16.04', '18.04'], # Ubuntu
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -905,6 +905,7 @@ def find_vm_distro(operation):
     Finds the Linux Distribution this vm is running on. 
     """
     vm_dist = vm_id = vm_ver =  None
+    parse_manually = False
     try:
         vm_dist, vm_ver, vm_id = platform.linux_distribution()
     except AttributeError:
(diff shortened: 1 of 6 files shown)


AzureMonitorAgent/shim.sh
```diff
@@ -0,0 +1,36 @@
...
+        eval ${python_exec_command}="python3"
+        # do not set future_path; future seems to cause interference with preexisting packages in python 3 environment
+    fi
+}
+
+find_python PYTHON FUTURE_PATH
+
+if [ -z "$PYTHON" ] # If python is not installed, we will fail the install with the following error, requiring cx to have python pre-installed
+then
+    echo "No Python interpreter found, which is an AMA extension dependency. Please install either Python 2 or 3." >&2
```
hestolz, line 29: My understanding is that eventually we want to be force-installing python3 on the system if we don't find any python. Is that scoped to a separate PR/release?

nidhanda: Currently we are inclining towards going with the non-intrusive approach of failing to install if python isn't present on the VM (similar to OMS) rather than force installing python. Need to get PM approval but seems like this current design will likely not change.

hestolz: Gotcha. I was also under the impression that we wanted to prefer python3; in that case `find_python` needs to reflect this since currently it will take python2 if both are available.

And if my impression is correct, maybe the wording on this line could be more targed—for instance, "Please install Python 3, or Python 2 if the former is unavailable."

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -561,8 +569,16 @@ def setup_me(is_lad):
     aad_auth_url = ""
     amrurl = "https://management.azure.com/subscriptions/" + subscription_id + "?api-version=2014-04-01"
     try:
-        req = urllib2.Request(amrurl, headers={'Content-Type':'application/json'})
-        res = urllib2.urlopen(req)
+        req = urllib.request.Request(amrurl, headers={'Content-Type':'application/json'})
+
+        # urlopen alias in future backport is broken on py2.6, fails on urls with HTTPS - https://github.com/PythonCharmers/python-future/issues/167
```
hestolz, line 574: This might be a bit of a gotcha in the future; if someone else has to add another call over HTTPS and isn't aware of this, they'll reintroduce the bug. Is there any easy fix?

AzureMonitorAgent/agent.py
```diff
@@ -1330,7 +1388,7 @@ def log_and_exit(operation, exit_code = 1, message = ''):
     """
     Log the exit message and perform the exit
     """
-    if exit_code is 0:
+    if exit_code == 0:
```
hestolz, line 1391: I can't comment on it bc. there is no diff, but in the `is_vm_supported_for_extension` function there is an instance of `is not` between two ints, which should probably also take this change. Might be worth auditing for more instances of bad use of `is [not]` in the old functions taken from OMS

hestolz (changes requested):

AzureMonitorAgent/agent.py
```diff
@@ -25,16 +34,15 @@
...
 import time
 import platform
 import subprocess
 import json
 import base64
 import inspect
-import urllib
-import urllib2
+import urllib.request, urllib.parse, urllib.error
+import urllib.request, urllib.error, urllib.parse
```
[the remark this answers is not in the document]
nidhanda: Fixed

AzureMonitorAgent/agent.py
```diff
@@ -57,12 +65,44 @@
...
+    class CalledProcessError(Exception):
+        def __init__(self, returncode, cmd, output=None):
+            self.returncode = returncode
+            self.cmd = cmd
+            self.output = output
+
+        def __str__(self):
+            return "Command '%s' returned non-zero exit status %d" % (self.cmd, self.returncode)
+
+    subprocess.check_output = check_output
```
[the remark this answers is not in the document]
nidhanda: fixed

AzureMonitorAgent/agent.py
nidhanda: This was the only instance in this file. Fixed now

AzureMonitorAgent/agent.py
```diff
@@ -879,7 +932,6 @@ def find_vm_distro(operation):
                         vm_dist = vm_dist.replace('\"', '').replace('\n', '')
                     elif line.startswith('VERSION_ID='):
                         vm_ver = line.split('=')[1]
-                        vm_ver = vm_ver.split('.')[0]
                         vm_ver = vm_ver.replace('\"', '').replace('\n', '')
         except:
             log_and_exit(operation, UndeterminateOperatingSystem, 'Undeterminate operating system')
```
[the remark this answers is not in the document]
nidhanda: Fixed

AzureMonitorAgent/shim.sh
nidhanda: We don't have a preference for which py to go for in the case they are both available, But after the older distros moved to py3, majority of the distros in our support matrix now support py3 so I agree that it makes sense to make that the default

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
nidhanda: This was the only 'fix' I could get working. This happens only on py2.6 due to issues with some py2.6 versions being built with older ssl libraries which makes it not support https. I think having this comment with the link to the detailed issue should be of sufficient help in diagnosis of any future bugs (if they are introduced) .

AzureMonitorAgent/agent.py
```diff
@@ -540,36 +580,66 @@ def metrics_watcher(hutil_error, hutil_log):
...
                                 HUtilObject.error(me_rm_msg)
                     else:
-                        crc = hashlib.sha256(data).hexdigest()                    
+                        crc = hashlib.sha256(data.encode('utf-8')).hexdigest()                    
+
+                        if(crc != last_crc):
+                            # Resetting the me_msi_token_expiry_epoch variable if we set up ME again.
+                            me_msi_token_expiry_epoch = None
+                            hutil_log("Start processing metric configuration")
+                            hutil_log(data)
```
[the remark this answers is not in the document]
nidhanda: Replaced HUtilObject calls with wrapper function

hestolz (approved):
later on we may want to revisit `open()`s where we might be doing implicit conversion, changing them to read with binary mode and then following with manual decode with `utf-8`. for now these are largely copies of usage from `omsagent.py` which we haven't noted any issues for, so this doesn't need to block the feature.

simathih (approved):

PR #1254, opened 2020-11-16
setup_me is deleting generating MSI auth token

simathih:
- generate MSI token after setup+me


simathih:
Nikhil made the same change

PR #1259, opened 2020-11-25
Update name from / to \

simathih:


LAD-AMA-Common/telegraf_utils/telegraf_name_map.py
```diff
@@ -60,8 +60,8 @@
 
 # VM Insights
 # "Network\ReadBytesPerSecond", "Network\WriteBytesPerSecond"
-"Network/ReadBytesPerSecond" : {"plugin":"net_vmi", "field":"Network/ReadBytesPerSecond", "op":"rate", "module":"network"},
-"Network/WriteBytesPerSecond" : {"plugin":"net_vmi", "field":"Network/WriteBytesPerSecond", "op":"rate", "module":"network"},
+"Network\\ReadBytesPerSecond" : {"plugin":"net_recv_vmi", "field":"Network\\\\\\\\ReadBytesPerSecond", "op":"rate", "module":"network"},
```
nidhanda, line 63: why are we using multiple \'s . Is it just a delimiter choice or a requirement from VMI?

hestolz: just \ would escape the proceeding R (\R), need \\ for literal \

simathih: Because this is written to the telegraf config file where it ends up with 4. Then when the GO code picks it up it generates metrics with 2 slashes which both ME and mdsd parse as just a single backslash. Is there an easier way?

simathih: Problem is it has to be picked up python followed by GO and finally in C++. Each layer does an escape

hestolz: do python/go/c++ support raw strings? i think in python you can do `r'your unescaped string'` but idk the others

simathih: GO picks it from the config file python generates and C++ picks it up from wire. Can we pass raw strings in these boundaries without escaping?

simathih: Btw I ended up with this after experimenting

simathih: I was pushing to use forward slash but PMs want backward 🙂

simathih: I have commented this now

nidhanda (approved):

PR #1268, opened 2020-12-03
autoconfig: set MONITORING_GCS_AUTH_ID_TYPE during autoconfig

abenbachir:


simathih (approved):

abenbachir:
the MONITORING_GCS_AUTH_ID_TYPE can be set in protected settings

PR #1273, opened 2020-12-09
Moving general lad sinks config back to protected from public. 

nidhanda:
The portal UI team needs to have AzMonSink config in public. So we had moved the entire sinks config from protected to public. But a better design approach is to only move AzMonSink config in public as the other sinks config can have urls and secrets that are better left in protected.

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -239,7 +239,7 @@ def create_core_components_configs():
     ladconfig = configurator._ladCfg()
     sinks = LadUtil.getFeatureWideSinksFromLadCfg(ladconfig, 'performanceCounters')
     for name in sinks:
-        sink = configurator._sink_configs.get_sink_by_name(name)
+        sink = configurator._sink_configs_public.get_sink_by_name(name)
         if sink is not None:
             if sink['name'] == 'AzMonSink':
                 enable_metrics_ext = True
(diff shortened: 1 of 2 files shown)


Diagnostic/diagnostic.py
```diff
@@ -238,11 +238,10 @@ def encrypt_string(cert, secret):
     global enable_metrics_ext
     ladconfig = configurator._ladCfg()
     sinks = LadUtil.getFeatureWideSinksFromLadCfg(ladconfig, 'performanceCounters')
```
hestolz, line 240: `sinks` no longer being used

Diagnostic/diagnostic.py
```diff
@@ -238,11 +238,10 @@ def encrypt_string(cert, secret):
     global enable_metrics_ext
```
hestolz, line 238: is this still being used

hestolz (changes requested):

Diagnostic/diagnostic.py
nidhanda: Yeah it is used down in the main and start_mdsd methods

Diagnostic/diagnostic.py
nidhanda: Ah true. removing it now

hestolz (approved):

PR #1279, opened 2020-12-15
Pass domain from protected settings

abenbachir:
Azure Stack in mooncake doesn't support IMDS, they filed an ICM, they need a way to skip domain retrieval from IMDS and use the one passed from from settings.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -588,8 +588,14 @@ def enable():
     if proxy is not None:
         proxyParam = '-p {0}'.format(proxy)
 
-    # detect opinsights domain using IMDS
-    domain = get_azure_cloud_domain()
+    # get domain from protected settings
+    domain = protected_settings.get('domain')
+    if domain is None:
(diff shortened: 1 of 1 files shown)


OmsAgent/omsagent.py
```diff
@@ -588,8 +588,14 @@ def enable():
...
 
-    # detect opinsights domain using IMDS
-    domain = get_azure_cloud_domain()
+    # get domain from protected settings
+    domain = protected_settings.get('domain')
+    if domain is None:
+        # detect opinsights domain using IMDS
+        domain = get_azure_cloud_domain()
+    else:
+        hutil_log_info("Domain retrieved from protected settings '{0}".format(domain))
```
hestolz, line 597: nit missing close `'`

hestolz (approved):

OmsAgent/omsagent.py
abenbachir: done

hestolz (approved):

PR #1290, opened 2021-01-14
Fixing AMA shim.sh 

nidhanda:
1) The handlermanifest's commands were missing the prefix  "./" to execute the shim.sh file
2) The packaging script was missing the command to make shim.sh executable


AzureMonitorAgent/packaging.sh
```diff
@@ -43,6 +43,9 @@ cp $input_path/MetricsExtension MetricsExtensionBin/
 # copy just the source of python-future
 cp -r ext/python-future/src/* ext/future
 
+# make the shim.sh file executable
+chmod +x shim.sh
```
hestolz, line 47: fwiw I am not sure this is necessary ... it is not done for omsagent

hestolz (approved):

AzureMonitorAgent/packaging.sh
nidhanda: hmm, I am not sure how else it would be made executable? OMS is also not using the "./" , so the way they're called is different maybe

hestolz: is it not working if you don't add that? 
i looked at waagent code, seems they are calling `Popen`, and using `shell=True`. so it should work fine without `chmod`

nidhanda: No, it isn't running without it. The default perms are 644, so I just tested by changing shim.sh back to 644 and running waagent restart (which calls enable each restart) and it failed with permission denied. It will need the executable permission to run.

PR #1297, opened 2021-02-09
Changing account name according to MetricsExtension builds requirement

nidhanda:
As per the request of the ME team, converting CUSTOMMETRIC_ to lowercase, as is required by the new ME versions to send metrics to MDM. 

This is the build of ME that we tested this against -buildId=38988363 (It is retained in the custom ME build pipeline that we use for LAD/AMA metrics)

Creating the PR and leaving it unmerged for now. Will need to merge it when we pull in the latest ME for future LAD/AMA releases.

### DIFF
--- a/LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
+++ b/LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
@@ -338,9 +338,9 @@ def start_metrics(is_lad):
         az_resource_id, subscription_id, location, data = get_imds_values(is_lad)
 
         if is_lad:
-            monitoringAccount = "CUSTOMMETRIC_"+ subscription_id
+            monitoringAccount = "custommetric_"+ subscription_id
         else:
-            monitoringAccount = "CUSTOMMETRIC_"+ subscription_id + "_" + location
+            monitoringAccount = "custommetric_"+ subscription_id + "_" + location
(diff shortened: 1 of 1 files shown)


LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -338,9 +338,9 @@ def start_metrics(is_lad):
         az_resource_id, subscription_id, location, data = get_imds_values(is_lad)
 
         if is_lad:
-            monitoringAccount = "CUSTOMMETRIC_"+ subscription_id
```
Dmitry-Matveev, line 341: I assume this does not introduce any possible back compat issues where older version of ME might still be on the machine expecting upper case?

Dmitry-Matveev (approved):

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
nidhanda: I am on an email thread with Michael. I'll ask him this to confirm as well. But we have to include the latest ME in the next release anyways due to the Custom metrics folk's ask (Not sure if that's only for LAD or for AMA as well, Luke will know).  So, this change should only go when we update the ME we ship with our extensions

PR #1299, opened 2021-02-10
Support adding GCS auth Id/type

abenbachir:
Setting protected settings for GCS auth to support AKV/dSMS cert rotation.
Fields: monitoringGCSAuthIdType, monitoringGCSAuthId

For testing you can set these values through protected settings and validate the value made it to /etc/default/mdsd
You should be looking for these: MONITORING_GCS_AUTH_ID_TYPE, MONITORING_GCS_AUTH_ID

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -253,7 +253,7 @@ def install():
     exit_if_vm_not_supported('Install')
 
     public_settings, protected_settings = get_settings()
-    
+
     package_directory = os.path.join(os.getcwd(), PackagesDirectory)
     bundle_path = os.path.join(package_directory, BundleFileName)
     os.chmod(bundle_path, 100)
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -270,12 +270,13 @@ def install():
...
         "MDSD_ROLE_PREFIX" : "/var/run/mdsd/default",
         "MDSD_SPOOL_DIRECTORY" : "/var/opt/microsoft/linuxmonagent",
-        "MDSD_OPTIONS" : "\"-l -A -c /etc/mdsd.d/mdsd.xml -d -r $MDSD_ROLE_PREFIX -S $MDSD_SPOOL_DIRECTORY/eh -e $MDSD_LOG/mdsd.err -w $MDSD_LOG/mdsd.warn -o $MDSD_LOG/mdsd.info\"",
+        "MDSD_OPTIONS" : "\"-A -c /etc/mdsd.d/mdsd.xml -d -r $MDSD_ROLE_PREFIX -S $MDSD_SPOOL_DIRECTORY/eh -e $MDSD_LOG/mdsd.err -w $MDSD_LOG/mdsd.warn -o $MDSD_LOG/mdsd.info\"",
         "MCS_ENDPOINT" : "handler.control.monitor.azure.com",
         "AZURE_ENDPOINT" : "https://monitor.azure.com/",
         "ADD_REGION_TO_MCS_ENDPOINT" : "true",
         "ENABLE_MCS" : "false",
         "MONITORING_USE_GENEVA_CONFIG_SERVICE" : "false",
+        "MDSD_USE_LOCAL_PERSISTENCY" : "true",
```
nidhanda, line 279: Why are we changing this from a command line argument to an env variable? Can this be modified through protected settings later on?

abenbachir: no, that can't changed from protected setting. However it better to use env variable than command line arguments to help with debugging.

nidhanda (approved):

PR #1308, opened 2021-02-24
Force reboot if normal reboot times out.

vimish:


vermashi:
Why only this specific state machine? Wouldn't all other state machines also potentially benefit from this?

What is the difference between the commands "reboot" and "systemctl -force reboot"?

Do we know the root cause behind failed reboots?

PR #1310, opened 2021-02-26
Adding Rate counters for VMI MDM metrics

nidhanda:
1) Added logic for generating config for the new telegraf aggregator plugin for ME (mdmratemetrics)
2) Fixed a bug where sourcing frequency was being set to 1 second for all VMI rate counters

### DIFF
--- a/LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
+++ b/LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
@@ -165,8 +165,8 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
                 if not is_vmi_rate_counter:
                     is_vmi_rate_counter = telegraf_json[omiclass][plugin][field]["displayName"] in vmi_rate_counters_list
             
-            if is_vmi_rate_counter:
-                min_interval = "1s"
+            # if is_vmi_rate_counter:
+            #     min_interval = "1s"
                 
             if is_vmi or is_vmi_rate_counter:
                 splitResult = plugin.split('_')
@@ -197,10 +197,17 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
             namespace = MetricsExtensionNamepsace
             if is_vmi or is_vmi_rate_counter:
                 namespace = "insights.virtualmachine"
+
+            if is_vmi_rate_counter:
+                # Adding "_rated" as a substring for vmi rate metrics to avoid renaming collisions
+                plugin_name = plugin + "_rated"
+            else:
+                plugin_name = plugin
+
             metricsext_rename_str += "\n[[processors.rename]]\n"
-            metricsext_rename_str += " "*2 + "namepass = [\"" + plugin + "\"]\n"
+            metricsext_rename_str += " "*2 + "namepass = [\"" + plugin_name + "\"]\n"
             metricsext_rename_str += "\n" + " "*2 + "[[processors.rename.replace]]\n"
-            metricsext_rename_str += " "*4 + "measurement = \"" + plugin + "\"\n"
+            metricsext_rename_str += " "*4 + "measurement = \"" + plugin_name + "\"\n"
             metricsext_rename_str += " "*4 + "dest = \"" + namespace + "\"\n"
 
             fields = ""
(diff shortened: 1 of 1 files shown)


LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -197,10 +197,17 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
             namespace = MetricsExtensionNamepsace
             if is_vmi or is_vmi_rate_counter:
                 namespace = "insights.virtualmachine"
+
+            if is_vmi_rate_counter:
+                # Adding "_rated" as a substring for vmi rate metrics to avoid renaming collisions
+                plugin_name = plugin + "_rated"
```
Dmitry-Matveev, line 203: Is that something that's expected / parsed OK on receiving end?

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -165,8 +165,8 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
                 if not is_vmi_rate_counter:
                     is_vmi_rate_counter = telegraf_json[omiclass][plugin][field]["displayName"] in vmi_rate_counters_list
             
-            if is_vmi_rate_counter:
-                min_interval = "1s"
+            # if is_vmi_rate_counter:
+            #     min_interval = "1s"
```
Dmitry-Matveev, line 169: Minor: If removal of this line helped to fix 1s vs. 60s interval for counters, it might mean that we used min_interval somewhere later in the code instead of using the provided value in config, so some subtle bug may still exist, although now it's not longer affected by our default min value.

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -289,14 +297,23 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
                     aggregator_str += " "*2 + "drop_original = true\n"
                     aggregator_str += " "*2 + "fieldpass = [" + non_ops_fields[:-2] + "]\n" #-2 to strip the last comma and space
                     aggregator_str += " "*2 + "stats = [\"mean\", \"max\", \"min\", \"sum\", \"count\"]\n\n"
+            
             elif is_vmi_rate_counter:
+                # Aggregator config for MDSD
                 aggregator_str += "[[aggregators.basicstats]]\n"
                 aggregator_str += " "*2 + "namepass = [\"" + plugin + "_mdsd\"]\n"
-                aggregator_str += " "*2 + "period = \"" + min_agg_period + "s\"\n"
+                aggregator_str += " "*2 + "period = \"" + min_interval + "\"\n"
```
Dmitry-Matveev, line 305: Replacing min_agg_period with min_interval seems correct if min_interval is influenced by configuration we receive, but I'm not sure if that's how min_interval is actually set - can you please confirm?

Dmitry-Matveev (approved):

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
nidhanda: Yes, the mdmratemetrics plugin for telegraf adds the "_rated" at the end of every field it aggregates. This was done so that later when we rename, it only renames the rate fields and not others. (there were renaming collisions with the other fields like memory percentage, processor utilization etc. that were getting lost due to all VMI metrics being renamed under the one measurement name - insights.virtualmachine

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
nidhanda: the min_interval value is initiated to a very large number in the code block above (line 161) and then later down in the for loop (line 228) when we iterate over every plugin, whenever we see an interval for a field lower than the current minimum we set it as the minimum. For ex, if cx sets CPU fields to- usage_active - 15sec, usage_system 10sec,  usage_user 30sec , we would go for min_interval of 10 sec polling to make sure we meet the minimum. This was our original design though it is certainly up for discussion/improvement.

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
nidhanda: This setting is only for VMI and their intervals are always set to 1 minute. but regardless whatever they set as their minimum interval, we will use that to aggregate metrics and poll them at half the interval time to make sure we meet their required time intervals

PR #1321, opened 2021-03-24
Fix systemctl crash on RHEL LVM.

vimish:
1. Don't make systemctl failure as hard failure in PatchBootSystem state
2. Force reboot if normal reboot doesn't work
3. Change is isolated RHEL 7 LVM where the crash is observed


ejarvi (approved):
LGTM

PR #1324, opened 2021-03-25
Optimizing Extension code

bpramod:
- made DiskUtil class Singleton to make sure that mount command is executed single time
- made changes in blobowriter to make GetBlobProperties only once per blob
- Optimized FSFreezer to get resource mount only once

### DIFF
--- a/VMBackup/main/blobwriter.py
+++ b/VMBackup/main/blobwriter.py
@@ -45,12 +45,16 @@ class BlobWriter(object):
         try:
             # get the blob type
             if(blobUri is not None):
-                blobType = self.GetBlobType(blobUri)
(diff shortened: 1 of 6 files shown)


VMBackup/main/blobwriter.py
```diff
@@ -45,20 +45,24 @@ def WriteBlob(self,msg,blobUri):
         try:
             # get the blob type
             if(blobUri is not None):
-                blobType = self.GetBlobType(blobUri)
```
vityagi, line 48: intialize blobType var with default -
blobType = "PageBlob"

vityagi (approved):
Looks good

PR #1325, opened 2021-03-25
Fsfreeze changes

mearvind:


VMBackup/main/fsfreezer.py
```diff
@@ -183,15 +184,15 @@ def freeze_safe(self,timeout):
                 try:
                     if not os.path.isdir('/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock'):
                         os.mkdir('/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock')
-                    file = open("/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock/SafeFreezeLockFile","w")
+                    self.file = open("/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock/SafeFreezeLockFile","w")
```
bpramod, line 187: do we need to keep the file pointer open till thaw.
can we close here and open in thaw at the time of releasing the lock ?

VMBackup/main/fsfreezer.py
```diff
@@ -248,10 +238,23 @@ def freeze_safe(self,timeout):
...
+        try:
+            os.remove("/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock/SafeFreezeLockFile")
+        except:
+            pass
+
     def thaw_safe(self):
         thaw_result = FreezeResult()
         unable_to_sleep = False
         if(self.skip_freeze == True):
+            self.releaseFileLock()
```
bpramod, line 257: if we have already skipped the freeze, do we need to call release here ?

VMBackup/main/fsfreezer.py
```diff
@@ -248,10 +238,23 @@ def freeze_safe(self,timeout):
             self.logger.log(error_msg, True, 'Error')
         return freeze_result,timedout
 
+    def releaseFileLock(self):
+        if (self.isAquireLockSucceeded == True):
+            try:
+                fcntl.lockf(self.file, fcntl.LOCK_UN)
+                self.file.close()
+            except:
+                pass
```
bpramod, line 247: even though you intend to eat the exception, better log the stack trace and pass the exception.
it would be helpful in debugging if there are any errors

VMBackup/main/fsfreezer.py
```diff
@@ -248,10 +238,23 @@ def freeze_safe(self,timeout):
...
+        try:
+            os.remove("/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock/SafeFreezeLockFile")
+        except:
+            pass
```
bpramod, line 251: same here

VMBackup/main/fsfreezer.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, patching, logger, hutil):
         self.isAquireLockSucceeded = True
         self.getLockRetry = 0
         self.maxGetLockRetry = 5
+        self.file = None
```
vityagi, line 127: rename to something like lockFile or safeFreezelockFile

VMBackup/main/fsfreezer.py
```diff
@@ -205,20 +206,9 @@ def freeze_safe(self,timeout):
             end_time = datetime.datetime.utcnow()
             self.logger.log("Wait time to aquire lock "+ str(end_time - start_time),True)
 
-            
             sig_handle = None
             if (self.isAquireLockSucceeded == True):
                 sig_handle=self.freeze_handler.startproc(args)
-                self.thaw_safe()
```
vityagi, line 212: was this a bug ?

bpramod: @vityagi - yeah...though got pushed 2 days back only. I found it while I was testing my code optimization changes and asked arvind to do needful changes

VMBackup/main/fsfreezer.py
```diff
@@ -287,6 +290,7 @@ def thaw_safe(self):
             self.log_binary_output()
             self.logger.log(error_msg, True, 'Error')
         self.logger.enforce_local_flag(True)
+        self.releaseFileLock()
```
vityagi, line 293: can this be under finally ? we may not need to add in different places then

VMBackup/main/fsfreezer.py
```diff
@@ -248,10 +238,23 @@ def freeze_safe(self,timeout):
...
 
+    def releaseFileLock(self):
+        if (self.isAquireLockSucceeded == True):
+            try:
+                fcntl.lockf(self.file, fcntl.LOCK_UN)
+                self.file.close()
+            except:
+                pass
+        try:
+            os.remove("/etc/azure/MicrosoftRecoverySvcsSafeFreezeLock/SafeFreezeLockFile")
```
vityagi, line 249: Is delete really required ?

VMBackup/main/fsfreezer.py
mearvind: No. If we close the file then another process is able to acquire the lock on it.

VMBackup/main/fsfreezer.py
mearvind: yes if in case lock is not released then delete will rule out any possibility of starving of another process

VMBackup/main/fsfreezer.py
mearvind: but there is no try block in thaw so do we add a try block as well? and we are calling it in the single place. under skip freeze we don't need that call

VMBackup/main/fsfreezer.py
mearvind: yes. removed it.

VMBackup/main/fsfreezer.py
mearvind: added logs

VMBackup/main/fsfreezer.py
mearvind: added logs

VMBackup/main/fsfreezer.py
```diff
@@ -248,6 +238,18 @@ def freeze_safe(self,timeout):
             self.logger.log(error_msg, True, 'Error')
         return freeze_result,timedout
 
+    def releaseFileLock(self):
+        if (self.isAquireLockSucceeded == True):
+            try:
+                fcntl.lockf(self.safeFreezelockFile, fcntl.LOCK_UN)
+                self.safeFreezelockFile.close()
+            except Exception as e:
+                self.logger.log("Failed to unlock:  "+ str(e),True)
```
bpramod, line 247: use this format for stack trace...you can refer other code locations

str(e), traceback.format_exc()

vityagi (commented):
Looks fine to me

VMBackup/main/fsfreezer.py
mearvind: updated the pr

bpramod (approved):

PR #1330, opened 2021-03-31
AMA: instruct the pkg manager to overwrite the conf file instead of prompting

canfikret:
This is to fix the rolling upgrade scenario.


canfikret:
agent.version file is not up-to-date?

Dmitry-Matveev (dismissed):

Dmitry-Matveev (approved):
`--force-confnew` looked good enough to untrained eye :)
Let's give it one more shot.

abenbachir (approved):

PR #1358, opened 2021-04-28
[AMA] Redirect python version to stdout

hestolz:


AzureMonitorAgent/shim.sh
```diff
@@ -29,7 +29,7 @@ then
     echo "No Python interpreter found, which is an AMA extension dependency. Please install Python 3, or Python 2 if the former is unavailable." >&2
     exit 52 # Missing Dependency
 else
-    ${PYTHON} --version
+    ${PYTHON} --version 2>&1
```
canfikret, line 32: And stdout is directed to file, right? Possibly waagent.log?

canfikret (approved):

AzureMonitorAgent/shim.sh
hestolz: yes, it is captured by `waagent`

PR #1359, opened 2021-05-05
Bug fix9740865 : No status-file update for AlreadyProcessed and timeout, multiple run handlings

vityagi:


VMBackup/main/freezesnapshotter.py
```diff
@@ -378,3 +394,33 @@ def takeSnapshotFromOnlyHost(self):
             self.logger.log('T:S snapshotall ends...', True)
 
         return run_result, run_status, blob_snapshot_info_array, all_failed, unable_to_sleep, is_inconsistent
+
+    def is_command_timedout(self, para_parser):
```
bpramod, line 398: to be on safer side , better have everything under try-catch so that it wont break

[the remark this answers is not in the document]
vityagi: As discussed offline, did not do it as HandlerUtil currently does not honour the "False" flag to not log during freeze

vityagi: Added

bpramod (approved):

PR #1360, opened 2021-05-11
AMA IMDS helpers

hestolz:
Originally needed for private links. Though I moved endpoint logic to mdsd, we may need the IMDS code in the future so I left it here along with a small fix for metrics.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -144,9 +144,9 @@ CloudDomainMap = {
     USSecCloudName:    ".microsoft.scloud"
 }
 
-# Endpoints, sans-domain(s)
-McsEndpoint   = "https://global.handler.control.monitor.azure"
-AzureEndpoint = "https://monitor.azure"
+# Endpoints, sans-region/domain(s)
+McsEndpointTemplate   = "https://{0}.handler.control.monitor.azure{1}"
+AzureEndpointTemplate = "https://monitor.azure{0}/"
 
 
 # Change permission of log path - if we fail, that is not an exit case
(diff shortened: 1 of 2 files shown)


AaronRM (dismissed):

AaronRM (approved):

AzureMonitorAgent/agent.py
```diff
@@ -1022,6 +1044,116 @@ def exit_if_vm_not_supported(operation):
...
+    mcs_regional_endpoint = McsEndpointTemplate
+    mcs_global_endpoint = McsEndpointTemplate.format("global", "{0}")
+    azure_endpoint = AzureEndpointTemplate
+
+    if region:
+        hutil_log_info('Detected cloud region "{0}" via IMDS'.format(region))
+        mcs_regional_endpoint = mcs_regional_endpoint.format(region, "{0}")
+    else:
+        hutil_log_error('Unknown cloud region, falling back to _____')
+        mcs_regional_endpoint = mcs_regional_endpoint.format("", "{0}") # TODO what should this be?
```
hestolz, line 1135: TODO

AzureMonitorAgent/agent.py
```diff
@@ -99,8 +99,8 @@ def __str__(self):
 
 # Global Variables
 PackagesDirectory = 'packages'
-# TO BE CHANGED WITH EACH NEW RELEASE IF THE BUNDLE VERSION CHANGES
-# TODO: Installer should automatically figure this out from the folder instead of requiring this update
+# This changes during release if there are changes to mdsd. The release pipeline
```
Dmitry-Matveev, line 102: We'd need to change those to another package soon with AMA 2.0 repackaging, do we need to plan some back-compat here or we can simply have a separate 2.0 extension that only knows of 2.0 packages?

AzureMonitorAgent/agent.py
```diff
@@ -1087,6 +1084,79 @@ def exit_if_vm_not_supported(operation):
...
+    except:
+        hutil_log_error('Unable to load Arc IMDS endpoint from {0}'.format(endpoint_filepath))
+    return endpoint
+
+
+def get_imds_endpoint():
+    """
+    Find the appropriate endpoint (Azure or Arc) for IMDS
+    """
+    azure_imds_endpoint = 'http://169.254.169.254/metadata/instance?api-version=2018-10-01'
```
Dmitry-Matveev, line 1115: If HTTP request to this URL is to be executed outside of Azure Networking, it might be easy enough to spoof it and feed something into the extension that's not expected, we may need to add some input validation next time around for basics like string length limits and so on for extra protection.

Dmitry-Matveev (approved):
Some non-blocking future-looking comments only.

AzureMonitorAgent/agent.py
AaronRM: Right now, it's split per-branch. See 'ama-2.0-preview' branch: https://github.com/Azure/azure-linux-extensions/blob/release/ama-2.0.0-preview/AzureMonitorAgent/agent.py

AzureMonitorAgent/agent.py
hestolz: IMDS actually comes from the host node, so HTTP isn't an issue (in fact, IMDS won't even respond over HTTPS). [Docs](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=linux):
```
IMDS is a REST API that's available at a well-known, non-routable IP address (169.254.169.254). You can only access it from within the VM. Communication between the VM and IMDS never leaves the host. Have your HTTP clients bypass web proxies within the VM when querying IMDS, and treat 169.254.169.254 the same as 168.63.129.16.
```

Dmitry-Matveev: My worry is that holds true for Azure VMs only where MS controls networking (but I'm not certain). If AMA is on non-azure VM (via ARC, or later without ARC), can it be routable/spoofable IP? Last time I was working on similar subject with SDK team, it turned out that SDK may try to reach it from inside the app on-prem or on laptop or device, so IP is not controlled by MS at this situation and we needed to do some protection. I'm trying to ensure we do not miss similar case for agent that might be running off-Azure controlled network.

hestolz: On Arc, Arc IMDS comes from localhost so I think it should be ok (in the sense that if someone is spoofing IMDS on the box, there are bigger problems at hand). I wasn't aware of plans to make this work in non-Azure/non-Arc scenario, but seems like the kind of thing that would require significant re-working of IMDS-reliant logic regardless. 

If we did want to add protection, what options do we have? We can't use HTTPS because IMDS will not respond to that.

Dmitry-Matveev: Makes sense for ARC, if ARC is not there, extension won't be setup in the first place. I'm not sure on our architecture to support no-ARC machine and how we are going (if at all) to communicate to IMDS there, this particular script might not be executed at all then.
Generally (when time comes to it), in such a case (when we cannot SSL to endpoint), best we can do is protecting against bad input (basic input validation), ensuring that we won't pass too big of a string into unamanaged code and so on.

PR #1361, opened 2021-05-13
Set task limits for mdsd in SUSE

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -266,7 +266,15 @@ def install():
     exit_code, output = run_command_with_retries_output(OneAgentInstallCommand, retries = 15,
                                          retry_check = retry_if_dpkg_locked,
                                          final_check = final_check_if_dpkg_locked)    
-    
+    # set task limits in suse
+    vm_dist, vm_ver = find_vm_distro('Install')
+    if vm_dist.lower().startswith('suse'):
+        run_command_and_log("mkdir -p /etc/systemd/system/mdsd.service.d")
+        run_command_and_log("touch /etc/systemd/system/mdsd.service.d/override.conf")
```
hestolz, line 273: nit touch not needed here, file would be created anyways with next line

hestolz, line 273: nit touch not needed here, file would be created anyways with next line

AzureMonitorAgent/agent.py
```diff
@@ -266,7 +266,15 @@ def install():
...
+        run_command_and_log("echo '[Service]' > /etc/systemd/system/mdsd.service.d/override.conf")
+        run_command_and_log("echo 'TasksMax=10000' >> /etc/systemd/system/mdsd.service.d/override.conf")
+        run_command_and_log("systemctl daemon-reload")
```
hestolz, line 276: could we check the exit code(s)? if this fails, sles12 agent will be broken anyways; might as well fail fast and let customer know rather than proceeding

AzureMonitorAgent/agent.py
```diff
@@ -266,7 +266,15 @@ def install():
     exit_code, output = run_command_with_retries_output(OneAgentInstallCommand, retries = 15,
                                          retry_check = retry_if_dpkg_locked,
                                          final_check = final_check_if_dpkg_locked)    
-    
+    # set task limits in suse
```
hestolz, line 269: nit suse 12

hestolz (changes requested):

AzureMonitorAgent/agent.py
```diff
@@ -266,7 +266,15 @@ def install():
...
+    vm_dist, vm_ver = find_vm_distro('Install')
+    if vm_dist.lower().startswith('suse'):
+        run_command_and_log("mkdir -p /etc/systemd/system/mdsd.service.d")
+        run_command_and_log("touch /etc/systemd/system/mdsd.service.d/override.conf")
+        run_command_and_log("echo '[Service]' > /etc/systemd/system/mdsd.service.d/override.conf")
+        run_command_and_log("echo 'TasksMax=10000' >> /etc/systemd/system/mdsd.service.d/override.conf")
```
Dmitry-Matveev, line 275: Is 10000 a data-driven number, e.g. cardinalty or severities we may need to track * cardinality of log levels / filters we may need to support (if we create task for each) + other tasks we typically create in AMA? We can be explicit with the limit in the docs (e.g. ask not to set more than X facilities, more than Y levels, more than Z filters and so on to control it / deflect to in case of issues).

simathih (commented):
I have set it to 65K which is the max. Basically we are making it limitless. I guess this makes more sense rather than calculating a specific limit

AzureMonitorAgent/agent.py
simathih: I have set it to 65K which is the max. Basically we are making it limitless. I guess this makes more sense rather than calculating a specific limit

Dmitry-Matveev: Typically (in case our limits are order of magnitude less than max), it's better to understand the actual limits. This helps to remove guesswork and to set expectations right for the consumers as well as control the behavior of the product better (if such point needs control at all). However, in case where our limit is 2-9x times of the max I can see a point where figuring out the actual limit is overhead as we're close enough to maximum to rely on it.

aswatt (dismissed):

Dmitry-Matveev (approved):

hestolz (approved):

hestolz:
LGTM, @simathih please merge whenever you are ready

PR #1362, opened 2021-05-14
ADE DP QOS error codes for luksAddKey path

ejarvi:
- add QOS specific error constant for improved tracking of errors during luks add key path
- exit with custom error if keypath is invalid or cryptsetup luksAddKey returns a nonzero exit code 
- exit with custom error if key slot not found after add ("ValueError: True is not in list")


vimish (approved):
LGTM

PR #1363, opened 2021-05-17
ADE SP KV URL equivalence

ejarvi:
This change considers KV URL's that only differ in case or trailing slash to be the same
https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#vault-name-and-object-name


VMEncryption/main/ExtensionParameter.py
```diff
@@ -156,6 +156,13 @@ def clear_config(self):
     def _is_encrypt_command(self, command):
         return command in [CommonVariables.EnableEncryption, CommonVariables.EnableEncryptionFormat, CommonVariables.EnableEncryptionFormatAll]
 
+    def _is_kv_equivalent(self, a, b):
+        # ignore trailing slash if present, and ensure case insensitive string comparison for key vault and object name
+        # https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#vault-name-and-object-name 
+        if a and a[-1] == '/': a = a.lower()[:-1]
```
vimish, line 162: Can the python's rstrip function be used to remove / from the end of the string?
Thinking of case if customers by multiple / at the end of the url.

ejarvi:
I hadn't considered that to be a valid KeyVault URL (the key vault spec doesn't allow for multiple trailing slashes for URI's).  I also have not observed that in any telemetry like we have the other cases. Maybe adding some code to reject that case in check_util would be preferable.  

Right now the code for checkutil won't block this, urlparse just treats trailing slashes as part of the path: 
>>> from urllib.parse import urlparse
>>> urlparse("https://keyvault.windows.net/foo/foo/////")
ParseResult(scheme='https', netloc='keyvault.windows.net', path='/foo/foo/////', params='', query='', fragment='')

VMEncryption/main/ExtensionParameter.py
ejarvi: I hadn't considered that to be a valid KeyVault URL (the key vault spec doesn't allow for multiple trailing slashes for URI's). I also have not observed that in any telemetry like we have the other cases. Maybe adding some code to reject that case in check_util would be preferable.

Right now the code for check_util won't block it though, urlparse just treats trailing slashes as part of the path:

from urllib.parse import urlparse
urlparse("https://keyvault.windows.net/foo/foo/////")
ParseResult(scheme='https', netloc='keyvault.windows.net', path='/foo/foo/////', params='', query='', fragment=''

vimish (approved):

VMEncryption/main/ExtensionParameter.py
ejarvi: This PR that was not yet merged may be an option for blocking multiple trailing slashes or other URI's that do not meet the more restrictive KV rules:  https://github.com/ejarvi/azure-linux-extensions/pull/3/files

ejarvi: latest commit should cover this

PR #1365, opened 2021-05-26
ADE SP add check for pre-installed packages to ubuntu patching

ejarvi:
- checks for presence of pre-installed packages on ubuntu, and skip if already installed
- better performance in mainline scenario when packages are present in base image
- better reliability when external package management servers are unavailable or within isolated networks

### DIFF
--- a/VMEncryption/main/patch/UbuntuPatching.py
+++ b/VMEncryption/main/patch/UbuntuPatching.py
@@ -47,33 +47,44 @@ class UbuntuPatching(AbstractPatching):
         self.umount_path = '/bin/umount'
         self.touch_path = '/usr/bin/touch'
 
+    def packages_installed(self, packages):
+        ''' return true if all packages in list are already installed '''
+        installed = True
+        for package in packages:
+            cmd = "dpkg-query -s {package} | grep -q 'install ok installed'"
+            if not self.command_executor.ExecuteInBash(cmd, False, None, None, True):
+                installed = False
+                self.logger.log("{1} package not yet installed".format(package))
+        return installed
+
     def install_cryptsetup(self):
         packages = ['cryptsetup-bin']
-        cmd = " ".join(['apt-get', 'install', '-y', '--no-upgrade'] + packages)
-        return_code = self.command_executor.Execute(cmd, timeout=30)
-        if return_code == -9:
-            msg = "Command: apt-get install timed out. Make sure apt-get is configured correctly and there are no network problems."
-            raise Exception(msg)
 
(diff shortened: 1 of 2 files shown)


ejarvi:
this didn't get merged with other PR's to the test extension, may require a test extension rebuild/retest

ejarvi:
testing in azure:ej-ade-spd-june branch ( https://github.com/Azure/azure-linux-extensions/pull/1392 )

VMEncryption/main/patch/UbuntuPatching.py
```diff
@@ -47,33 +47,44 @@ def __init__(self, logger, distro_info):
...
-            # Fail early if apt-get update times out.
-            if return_code == -9:
-                msg = "Command: apt-get -o Acquire::ForceIPv4=true -y update timed out. Make sure apt-get is configured correctly."
-                raise Exception(msg)
-            cmd = " ".join(['apt-get', 'install', '-y'] + packages)
+
+        if self.packages_installed(packages):
+            return
+        else:
+            cmd = " ".join(['apt-get', 'install', '-y', '--no-upgrade'] + packages)
```
vimish, line 66: I thought --no-upgrade will not download if the package is already installed. Is that not working as expected.
See the --no-upgrade option description: https://linux.die.net/man/8/apt-get

VMEncryption/main/patch/UbuntuPatching.py
```diff
@@ -89,8 +100,14 @@ def install_extras(self):
                     'procps',
                     'psmisc']
 
-        cmd = " ".join(['apt-get', 'install', '-y'] + packages)
-        self.command_executor.Execute(cmd)
+        if self.packages_installed(packages):
```
vimish, line 103: This can prevent packages to get upgraded. It can become an issue if image package has some bug and we need the new package that contains the fix.
Should not happen often but something to note.

VMEncryption/main/patch/UbuntuPatching.py
ejarvi: The issue here is that we are trying to prevent an outbound call, which can timeout on isolated networks, and then cause an extension timeout.

VMEncryption/main/patch/UbuntuPatching.py
ejarvi: Yes, this shifts any package update responsibility to the image builder and system administrator who can do it outside of the time constraints imposed by the guest agent, or outside of network connectivity constraints that break this today.

PR #1369, opened 2021-06-04
use bytes type compare for p2/p3 compat

ejarvi:
-avoid TypeError on Python 3.x and support backwards compatibility with Python 2.7 by using a bytes-like object for comparison

### DIFF
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -360,7 +360,7 @@ def update_encryption_settings(extra_items_to_encrypt=[]):
             logger.log('Secret has already been updated')
             disk_util.log_lsblk_output()
 
-            if extension_parameter.passphrase and extension_parameter.passphrase != file(existing_passphrase_file).read():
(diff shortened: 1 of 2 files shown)


VMEncryption/main/handle.py
```diff
@@ -604,8 +604,8 @@ def mark_encryption(command, volume_type, disk_format_query):
 
 
 def is_daemon_running():
-    handler_path = os.path.join(os.getcwd(), __file__)
-    daemon_arg = "-daemon"
+    handler_path = os.path.join(os.getcwd(), __file__).encode('utf-8')
+    daemon_arg = b'-daemon'
```
vimish, line 608: Will it be possible to add a unit test for this function so that we can run it against both py 2 and 3?

ejarvi: That was bothering me too. I don't see an existing unit test wrapper for handle.py, and that blind spot may be how this slipped through prior testing.  I will look into moving this function out to a utility class, and make handle.py a little bit smaller   too.

PR #1370, opened 2021-06-04
Add AMA proxy support

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -310,8 +310,25 @@ def install():
     if public_settings is not None and public_settings.get("GCS_AUTO_CONFIG") == "true":
         hutil_log_info("Detecting Auto-Config mode.")
         return 0, ""
-    elif protected_settings is None or len(protected_settings) == 0:
+    elif (protected_settings is None or len(protected_settings) == 0) or (public_settings is not None and "proxy" in public_settings and "mode" in public_settings.get("proxy") and public_settings.get("proxy").get("mode") == "application"):
```
Dmitry-Matveev, line 313: Minor: I understand that it's "OR" clause, but I got a bit confused that on entry into the "ELIF" we check "protected_settings is None", but then on line 326 we check for "protected_settings is not None". Altough, technically it seems to work out just fine, it's a bit consusing in reading - is there a way to restructure IF/ELIF for readability?

AzureMonitorAgent/agent.py
```diff
@@ -310,8 +310,25 @@ def install():
...
+            
+            if "address" in public_settings.get("proxy"):
+                default_configs["MDSD_PROXY_ADDRESS"] = public_settings.get("proxy").get("address")
+            else:
+                log_and_exit("install", MissingorInvalidParameterErrorCode, 'Parameter "address" is required in proxy public setting')
+
+            if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == "true":
+                if protected_settings is not None and "proxy" in protected_settings and "username" in protected_settings.get("proxy") and "password" in protected_settings.get("proxy"):
+                    default_configs["MDSD_PROXY_USERNAME"] = protected_settings.get("proxy").get("username")
+                    default_configs["MDSD_PROXY_PASSWORD"] = protected_settings.get("proxy").get("password")
```
Dmitry-Matveev, line 328: I assume default_configs["VARIABLE"] are also protected by Root/Syslog level access, so that those secrets are not exposed to other user account on the same machine?

Dmitry-Matveev (approved):
Approved with comments.

AzureMonitorAgent/agent.py
simathih: Yeah only syslog user and root

AzureMonitorAgent/agent.py
simathih: The commonality in the elif is enabling MCS

PR #1377, opened 2021-06-09
handle empty string in _resource_disk_exists

ejarvi:


VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -86,8 +86,8 @@ def _is_luks_device(self):
         return (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success
 
     def _resource_disk_exists(self):
-        """ true if udev name for resource disk exists """
-        cmd = 'test -b ' + self._get_rd_base_dev_path()
+        """ true if udev name for resource disk exists and is mounted """
+        cmd = 'mountpoint -q ' + self._get_rd_base_dev_path()
         if (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success:
             self.logger.log("Resource disk exists.")
```
ShridharParameshwarBhat, line 92: Update in the comment saying "Resource disk exists and is mounted"

VMEncryption/main/ResourceDiskUtil.py
```diff
@@ -86,8 +86,8 @@ def _is_luks_device(self):
         return (int)(self.executor.Execute(cmd, suppress_logging=True)) == CommonVariables.process_success
 
     def _resource_disk_exists(self):
-        """ true if udev name for resource disk exists """
-        cmd = 'test -b ' + self._get_rd_base_dev_path()
+        """ true if udev name for resource disk exists and is mounted """
+        cmd = 'mountpoint -q ' + self._get_rd_base_dev_path()
```
vimish, line 90: I think self._get_rd_base_dev_path() returns the path to the resource disk itself (like /dev/disk/azure/resource) and not the mountpoint (/mnt/resource).
Will the logic not break in that case?
Correct me if my understanding is wrong here.

ejarvi: I think you're right - maybe need to split this into two separate checks:  

1. if base dev path exists
2. if fs is mounted on it

ejarvi: Actually the issue here is that self.get_rd_base_dev_path() returns empty string, and test -b on empty string returns 0 

(REPL)
>>> if rdu._get_rd_base_dev_path():
...   print("found")
... else:
...   print("not found")
...
not found

(bash)
[root@ejazwxwmhg02 main]# test -b
[root@ejazwxwmhg02 main]# echo $?
0

ejarvi: added check for empty string

VMEncryption/main/ResourceDiskUtil.py
ejarvi: reverted

PR #1383, opened 2021-06-17
Added decode in sub process output read method

mearvind:


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -540,8 +540,10 @@ def get_wala_version_from_command(self):
             while(process_wait_time > 0 and p.poll() is None):
                 time.sleep(1)
                 process_wait_time -= 1
-            out = p.stdout.read()
+            out = p.stdout.read().decode()
             out = str(out)
+            if len(out) == 0:
+                raise Exception("/usr/sbin/waagent returns null")
```
bpramod, line 546: does raising this exception cause backup failure. If Yes, don't throw any exception. this can be a soft failure.

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -442,7 +442,7 @@ def get_total_used_size(self):
                 time.sleep(1)
```
bpramod, line 442: Suggestion: I know that this is the existing code. Can we refactor by creating a new method that takes command as input and output as a response?

I see that this being used at other places. this helps in future changes to update in a single place.

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -540,8 +540,10 @@ def get_wala_version_from_command(self):
             while(process_wait_time > 0 and p.poll() is None):
                 time.sleep(1)
                 process_wait_time -= 1
-            out = p.stdout.read()
+            out = p.stdout.read().decode()
```
bpramod, line 543: will this decode not cause an issue with plain strings?
i.e. already existing VMs where byte array is not coming

bpramod, line 543: as mentioned offline, can you just check in BCM that whether we are throwing warning when version deserialization is failing

VMBackup/main/Utils/HandlerUtil.py
mearvind: I added it for testing, by mistake it came in the PR.removing it.

VMBackup/main/Utils/HandlerUtil.py
mearvind: No. It doesn't throw an error for normal string. I tested it.

mearvind: The version() throwing below error..
Exception in IsUnsupportedGuestAgentVersion System.ArgumentException: Version string portion was too short or too long.
   at System.Version.VersionResult.SetFailure(ParseFailureKind failure, String argument)
   at System.Version.TryParseVersion(String version, VersionResult& result)
   at System.Version.Parse(String input)
   at System.Version..ctor(String version)
   at Microsoft.Internal.CloudBackup.Service.Common.ExtensionManager.ExtensionManagerHelper.IsUnsupportedGuestAgentVersion(String osType, String guestAgentVersion) in X:\bt\1213499\repo\src\Service\ExtensionManager\ExtensionMgrHelper.cs:line 559

VMBackup/main/Utils/HandlerUtil.py
mearvind: done

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -811,6 +799,15 @@ def get_workload_running(self):
         
     def set_pre_post_enabled(self):
         self.pre_post_enabled = True
+
+    def command_output_from_subprocess(self , args, process_wait_time):
+        p = subprocess.Popen(args, stdout=subprocess.PIPE)
+        while(process_wait_time > 0 and p.poll() is None):
```
bpramod, line 805: where are we setting process_wait_time?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -811,6 +799,15 @@ def get_workload_running(self):
         
     def set_pre_post_enabled(self):
         self.pre_post_enabled = True
+
+    def command_output_from_subprocess(self , args, process_wait_time):
+        p = subprocess.Popen(args, stdout=subprocess.PIPE)
```
bpramod, line 804: Nit : Can we rename it instead of p

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -811,6 +799,15 @@ def get_workload_running(self):
         
     def set_pre_post_enabled(self):
         self.pre_post_enabled = True
+
+    def command_output_from_subprocess(self , args, process_wait_time):
```
bpramod, line 803: do we need try-catch?
or caller will handle any exceptions?

mearvind: It's better the caller function handles it. As we cannot suppress this error and we need to throw it so the caller function anyway needs to handle it.

VMBackup/main/Utils/HandlerUtil.py
mearvind: It's being set we call the function. We are passing it as a parameter

VMBackup/main/Utils/HandlerUtil.py
mearvind: done

bpramod (approved):

PR #1394, opened 2021-07-28
Add BEK disk validations.

vimish:


vimish:
@ejarvi 
I am using get_mount_items function ([(https://github.com/Azure/azure-linux-extensions/blob/ade-singlepass-dev/VMEncryption/main/DiskUtil.py#L463)). Please have a look if it can have any py2 and py3 compatibility issues.

@ShridharBhatGokarna Let me know if the exception message needs to change for our monitors to catch BEK missing exceptions.

ejarvi (commented):
This change has already been made in the 1.2.0.89 release branch and probably needs to be PR'd from there if not yet in ade-singlepass-dev (that way you also get release notes, etc.)

vimish:
@ejarvi I updated the version number to build a test extension. I will revert it before merging and then sync the dev and master branch to get the latest version and release notes.

PR #1397, opened 2021-08-06
Fix defaults file ordering

hestolz:
Have seen cases where `MDSD_OPTIONS` was defined before `MDSD_LOG` which meant that mdsd was trying to log in `/`. Also fixed trailing whitespace/tabs.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -421,7 +421,10 @@ def install():
             data = []
             new_data = ""
             vars_set = set()
-            dependent_vars = set(default_configs.keys()).intersection(["MDSD_OPTIONS"])
+            dependent_vars = ["MDSD_OPTIONS"]
+
+            # Scope to only dependent envvar being set by extension wrapper
+            dependent_vars = set(default_configs.keys()).intersection(dependent_vars)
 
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -421,23 +421,41 @@ def install():
             data = []
             new_data = ""
             vars_set = set()
```
hestolz, line 423: functional changes between here and line `455`

AzureMonitorAgent/agent.py
```diff
@@ -421,23 +421,41 @@ def install():
...
+            dependent_vars = set(default_configs.keys()).intersection(["MDSD_OPTIONS"])
+
+            # Copy existing comments/envvar to the updated defaults file; replace existing envvar values if appropriate
             with open(config_file, "r") as f:
```
simathih, line 427: Separate variable for mdsd options array

simathih (approved):

AzureMonitorAgent/agent.py
hestolz: @simathih fixed, thanks

simathih (approved):

PR #1400, opened 2021-08-13
Change output of subprocess.communicate() to str

aswatt:
subprocess.communicate() outputs a bytes-like object in this case, when we want it to be a string (for comparison purposes). In Python3, checking to see if "arc" is in output causes an error, since it's trying to compare a string and a bytes-like object. Using decode() on this bytes-like object will change it to a string, fixing this issue.

For this work item: https://msazure.visualstudio.com/One/_workitems/edit/10234552


hestolz (changes requested):
Please check `run_get_output` to see how best to handle this (e.g. the error handling mode should be set). Also special casing for just py3 may be needed.

aswatt:
Arc watcher function (which contained this change) has been deprecated: https://github.com/Azure/azure-linux-extensions/pull/1460

Original issue should no longer occur due to the removal of the function, and this change is therefore no longer necessary.

PR #1403, opened 2021-08-25
AMA: short circuit enable() logic when sequence number has not changed

thperapp:
Added logic to short circuit the enable() method when the [n].settings file in the config folder has not changed. 

The call to HandlerUtil.save_seq() stores the last sequence number that enable successfully ran for in mrseq. 
The check at the start of the function determines whether n (from most recent [n].settings file)> saved sequence number in mrseq before proceeding with enable()

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -524,7 +524,7 @@ def enable():
     the settings provided are incomplete or incorrect.
     Note: enable operation times out from WAAgent at 5 minutes
     """
-    global AMAServiceStartCommand, AMAServiceStatusCommand, HUtilObject
+    global AMAServiceStartCommand, AMAServiceStatusCommand
 
     if HUtilObject:
         if(HUtilObject.is_seq_smaller()):
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -524,7 +524,12 @@ def enable():
     the settings provided are incomplete or incorrect.
     Note: enable operation times out from WAAgent at 5 minutes
     """
-    global AMAServiceStartCommand, AMAServiceStatusCommand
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -524,7 +524,12 @@ def enable():
     the settings provided are incomplete or incorrect.
     Note: enable operation times out from WAAgent at 5 minutes
     """
-    global AMAServiceStartCommand, AMAServiceStatusCommand
+    global AMAServiceStartCommand, AMAServiceStatusCommand, HUtilObject
+
+    if HUtilObject:
+        if(HUtilObject.is_seq_smaller()):
+            HUtilObject.log("Current sequence number, " + HUtilObject._context._seq_no + ", is not greater than the sequence number of the most recent executed configuration. Exiting...")
+            sys.exit(0)
```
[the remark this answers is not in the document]
thperapp: Based on testing this, it looks like using sys.exit(0) doesn't result in an error in waagent.log, however it also doesn't update the status in Azure portal if the enable had previously failed. Updated to using the return code indicating success.

AzureMonitorAgent/agent.py
```diff
@@ -524,7 +524,11 @@ def enable():
     the settings provided are incomplete or incorrect.
     Note: enable operation times out from WAAgent at 5 minutes
     """
-    global AMAServiceStartCommand, AMAServiceStatusCommand
+    global AMAServiceStartCommand, AMAServiceStatusCommand, HUtilObject
```
hestolz, line 527: fwiw you don't need to `global` the HUtilObject; only needed if you are assigning

thperapp: Thanks for calling this out. Removed global HUtilObject and tested with new changes

hestolz (approved):
LGTM, thanks for the changes

simathih (approved):

PR #1404, opened 2021-08-26
Allow to use local path rather than passing the certificate and its key as parameter

williambernardet:
The goal of this change is to simplify the deployment by not having to pass the certificates as a parameter, but rely on an existing location under the provisioned VM.
This will allow to leverage the secrets option in the ARM template:

                    "secrets": [
                        {
                            "sourceVault": {
                                "id": "..."
                            },
                            "vaultCertificates": [
                                {
                                    "certificateUrl": "..."
                                }
                            ]
                        }
                    ]

and reduce the need to manage the certificates externally by passing parameter like this:

                    "certificate": "file:/var/lib/waagent/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.crt",
                    "certificateKey": "file:/var/lib/waagent/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.prv"

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -347,7 +347,7 @@ def install():
 
         # check if required GCS params are available
         MONITORING_GCS_CERT_CERTFILE = None
-        if "certificate" in protected_settings:           
+        if "certificate" in protected_settings:
             MONITORING_GCS_CERT_CERTFILE = base64.standard_b64decode(protected_settings.get("certificate"))
 
         if "certificatePath" in protected_settings:
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -348,11 +348,27 @@ def install():
         # check if required GCS params are available
         MONITORING_GCS_CERT_CERTFILE = None
         if "certificate" in protected_settings:
-            MONITORING_GCS_CERT_CERTFILE = base64.standard_b64decode(protected_settings.get("certificate"))
+            certificate = protected_settings.get("certificate")
+            # Try to handle a local path first if url style is provider,
+            # then fallback as base64 encoded certificate payload
+            if certificate.startsWith('file:'):
+                certificate = certificate[5:]
+                with open(certificate, 'r') as f:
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -350,10 +350,18 @@ def install():
         if "certificate" in protected_settings:
             MONITORING_GCS_CERT_CERTFILE = base64.standard_b64decode(protected_settings.get("certificate"))
 
+        if "certificatePath" in protected_settings:
+            with open(protected_settings.get("certificatePath"), 'r') as f:
```
hestolz, line 354: please wrap file IO in `try ... except`

hestolz (changes requested):

AzureMonitorAgent/agent.py
```diff
@@ -350,10 +350,25 @@ def install():
...
+                log_and_exit('Install', MissingorInvalidParameterErrorCode, 'Failed to read certificate {0}: {1}'.format(protected_settings.get("certificatePath"), ex))
+
         MONITORING_GCS_CERT_KEYFILE = None
         if "certificateKey" in protected_settings:
             MONITORING_GCS_CERT_KEYFILE = base64.standard_b64decode(protected_settings.get("certificateKey"))
 
+        if "certificateKeyPath" in protected_settings:
+            try:
+                with open(protected_settings.get("certificateKeyPath"), 'r') as f:
+                    MONITORING_GCS_CERT_CERTFILE = f.read()
```
hestolz, line 367: This should be `MONITORING_GCS_CERT_KEYFILE`

AzureMonitorAgent/agent.py
```diff
@@ -350,10 +350,25 @@ def install():
...
+            except Exception as ex:
+                log_and_exit('Install', MissingorInvalidParameterErrorCode, 'Failed to read certificate key {0}: {1}'.format(protected_settings.get("certificateKeyPath"), ex))
+
```
hestolz, line 370: nit remove extra newline

hestolz (changes requested):

hestolz (approved):
LGTM

PR #1408, opened 2021-08-31
fix for arc watcher

simathih:


simathih:
> Why are we doing this change? Also Fikret had to [make a PR](https://github.com/Azure/azure-linux-extensions/pull/1382) a few months back to address a case where the wrong processes were killed because no checking was done as to which PIDs were being killed

cx is hitting:
Status messageExtension returned non-zero exit code for Enable: 1. Extension Message: Enable failed with error: a bytes-like object is required, not 'str' Stacktrace: Traceback (most recent call last): File "./agent.py", line 190, in main exit_code, output = operations[operation]() File "./agent.py", line 528, in enable start_arc_process() File "./agent.py", line 866, in start_arc_process stop_arc_watcher() File "./agent.py", line 919, in stop_arc_watcher if output and "arc" in output: TypeError: a bytes-like object is required, not 'str'

AzureMonitorAgent/agent.py
```diff
@@ -687,10 +687,17 @@ def stop_metrics_process():
...
+                # Verify the pid actually belongs to AMA metrics watcher.
+                cmd_file = os.path.join("/proc", str(pid.strip("\n")), "cmdline")
+                if os.path.exists(cmd_file):
+                    with open(cmd_file, "r") as pidf:
+                        cmdline = pidf.readlines()
+                        if cmdline[0].find("agent.py") >= 0 and cmdline[0].find("-metrics") >= 0:
+                            kill_cmd = "kill " + pid
+                            run_command_and_log(kill_cmd)
+
+             run_command_and_log("rm "+pids_filepath)
```
hestolz, line 700: nit this should probably be 1 level un-indentend

hestolz (approved):

PR #1409, opened 2021-08-31
Logging in utc and writing status to temp file before renaming it

mearvind:
https://msazure.visualstudio.com/One/_workitems/edit/10503507

https://msazure.visualstudio.com/One/_workitems/edit/9293711

### DIFF
--- a/VMBackup/main/Utils/HandlerUtil.py
+++ b/VMBackup/main/Utils/HandlerUtil.py
@@ -167,13 +167,13 @@ class HandlerUtility:
                     pass
             else:
                 self._log(self._get_log_prefix() + message)
-            message = "{0}  {1}  {2} \n".format(str(datetime.datetime.now()) , level , message)
+            message = "{0}  {1}  {2} \n".format(str(datetime.datetime.utcnow()) , level , message)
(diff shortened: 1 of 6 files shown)


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -671,9 +671,11 @@ def do_status_report(self, operation, status, status_code, message, taskId = Non
 
     def write_to_status_file(self, stat_rept_file):
         try:
+            tempStatusFile =  os.path.join(self._context._status_dir, CommonVariables.TempStatusFileName)
             if self._context._status_file:
-                with open(self._context._status_file,'w+') as f:
+                with open(tempStatusFile,'w+') as f:
                     f.write(stat_rept_file)
+                os.rename(tempStatusFile, self._context._status_file)
```
vityagi, line 678: make sure this will overwrite when file already exist

vityagi (approved):

VMBackup/main/Utils/HandlerUtil.py
mearvind: ya tested that

PR #1411, opened 2021-09-01
Fix MONITORING_ROLE_INSTANCE typo

aswatt:
Code checks if MONITORING_TENANT exists before assigning MONITORING_ROLE_INSTANCE, rather than checking if MONITORING_ROLE_INSTANCE exists in this scenario (like the other settings do).


simathih (approved):

hestolz (approved):

AaronRM (approved):
Good catch Ashley!

PR #1413, opened 2021-09-02
Take Config sequence number from env variable

mearvind:


VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -97,7 +97,10 @@ def get_total_used_size(self):
                 self.logger.log("df return code"+str(df.returncode), True)
                 output = df.stdout.read().decode()
             if sys.version_info > (3,):
-                output = str(output, encoding='utf-8', errors="backslashreplace")
+                try:
```
mearvind, line 100: This is not related to this ConfidSeqNo work. but I saw many failures in kusto for this. line 101 throws below error which causes failure of size calculation. and report size as 0.

Unable to fetch total used space with error: decoding str is not supported, stack trace: Traceback (most recent call last):
  File "/var/lib/waagent/Microsoft.Azure.RecoveryServices.VMSnapshotLinux-1.0.9185.0/main/Utils/SizeCalculation.py", line 101, in get_total_used_size
    output = str(output, encoding='utf-8', errors="backslashreplace")
TypeError: decoding str is not supported

VMBackup/main/handle.sh
```diff
@@ -40,6 +40,16 @@ then
...
 fi
 
+configSeqNo="$(echo `printenv ConfigSequenceNumber`)"
+if [ -z ${configSeqNo} ]
+then
+	configSeqNo='seqNo:-1'
+	echo "`date -u`- ConfigSequenceNumber not found in environment variable ${configSeqNo}" >> $logfile
+else
+	configSeqNo='seqNo:'$configSeqNo
+	echo "`date -u`- ConfigSequenceNumber from environment variableb ${configSeqNo}" >> $logfile
```
vityagi, line 50: nit : variable typo

VMBackup/main/handle.py
```diff
@@ -87,12 +88,14 @@ def main():
                 update()
             elif re.match("^([-/]*)(daemon)", a):
                 daemon()
+            elif re.match("^([-/]*)(seqNo)", a):
```
vityagi, line 91: will it match as there is a suffix as well ? Doesnt the regex need some change to allow * at end ?

VMBackup/main/Utils/SizeCalculation.py
vityagi: ok good. 
Will the size calculation work now where it was failing earlier (or we can still get "Output of df command is not in desired format") ?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -271,12 +271,18 @@ def try_parse_context(self):
...
+                self._context._seq_no = self._get_current_seq_no(self._context._config_dir)
             if self._context._seq_no < 0:
                 self.error("Unable to locate a .settings file!")
                 return None
             self._context._seq_no = str(self._context._seq_no)
-            self.log('sequence number is ' + self._context._seq_no)
+            if seqNo != -1:
+                self.log('sequence number from environment varaible is ' + self._context._seq_no)
+            else:
+                self.log('sequence number is ' + self._context._seq_no)
```
vityagi, line 285: change this as well to 'sequence number based on config file-names is '

VMBackup/main/handle.sh
```diff
@@ -48,7 +58,7 @@ do
 	if [ -f "${cmnd}" ]
     then
 		echo "`date -u`- ${pythonVersion} path exists" >> $logfile
-		$cmnd main/handle.py -$1
+		$cmnd main/handle.py -$configSeqNo -$1
```
vityagi, line 61: although it should not matter based on the way parsing is happening in handle.py, to not change order, add $configSeqNo as second argument instead of first

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -235,7 +235,7 @@ def do_parse_context(self, operation):
             sys.exit(0)
         return _context
 
-    def try_parse_context(self):
+    def try_parse_context(self, seqNo):
```
vityagi, line 238: add a trace for seqNo

VMBackup/main/handle.py
mearvind: It matches, I tested it.  putting * means 0 or more times.

VMBackup/main/handle.sh
mearvind: It will matter if we don't pass config seq no. 1st as we use for loop and if enable is there before configSeqNo. then it will run without config seq no. So if we want to pass enable 1st then in handle.py I need to add a for loop to take sequence no. 1st then present for loop will run for enable

VMBackup/main/handle.py
vityagi, line 91: change "seqNo" to "seqNo:"

VMBackup/main/Utils/SizeCalculation.py
mearvind: Its not giving any such error on my VM now

VMBackup/main/handle.sh
vityagi: ok got it. Then keep it like this

VMBackup/main/Utils/HandlerUtil.py
mearvind: done

VMBackup/main/handle.py
vityagi: ok

mearvind: done

VMBackup/main/Utils/HandlerUtil.py
mearvind: We are already tracing it below  line 283

VMBackup/main/handle.sh
mearvind: done

VMBackup/main/Utils/HandlerUtil.py
vityagi: That will get traces only if it is not -1

vityagi: Mainly for diagnosability, better to trace original value received in parameter

VMBackup/main/handle.py
```diff
@@ -87,12 +88,14 @@ def main():
                 update()
             elif re.match("^([-/]*)(daemon)", a):
                 daemon()
+            elif re.match("^([-/]*)(seqNo:)", a):
+                configSeqNo = int(a.split(':')[1])
```
vityagi, line 92: I think, as a safety check, we should try to parse it as int and if that fails then reassign configSeqNo = -1

vityagi (approved):
Looks fine

PR #1414, opened 2021-09-02
Mountpoint with space

mearvind:


VMBackup/main/Utils/DiskUtil.py
```diff
@@ -67,7 +67,7 @@ def get_device_items_property(self, lsblk_path, dev_name, property_name):
         for i in range(0,len(lines)):
             item_value_str = lines[i].strip()
             if(item_value_str != ""):
-                disk_info_item_array = item_value_str.split()
+                disk_info_item_array = shlex.split(item_value_str)
```
vityagi, line 70: confirm and test that shlex is available in python 2.6 (this is the minimum version we support)

vityagi (approved):

PR #1426, opened 2021-09-16
Install latest OMI after omsagent

AaronRM:


Diagnostic/DistroSpecific.py
```diff
@@ -205,6 +205,11 @@ def install_lad_mdsd(self):
     def remove_lad_mdsd(self):
         return self.log_run_get_output('dpkg -P lad-mdsd')
 
+    def patch_lad_omi(self, openssl_ver):
+        # openssl_ver must be 100 or 110
+        cmd = "dpkg -i --refuse-downgrade omi-*.ssl_OPENSSLVER.ulinux.x64.deb".replace("OPENSSLVER", openssl_ver)
```
hestolz, line 210: Why is this wildcarded vs hardcode `1.6.8-1`?

AaronRM: Following pattern used in install_lad_mdsd. In case we would need to update/rebuild again, one less place to touch the code (it will just install whatever version is in the extension zip).

Diagnostic/diagnostic.py
```diff
@@ -177,6 +177,23 @@ def setup_dependencies_and_mdsd(configurator):
...
+    if re.match('^1.0.*', openssl_version):
+        openssl_ver = '100'
+    elif re.match('^1.1.*', openssl_version):
+        openssl_ver = '110'
+    else:
+        return 6, 'Unsupported openssl version. OpenSSL 1.0 or 1.1 must be installed. Detected version={0}'.format(openssl_version)
+
+    cmd_exit_code, cmd_output = g_dist_config.patch_lad_omi(openssl_ver)
+    if cmd_exit_code != 0:
+        return 7, 'omi pkg install failed. Exit code={0}, Output={1}'.format(cmd_exit_code, cmd_output)
```
hestolz, line 195: As Abder mentioned, for OMS we retry up to 5x, could we do retries here?

PR #1427, opened 2021-09-16
Update OMS shell bundle version in script

canfikret:


OmsAgent/apply_version.sh
```diff
@@ -15,9 +15,9 @@ sed -i "s/\"version\".*$/\"version\": \"$OMS_EXTENSION_VERSION\",/g" HandlerMani
 sed -i "s/^OMSExtensionVersion = .*$/OMSExtensionVersion = '$OMS_EXTENSION_VERSION'/"  watcherutil.py
 
 # updating omsagent.py
-# check BundleFileName = 'omsagent-1.12.7-0.universal.x64.sh'
+# check BundleFileName = 'omsagent-1.13.40-0.universal.x64.sh'
```
hestolz, line 18: Just make this generic; this comes from `omsagent.version`, don't want to have to update in multiple places each release

OmsAgent/omsagent.py
```diff
@@ -87,7 +87,7 @@ def __str__(self):
 ProceedOnSigningVerificationFailure = True
 PackagesDirectory = 'packages'
 keysDirectory = 'keys'
-BundleFileName = 'omsagent-1.13.11-0.universal.x64.sh'
+BundleFileName = 'omsagent-1.13.40-0.universal.x64.sh'
```
hestolz, line 90: Rather than updating like this, add comment explaining it is replaced at buildtime via `apply_version.sh` and change it to placeholder like "BundleFileName = 'omsagent-0.0.0-0.universal.x64.sh'

hestolz (changes requested):

hestolz (approved):

PR #1441, opened 2021-10-06
Prevent telegraf duplicates

hestolz:
It was reported that in some cases LAD may end up spawning numerous telegraf processes, leading to high resource consumption. This PR will address this by ensuring we
- Check for and stop previous telegraf processes before starting new one
- Do not start telegraf after Install and Enable both (should only happen for Enable); similarly remove redundant stop during Uninstall (already happens in Disable)
- When stopping telegraf, iterate through all PIDs that may be saved in the pidfile, in case there are multiple. Also, just read the `/proc/<pid>/cmdline` which seems simpler than making an `os.system` call, and is already well used in OMS/AMA wrappers.

263228506

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -324,19 +324,6 @@ def main(command):
                             'Exit code={0}, Output={1}'.format(cmd_exit_code, cmd_output))
             oms.tear_down_omsagent_for_lad(RunGetOutput, True)
 
-            #Stop the telegraf and ME services
-            tel_out, tel_msg = telhandler.stop_telegraf_service(is_lad=True)
-            if tel_out:
-                hutil.log(tel_msg)
-            else:
-                hutil.error(tel_msg)
-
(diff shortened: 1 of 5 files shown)


Diagnostic/diagnostic.py
```diff
@@ -361,27 +348,6 @@ def main(command):
                 hutil.do_status_report(g_ext_op_type, "error", '-1', "Install failed")
                 return
 
-            #Start the Telegraf and ME services on Enable after installation is complete
```
hestolz, line 364: No reason to be starting telegraf/ME yet during install, this is going to happen in enable

simathih:
> > Why are we updating the exceptions to returns? Do the callers in both LAD and omsagent handle it?
> 
> Are you referring to where I removed return statements after `raise Exception ...` ? Unless I'm really missing something, those returns are unreachable ... no point in those being there

sorry, for some reason, it looked like you were removing the raise and adding  the return :(

PR #1447, opened 2021-10-14
python 3 compat

simathih:


Diagnostic/shim.sh
```diff
@@ -0,0 +1,36 @@
...
+        eval ${python_exec_command}="python2"
+        eval ${future_path}="${PWD}/ext/future:"
+    fi
+}
+
+find_python PYTHON FUTURE_PATH
+
+if [ -z "$PYTHON" ] # If python is not installed, we will fail the install with the following error, requiring cx to have python pre-installed
+then
+    echo "No Python interpreter found, which is an AMA extension dependency. Please install Python 3, or Python 2 if the former is unavailable." >&2
```
AaronRM, line 29: Should be LAD, not AMA.

AaronRM: Is the intent that LAD will support Python3?

Diagnostic/shim.sh
```diff
@@ -0,0 +1,36 @@
...
+function find_python() {
+    local python_exec_command=$1
+    local future_path=$2
+
+    if command -v python3 >/dev/null 2>&1 ; then
+        eval ${python_exec_command}="python3"
+        # do not set future_path; future seems to cause interference with preexisting packages in python 3 environment
+    elif command -v python2 >/dev/null 2>&1 ; then
+        eval ${python_exec_command}="python2"
+        eval ${future_path}="${PWD}/ext/future:"
```
AaronRM, line 21: Are these files being added to the package?

Diagnostic/shim.sh
simathih: yeah, that will be part of this release

PR #1448, opened 2021-10-14
Sync Dev and Master branches

bamuir:
Seems that it has been a while since we last updated the master branch. I guess we must have been building release extensions off of dev, which I think is fine. Managing 4 different branches in our 2 different repos (vsts for releasing, github for working) for our extensions is a lot, I don't know that I personally see a need for the both the github branches, especially since the release branch here is dormant. 

We have ways of knowing which code branch commit/hash in git is released via VSTS, so maybe we can re think an easier way to keep things organized, as well as some OneNote updates to help us all release in the same manner and keep things in check with less complication. 

Open to suggestions, comments, etc. I can draft a proposal of how I think we can organize in better detail as well.


bamuir:
WE do not maintain master branch here, just using ade-singlepass-dev.

PR #1453, opened 2021-10-26
[AMA] Configure defaults in enable

hestolz:
- The configuration of the defaults file should occur during `enable()` not `install()`, otherwise subsequent enables with new configuration after initial install (e.g. to add a proxy) will have no effect.
- I believe we should be creating a **fresh defaults file each enable**, rather than updating/appending to an existing file. With the update/append model, we aren't actually applying the desired state, and bugs could occur (imagine user specifies a managed identity, but then wants to go back to using the default identity; with the update/append model, we will leave the `MANAGED_IDENTITY` variable set wrongly). It will incur a small penalty in that any addition to the `azuremonitoragent.default` file in the main repo will need to be reflected here.
- Use an `OrderedDict` to make the generation logic for the defaults file much cleaner

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -164,8 +164,6 @@ def main():
             operation = 'Update'
         elif re.match('^([-/]*)(metrics)', option):
             operation = 'Metrics'
-        elif re.match('^([-/]*)(arc)', option):
-            operation = 'Arc'
     except Exception as e:
         waagent_log_error(str(e))
 
@@ -356,7 +354,7 @@ def enable():
     global AMAServiceStartCommand, AMAServiceStatusCommand
 
     if HUtilObject:
-        if(HUtilObject.is_seq_smaller()):
+        if HUtilObject.is_seq_smaller():
             return 0, "Current sequence number, " + HUtilObject._context._seq_no + ", is not greater than the sequence number of the most recent executed configuration. Skipping enable"
 
     exit_if_vm_not_supported('Enable')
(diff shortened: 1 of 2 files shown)


aswatt (dismissed):

OmsAgent/omsagent.py
```diff
@@ -1059,6 +1059,8 @@ def detect_multiple_connections(workspace_id):
         # default encoding in python is ascii in python < 3
         if sys.version_info < (3,):
             output = utfoutput.decode('utf8').encode('utf8')
+        else:
+            output = utfoutput
```
simathih, line 1063: not related rt?

AzureMonitorAgent/agent.py
```diff
@@ -47,6 +47,7 @@
 import xml.dom.minidom
```
simathih, line 47: to be safe, I would recommend a e2e test in Canary where we first do install with no args, then call install again with proxy or any other settings to see if defaluts file is correct

simathih (dismissed):

AzureMonitorAgent/agent.py
```diff
@@ -464,116 +517,31 @@ def install():
...
-            # Set remaining non-dependent envvar that weren't present in the old defaults file
-            for var in list(default_configs.keys()):
-                if var not in vars_set and var not in dependent_vars:
-                    new_data += "export " + var + "=" + default_configs[var] + "\n"
-
-            # Finally, set envvar with dependencies (e.g. MDSD_OPTIONS depends on MDSD_LOG)
-            for var in dependent_vars:
-                new_data += "export " + var + "=" + default_configs[var] + "\n"
-                vars_set.add(var)
+            new_config = "\n".join(["export {0}={1}".format(key, value) for key, value in default_configs.items()])
```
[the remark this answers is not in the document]
AaronRM: I had missed that, thanks Henry. :)

OmsAgent/omsagent.py
hestolz: yes, but is a real issue surfaced by Pylance; if you prefer a separate PR I can remove this change

AzureMonitorAgent/agent.py
hestolz: testing this now locally

simathih (approved):

AzureMonitorAgent/agent.py
hestolz: @simathih on my test machine, this works as expected;
1. -enable with null public settings in 0.settings
2. add 1.settings with proxy content
3. add random # comment in defaults file
4. -enable again
5. defaults file now has proxy content and # comment is gone

hestolz:
There remains an issue (not introduced by this PR) where if AMA were enabled in multi-tenant mode and then subsequently enabled in MCS or single-tenant mode, the multi-tenant manager service will still be running (or vice versa). But this behavior is going to be reworked for the 1P/3P feature, so I'll leave the fix to a subsequent PR.

PR #1456, opened 2021-11-02
Remove logic to install insserv-compat package on SUSE 15

AaronRM:


AaronRM:
Separate issue from the other systemd logic detection, so the logic for SUSE is still needed. Closing this PR without merging.

PR #1457, opened 2021-11-10
Clearer logging when exiting due to seq number

aswatt:
Added a logging statement to this specific case, to make it clearer when enable() exits early due to the seq number.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -535,7 +535,9 @@ def enable():
 
     if HUtilObject is not None:
         if HUtilObject.is_seq_smaller():
-            return 0, "Current sequence number " + HUtilObject._context._seq_no + " is not greater than the sequence number of the most recent executed configuration, skipping enable."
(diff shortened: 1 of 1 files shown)


AaronRM (approved):

OmsAgent/omsagent.py
```diff
@@ -535,7 +535,9 @@ def enable():
 
     if HUtilObject is not None:
         if HUtilObject.is_seq_smaller():
-            return 0, "Current sequence number " + HUtilObject._context._seq_no + " is not greater than the sequence number of the most recent executed configuration, skipping enable."
+            log_output = "Current sequence number {0} is not greater than the sequence number of the most recent executed configuration, skipping enable.".format(HUtilObject._context._seq_no)
+            hutil_log_info(log_output)
+            return 0, log_output
```
hestolz, line 540: can you add this to AMA as well

aswatt: Made another PR for AMA: https://github.com/Azure/azure-linux-extensions/pull/1459

PR #1461, opened 2021-11-16
Fallback to platform-python on RHEL 8.0

AaronRM:
On RHEL 8.0, fallback to use platform-python if a user-installed python version is not present on the machine.

### DIFF
--- a/AzureMonitorAgent/shim.sh
+++ b/AzureMonitorAgent/shim.sh
@@ -21,7 +21,7 @@ function find_python() {
         eval ${future_path}="${PWD}/ext/future:"
     elif command -v /usr/libexec/platform-python >/dev/null 2>&1 ; then
         # If a user-installed python isn't available, check for a platform-python. This is typically only used in RHEL 8.0.
-        echo "User-installed python not found. Using /usr/libexec/platform-python as the python interpreter." >&1
+        echo "User-installed python not found. Using /usr/libexec/platform-python as the python interpreter."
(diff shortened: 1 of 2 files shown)


hestolz (changes requested):
I don't think we should be using this ... [per Red Hat](url), "It only includes the parts of Python needed for the system to function, and there are no guarantees that any particular feature won't be removed from it in the future ... If you are not re-building the distro, do not use Platform-Python directly. Install python3 and use that."

AzureMonitorAgent/shim.sh
```diff
@@ -19,6 +19,10 @@ function find_python() {
     elif command -v python2 >/dev/null 2>&1 ; then
         eval ${python_exec_command}="python2"
         eval ${future_path}="${PWD}/ext/future:"
+    elif command -v /usr/libexec/platform-python >/dev/null 2>&1 ; then
+        # If a user-installed python isn't available, check for a platform-python. This is typically only used in RHEL 8.0.
+        eval ${python_exec_command}="/usr/libexec/platform-python"
```
hestolz, line 24: consider adding some logging with `echo` so that we can see in telemetry whether an install/enable occurred with the `platform-python` package and if there were any subsequent errors as a result

hestolz (dismissed):
LGTM besides merge conflict

hestolz (dismissed):

hestolz (approved):

PR #1473, opened 2021-12-31
Arm64 support for AMA

thperapp:
Added code to check for aarch64 machines and replace the package bundle name. aarch64 packages are included in the /package directory in the build pipeline.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -109,6 +109,8 @@ PackageManager = ''
 PackageManagerOptions = ''
 MdsdCounterJsonPath = '/etc/opt/microsoft/azuremonitoragent/config-cache/metricCounters.json'
 
+SupportedArch = set(['x86_64', 'aarch64'])
+
 # Commands
 AMAInstallCommand = ''
 AMAUninstallCommand = ''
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -956,6 +957,15 @@ def parse_context(operation):
             raise ParameterMissingException
     return hutil
 
+def set_os_arch():
+    """
+    Checks if the system is x86 or aarch64 based and replaces package name
```
AaronRM, line 962: Can we make this more generic to allow for other architectures?

AaronRM (approved):

AzureMonitorAgent/agent.py
thperapp: We could have a list of supported architecture as a global variable and modify set_os_arch() as follows- 

supported_arch = ["x86_64","aarch64"]

def set_os_arch():

    global BundleFileName, supported_arch
    current_arch = platform.machine()
    
    for arch in supported_arch:
        if arch == current_arch:
            BundleFileName = BundleFileName.replace('x86_64',arch)

AaronRM (dismissed):

simathih (approved):

PR #1474, opened 2022-01-07
Setup encryption layer for all disks and do encryption in background.

vimish:
1. Setup encryption layer on all disk before starting the background encryption.
2. Return success after the encryption layer is set and VM has rebooted for OS encryption.
3. Reboot resilience for background encryption phase

### DIFF
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -623,11 +623,12 @@ def mark_encryption(command, volume_type, disk_format_query, encryption_mode=Non
     return encryption_marker
 
 def should_perform_online_encryption(disk_util, encryption_command, volume_type):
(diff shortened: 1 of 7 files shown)


VMEncryption/main/handle.py
```diff
@@ -595,17 +600,40 @@ def main():
...
+        hutil.do_status_report(operation='StartEncryption',
+                               status=CommonVariables.extension_success_status,
+                               status_code=str(CommonVariables.success),
+                               message='Online Encryption initialized.')
     return encryption_marker
 
+def should_perform_online_encryption(disk_util, encryption_command):
+    if disk_util.get_luks_header_size() != CommonVariables.luks_header_size_v2:
+        return False
+    if DistroPatcher.distro_info[0].lower() != "redhat":
```
bamuir, line 627: Is online encryption only feasible for redhat distros currently?

PR #1481, opened 2022-01-25
[AMA] Move enable short-circuit to after defaults generation

hestolz:
In upgrade scenario, waagent will do old disable > new update > old uninstall > new install > new enable. When AMA package is uninstalled it takes the /etc/default/azuremonitoragent file with it. After we added the enable short-circuit in #1403 as well as the config-in-enable changes in #1453, the upgrade scenario broke because we skipped the defaults file creation. This fixes that issue by only short-circuiting the service restart steps.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -117,6 +117,7 @@ SupportedArch = set(['x86_64', 'aarch64'])
 AMAInstallCommand = ''
 AMAUninstallCommand = ''
 AMAServiceStartCommand = ''
+AMAServiceRestartCommand = ''
 AMAServiceStopCommand = ''
 AMAServiceStatusCommand = ''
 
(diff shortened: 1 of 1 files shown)


thperapp (approved):

simathih (dismissed):

AzureMonitorAgent/agent.py
```diff
@@ -666,16 +672,16 @@ def update():
 def restart_pa():
     # start PA and agent launcher
     hutil_log_info('Handler initiating PA')
-    if is_systemd():       
-        exit_code, output = run_command_and_log('systemctl restart azuremonitor-coreagent && systemctl enable azuremonitor-coreagent')
+    if is_systemd():
+        exit_code, output = run_command_and_log('systemctl start azuremonitor-pipelineagent && systemctl enable azuremonitor-pipelineagent')
```
simathih: this is azuremonitor-coreagent

simathih (approved):

PR #1488, opened 2022-02-15
Changes to support ARM64 architecture

bpramod:
-> Changes to create a new binary for ARM64 architecture
-> CHanges to package newly created ARM64 binary folder

### DIFF
--- a/VMBackup/main/fsfreezer.py
+++ b/VMBackup/main/fsfreezer.py
@@ -145,6 +145,7 @@ class FsFreezer:
         self.isAquireLockSucceeded = True
         self.getLockRetry = 0
         self.maxGetLockRetry = 5
+        self.safeFreezelockFile = None
 
     def should_skip(self, mount):
         if(self.resource_disk_mount_point is not None and mount.mount_point == self.resource_disk_mount_point):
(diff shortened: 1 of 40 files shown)


mearvind (approved):

VMBackup/main/fsfreezer.py
```diff
@@ -150,7 +171,7 @@ def freeze_safe(self,timeout):
             self.logger.log(errMsg,True,'Warning')
         try:
             freeze_result = FreezeResult()
-            freezebin=os.path.join(os.getcwd(),os.path.dirname(__file__),"safefreeze/bin/safefreeze")
+            freezebin=os.path.join(os.getcwd(),os.path.dirname(__file__),self.safeFreezeFolderPath)
```
vityagi, line 174: Can we also test the error we get when we run wrong architecture executable. And based on that can we fallback to the other one ?

VMBackup/setup.py
```diff
@@ -188,6 +202,7 @@ def copy(src, dst):
 copybinary(workloadutils_folder, final_workloadutils_path)
 final_plugin_conf_path = final_folder_path + '/main'
 copybinary(binary_entry, final_binary_path)
+copybinary(binary_entry, final_Arm64binary_path)
```
vityagi, line 205: is the same binary being copied to both folders ?

vityagi (commented):
Some comments added

VMBackup/setup.py
bpramod: good catch....missed updating source path

VMBackup/main/fsfreezer.py
bpramod: actually, when we create the child process, we don't check whether that succeed....we were waiting for it to complete but it is getting crashed. it is tough to handle in such a way

vityagi (approved):
looks fine

PR #1493, opened 2022-03-02
Add support for Arm64 Metrics Extension

thperapp:
This PR is related to changes made here - 
[Pull Request 5771163](https://dev.azure.com/msazure/One/_git/Compute-Runtime-Tux/pullrequest/5771163): Arm64 Metrics Extension
The pipeline will download both x86 and arm64 versions of ME and agent.py will determine which binary to use during installation.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -1030,13 +1030,28 @@ def parse_context(operation):
 def set_os_arch():
     """
     Checks if the current system architecture is present in the SupportedArch set and replaces 
-    the package name accordingly
+    the package names accordingly
     """
     global BundleFileName, SupportedArch
     current_arch = platform.machine()
 
     if current_arch in SupportedArch:
+
+        # Replace the AMA package name according to architecture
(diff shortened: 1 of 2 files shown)


simathih:
would this code be redundant now? https://github.com/Azure/azure-linux-extensions/blob/f67e2509f5bff54c108abdb4cabe59ca72bf8566/LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py#L653

simathih (changes requested):
set_os_arch() would happen before this right? https://github.com/Azure/azure-linux-extensions/blob/f67e2509f5bff54c108abdb4cabe59ca72bf8566/LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py#L653

thperapp:
@simathih, yes. set_os_arch() will be called before that. The metrics_ext_handler code is still valid since set_os_arch() renames the appropriate binary to MetricsExtension.

simathih (approved):

PR #1503, opened 2022-03-18
Add Debian 10 Support

aswatt:


hestolz (commented):
is this added to https://github.com/Azure/azure-linux-extensions/blob/master/OmsAgent/extension-test/oms_extension_tests.py too?
Also need to update [here](https://docs.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview#linux)

NarineM:
We need to add a new error code for gpg and define it as 55 like we have for other pre -reqs (curl, etc.) and check it first before calling installer

aswatt:
> is this added to https://github.com/Azure/azure-linux-extensions/blob/master/OmsAgent/extension-test/oms_extension_tests.py too? Also need to update [here](https://docs.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview#linux)

For oms_extension_tests.py, it looks like we're missing several newer distros (ubuntu 20, rhel / centos / oracle 8, etc), should I add those in with this PR as well or should we make a separate PR to add all the missing distros?

aswatt:
Debian 10 support added in separate PR (https://github.com/Azure/azure-linux-extensions/pull/1540), closing out this PR.

aswatt:
Debian 10 support added in separate PR (https://github.com/Azure/azure-linux-extensions/pull/1540), closing out this PR.

PR #1505, opened 2022-03-23
Restore py2 compatibility

thperapp:
change os.replace() to os.rename() to restore compatibility with distros that only have python2. 
The replace call isn't supported in python2 and causes install failures.


hestolz (approved):
just curious, was this error not surfaced in our pre-release testing?

thperapp:
@hestolz we ended up abandoning the 1.17.1 at pre-release stage since we were seeing issues with custom logs before completing e2e on redhat, so this is currently only present in canary.

PR #1510, opened 2022-03-25
Fix for arm64 rhel8

thperapp:
AMA install on rhel8 initially does not create required symlinks and enable fails. Re-trying install creates the required symlinks.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -327,6 +327,14 @@ def install():
     # Copy the PA and agentlauncher binaries
     copy_pa_binaries()
 
+    # Retry install for aarch64 rhel8 VMs as initial install fails to create symlink to /etc/systemd/system/azuremonitoragent.service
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -327,6 +327,14 @@ def install():
     # Copy the PA and agentlauncher binaries
     copy_pa_binaries()
 
+    # Retry install for aarch64 rhel8 VMs as initial install fails to create symlink to /etc/systemd/system/azuremonitoragent.service
+    # in /etc/systemd/system/multi-user.target.wants/azuremonitoragent.service
+    if vm_dist.replace(' ','').lower().startswith('redhat') and vm_ver == '8.6' and platform.machine() == 'aarch64':
```
AaronRM, line 332: Use [lstrip](https://python-reference.readthedocs.io/en/latest/docs/str/lstrip.html) instead of .replace(' ','')

AaronRM (approved):

PR #1512, opened 2022-03-31
Create AMA Log Collector

aswatt:


AzureMonitorAgent/ama_tst/AMA-Troubleshooting-Tool.md
```diff
@@ -0,0 +1,37 @@
+# Troubleshooting Tool for Azure Monitor Linux Agent
```
NarineM, line 1: Can the tool be part of ama agent and be installed as part of ama extension install?  I assume this is just temporary solution till then?

AzureMonitorAgent/ama_tst/modules/logcollector.py
```diff
@@ -0,0 +1,243 @@
...
+            outfile.write("Linux OS detected: {0}\n".format(vm_dist))
+            outfile.write("Linux OS version detected: {0}\n".format(vm_ver))
+        else:
+            outfile.write("Indeterminate OS.\n")
+
+        # detected package manager
+        pkg_manager = helpers.find_package_manager()
+        if (pkg_manager != ""):
+            outfile.write("Package manager detected: {0}\n".format(pkg_manager))
+        else:
```
NarineM, line 124: Do you want to have a fall back here in case distro name match in the helper is returning "" as package manager name but it is still available? In case your helper code needs fixing, we will be still able to get customer env info correctly. Like below (sorry it is in shell script: format):
if [ "$pkg_manager" = "" ]; then
        if [ -x "$(command -v dpkg)" ]; then
            pkg_manager="DPKG"
        elif [ -x "$(command -v rpm)" ]; then
            pkg_manager="RPM"
        fi
    fi
if [ "$pkg_manager" = "" ]; then
        echo "Error: Indeterminate package manager"
        echo "Supported Sytems: 'DPKG' & 'RPM'" 
    fi

AzureMonitorAgent/ama_tst/AMA-Troubleshooting-Tool.md
NarineM: After giving it a second thought, if you are going to put the tool tgz on github, can you also add tool version info to the log when the tool is launched? Either add the version to the collector py in the header as a const and print that in the tool output, or create a version file? It will be hard to figure out what version of the tool css or customer is using in case we find issues with the tool when going over the logs.

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def find_package_manager():
+    """
+    Checks if the dist is debian based or centos based and assigns the package manager accordingly
+    """
+    pkg_manager = ""
+    dist, _ = find_vm_distro()
+
+    dpkg_set = set(["debian", "ubuntu"])
```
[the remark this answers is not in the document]
aswatt: I originally went with this because I think it's how we check the package manager in agent.py, but there have definitely been issues with distros needing to be updated manually (and having a second static list here will likely cause more issues as we add distros).

One concern though is that I think Debian systems have `/etc/*-release` as well - doing the `/etc/debian_version` check first should make this a moot point, but is there a better way of confirming this (i.e. some sort of `/etc/redhat_version` file or similar)?

AzureMonitorAgent/ama_tst/AMA-Troubleshooting-Tool.md
aswatt: @NarineM the tool is eventually planned to be installed with AMA, this is just temporary. I like the idea of having a version number though for the troubleshooting tool, added this to the ama_troubleshooter.sh file to be printed upon startup.

AzureMonitorAgent/ama_tst/modules/logcollector.py
aswatt: This is definitely valid, I added this logic into the `find_package_manager()` function (in case the function is used in other locations in the future).

AzureMonitorAgent/ama_tst/modules/helpers.py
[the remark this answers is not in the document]
aswatt: I found on Redhat VMs there's a file `/etc/redhat_version`, so I have checks for those two files and if neither exist, I added a fallback to just try and run the `command -v <pkg manager>` command you mentioned in a previous comment.

NarineM (approved):

PR #1527, opened 2022-04-27
Add support for Oracle.

vimish:
1. Add support for Oracle 8.5
2. Auto support minor version upgrade for redhat 8 and Oracle 8

### DIFF
--- a/VMEncryption/main/SupportedOS.json
+++ b/VMEncryption/main/SupportedOS.json
@@ -1,20 +1,9 @@
 {
   "redhat": [
         {
-          "Version" : "8.5"
-        },
-        {
-          "Version" : "8.4"
-        },
-        {
-          "Version" : "8.3"
-        },
-        {
-          "Version" : "8.2"
-        },
-        {
-          "Version" : "8.1"
-        },    
+          "Version" : "8",
+          "MinSupportedVersion" : "8.1"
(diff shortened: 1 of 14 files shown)


VMEncryption/main/SupportedOS.json
```diff
@@ -1,20 +1,9 @@
...
-        },
-        {
-          "Version" : "8.2"
-        },
-        {
-          "Version" : "8.1"
-        },    
+          "Version" : "8",
+          "MinSupportedVersion" : "8.1"
+        },
```
bamuir, line 6: can we do this with centOS as well?

bamuir (approved):

VMEncryption/main/SupportedOS.json
vimish: Online encryption changes are not yet fully enabled for CentOS. Also, I think CentOS is no longer in active development hence I avoided doing the change for CentOS. I will revisit it in future if there is demand from customers.

vimish: Online encryption changes are not yet fully enabled for CentOS. Also, I think CentOS is no longer in active development hence I avoided doing the change for CentOS. I will revisit it in future if there is demand from customers.

PR #1529, opened 2022-05-02
Fix resume encryption for new data disks.

vimish:
Fix the bug that does not call resume encryption for new data disks.


bamuir (approved):
Looks good to me

bkoonce-ms (approved):
LGTM

PR #1530, opened 2022-05-03
Move sequence number check to be for service restart only to fix patch version upgrade scenarios when agent onboarding is being skipped

NarineM:
The sequence number check was intended to determine if there was a config change from previous install/upgrade and whether enable () should be skipped because it is  called  after VM restart. However, this check only works when upgrade happens with minor version change and not with patch versions. Patch version falls into no config change path and in result agent onboarding is being skipped after upgrade .

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -541,13 +541,6 @@ def enable():
     the settings provided are incomplete or incorrect.
     Note: enable operation times out from WAAgent at 5 minutes
     """
-
-    if HUtilObject is not None:
-        if HUtilObject.is_seq_smaller():
-            log_output = "Current sequence number {0} is not greater than the sequence number of the most recent executed configuration, skipping enable.".format(HUtilObject._context._seq_no)
-            hutil_log_info(log_output)
-            return 0, log_output
-
     exit_if_vm_not_supported('Enable')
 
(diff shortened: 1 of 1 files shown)


hestolz:
The original change is still righteous; there is a real issue where enable() is called over and over again (because guest agent will call enable() on ALL extensions when there is a goal state change, such as when a different extension is being installed). It causes omsagent to restart and we've gotten many CRIs around this before we made this change. We can't just remove this.

We need to fix this in a different way. Maybe guest agent team can advise on whether there is a flag or other way to know whether we are doing an enable as part of a version upgrade.

NarineM:
@hestolz I discussed this with guest agent team, their recommendation was using running process state to determine if the extension needs to call enable() again. In OMS enable is the onboarding step that can not be easily determined by running process state. If original PR change was made only to prevent restart of oms processes, I can move this condition to be for line#  "run_command_and_log(RestartOMSAgentServiceCommand)".   Onboarding script already handles  re onboarding to same workspace cases and becomes no op if calling it for the same workspace and key.  Leaving extension enable () as no op breaks patch upgrade scenarios and can potentially impact minor version upgrades too so can not stay as it is now.

hestolz (approved):
LGTM now 👍

PR #1536, opened 2022-05-09
arm64 alma linux support

thperapp:


AzureMonitorAgent/agent.py
```diff
@@ -533,6 +533,11 @@ def enable():
     # Service(s) were successfully configured and started; increment sequence number
     HUtilObject.save_seq()
 
+    # Syslog logging doesn't work on arm64 AlmaLinux until the rsyslog service has been restarted
```
hestolz, line 536: As discussed in scrum this should not be necessary with the necessary changes in `common.sh`

hestolz (changes requested):

hestolz (approved):

PR #1540, opened 2022-05-19
Install GPG if not on VM + Debian 10 Support

aswatt:
We need GPG to validate the signature of the bundle before it's installed, adding a check for GPG to try and install it if it doesn't exist.

In addition, adding Debian 10 support, as this is the only blocker for Debian 10 (data verified to flow without issue if GPG is on VM).

(Note: PR also made in shell bundle code to attempt to install GPG at shell bundle step, https://github.com/microsoft/OMS-Agent-for-Linux/pull/1460, but as GPG is needed to validate the signature before the shell bundle, we need to have this check here as well.)

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -1024,9 +1024,12 @@ def exit_if_gpg_unavailable(operation):
     """
     check_exit_code, _ = run_get_output('which gpg', True, False)
     if check_exit_code is not 0:
+        hutil_log_info('GPG not found, attempting to install')
         exit_code, output = run_get_output(InstallExtraPackageCommand.format('gpg'))
         if exit_code is not 0:
             log_and_exit(operation, UnsupportedGpg, 'GPG could not be installed: {0}'.format(output))
+        else:
+            hutil_log_info('GPG successfully installed')
     return 0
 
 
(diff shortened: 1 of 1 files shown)


OmsAgent/omsagent.py
```diff
@@ -1010,6 +1016,20 @@ def exit_if_openssl_unavailable(operation):
...
+    """
+    Check if gpg is available to use
+    If not, attempt to install
+    If install fails, throw error to return UnsupportedGpg error code
+    """
+    check_exit_code, _ = run_get_output('which gpg', True, False)
+    if check_exit_code is not 0:
+        exit_code, output = run_get_output(InstallExtraPackageCommand.format('gpg'))
+        if exit_code is not 0:
+            log_and_exit(operation, UnsupportedGpg, 'GPG could not be installed: {0}'.format(output))
```
NarineM, line 1029: Do you want to log an informational message prior to install that says "Installing missing dependancy gpg.." then add another message "Gpg installed." if install completes with no error?

aswatt: Added some logging statements before and after, thank you for the catch! Let me know if I should change up the wording.

NarineM (approved):

PR #1549, opened 2022-05-26
Improve RD Only logic

ejarvi:
- Generate passphrase in case of resource disk only encryption when recovering from service heal or reallocation


bamuir (approved):
Looks good to me, passes unit tests locally.

PR #1554, opened 2022-06-08
[Abandoned] Skip gpg install if it is not available

NarineM:
Some Debian 10 are slow installing gpg through guest agent cmdline interface. When RDPing to the machine it shows gpg is installed, but agent vm extension stays in connecting state. Changing this logic to log missing gpg eror, and ask customers to install gpg manually before attempting to install extension till root cause is determined, and more robust solution is found with testing more images.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -287,7 +287,7 @@ def main():
     if exit_code is not 0:
         message = '{0} failed due to low disk space'.format(operation)
         log_and_exit(operation, exit_code, message)
-        
+
     exit_if_gpg_unavailable(operation)
 
     # Invoke operation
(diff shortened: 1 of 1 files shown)


OmsAgent/omsagent.py
```diff
@@ -1021,17 +1021,11 @@ def exit_if_openssl_unavailable(operation):
 def exit_if_gpg_unavailable(operation):
     """
     Check if gpg is available to use
-    If not, attempt to install
-    If install fails, throw error to return UnsupportedGpg error code
+    If not found, throw error to return UnsupportedGpg error code
     """
     check_exit_code, _ = run_get_output('which gpg', True, False)
     if check_exit_code is not 0:
-        hutil_log_info('GPG not found, attempting to install')
```
hestolz, line 1029: could you check extension install telemetry to see if this is a common thing that we do on distros other than debian 10? want to make sure that this isn't a common scenario, because if it is this could be a big regression in install success for customers

NarineM: @hestolz , this is a code Ashley added just recently specifically to enable Debian 10 but it currently has issues and needs to be tested with more distros before allowing it in extension. My change is reverting the install logic till more testing is done.

hestolz: great, thanks for the context, i had been under the impression this was a long-standing chunk of code

hestolz (approved):

aswatt (approved):
Lgtm, would recommend removing the [Abandoned] from the title, but otherwise is good!

PR #1555, opened 2022-06-08
GPG install only for Debian 10

aswatt:
Temporary workaround to unblock 1.14.15 release and allow for Debian 10 support.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -117,8 +117,9 @@ OnboardCommandWithOptionalParams = '{0} -w {1} -s {2} {3}'
 RestartOMSAgentServiceCommand = '{0} restart'.format(OMSAgentServiceScript)
 DisableOMSAgentServiceCommand = '{0} disable'.format(OMSAgentServiceScript)
 
-InstallExtraPackageCommand = 'apt-get -y update && apt-get -y install {0} || yum install -y {0} '\
-                    '|| zypper --non-interactive refresh && zypper --non-interactive install {0}'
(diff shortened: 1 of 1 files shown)


OmsAgent/omsagent.py
```diff
@@ -1024,14 +1025,21 @@ def exit_if_gpg_unavailable(operation):
...
-    if check_exit_code is not 0:
-        hutil_log_info('GPG not found, attempting to install')
-        exit_code, output = run_get_output(InstallExtraPackageCommand.format('gpg'))
-        if exit_code is not 0:
-            log_and_exit(operation, UnsupportedGpg, 'GPG could not be installed: {0}'.format(output))
-        else:
-            hutil_log_info('GPG successfully installed')
+    # Check if VM is Debian 10 (temporary fix to unblock Deb 10)
+    vm_supp, vm_dist, vm_ver = is_vm_supported_for_extension()
+    if (vm_supp and (vm_dist.lower().startswith('debian')) and (vm_ver.startswith('10')):
```
NarineM, line 1030: you can only check for debian , no need to check the version. We already know other images come with gpg installed and in case 11 needs install , you would not need to change this code. This gate is only needed to scope it to debian to remove the remaining piped commands from install cmdline.

NarineM, line 1030: you can only check for debian , no need to check the version. We already know other images come with gpg installed and in case 11 needs install , you would not need to change this code. This gate is only needed to scope it to debian to remove the remaining piped commands from install cmdline.

aswatt:
PR is out of date, created new one with cleaned-up changes here: https://github.com/Azure/azure-linux-extensions/pull/1570

PR #1570, opened 2022-07-13
GPG Install for Debian VMs

aswatt:
Adding ability to install GPG on Debian VMs, to allow for Debian 10 to be supported.

### DIFF
--- a/OmsAgent/omsagent.py
+++ b/OmsAgent/omsagent.py
@@ -117,8 +117,9 @@ OnboardCommandWithOptionalParams = '{0} -w {1} -s {2} {3}'
 RestartOMSAgentServiceCommand = '{0} restart'.format(OMSAgentServiceScript)
 DisableOMSAgentServiceCommand = '{0} disable'.format(OMSAgentServiceScript)
 
-InstallExtraPackageCommand = 'apt-get -y update && apt-get -y install {0} || yum install -y {0} '\
-                    '|| zypper --non-interactive refresh && zypper --non-interactive install {0}'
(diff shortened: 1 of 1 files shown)


OmsAgent/omsagent.py
```diff
@@ -117,8 +117,9 @@ def __str__(self):
 RestartOMSAgentServiceCommand = '{0} restart'.format(OMSAgentServiceScript)
 DisableOMSAgentServiceCommand = '{0} disable'.format(OMSAgentServiceScript)
 
-InstallExtraPackageCommand = 'apt-get -y update && apt-get -y install {0} || yum install -y {0} '\
-                    '|| zypper --non-interactive refresh && zypper --non-interactive install {0}'
+InstallExtraPackageCommandApt = 'apt-get -y update && apt-get -y install {0}'
+InstallExtraPackageCommandYum = 'yum install -y {0}'
+InstallExtraPackageCommandZypper = 'zypper --non-interactive refresh && zypper --non-interactive install {0}'
```
NarineM, line 122: Please remove yum install related constants as we are only adding  gpg install for only debian in this change.

NarineM (changes requested):

OmsAgent/omsagent.py
NarineM: Please also confirm that you tested these change and can see extension being successfully installed on clean debian 10 VM.

aswatt: Removed the yum and zypper commands, and tested the install through manually changing the omsagent.py file - GPG install was triggered and took about 5 seconds.

NarineM (approved):

PR #1484, opened 2022-01-28
LAD: fixing issue #1179. Redhat 8 policycoreutils-python to policycoreutils-python-utils

canfikret:

PR #1571, opened 2022-07-19
update policycoreutils-python

schoonercg:
update to policycoreutils-python-utils for class RedhatActions(CommonActions): to fix  Diagnostics extension install for RedHat.

### DIFF
--- a/Diagnostic/DistroSpecific.py
+++ b/Diagnostic/DistroSpecific.py
@@ -318,7 +318,7 @@ class CentosActions(RedhatActions):
 
     def install_required_packages(self):
         # policycoreutils-python missing on CentOS (still needed to manipulate SELinux policy)
-        return self.install_extra_packages(('policycoreutils-python',), True)
+        return self.install_extra_packages(('policycoreutils-python-utils',), True)
 
 
 DistroMap = {
--- a/Diagnostic/DistroSpecific.py
+++ b/Diagnostic/DistroSpecific.py
@@ -246,7 +246,7 @@ class RedhatActions(CommonActions):
     def install_required_packages(self):
         # policycoreutils-python missing on Oracle Linux (still needed to manipulate SELinux policy).
         # tar is really missing on Oracle Linux 7!
-        return self.install_extra_packages(('policycoreutils-python', 'tar'), True)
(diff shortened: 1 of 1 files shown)


Diagnostic/DistroSpecific.py
```diff
@@ -246,7 +246,7 @@ def install_extra_packages(self, packages, with_timeout=False):
     def install_required_packages(self):
         # policycoreutils-python missing on Oracle Linux (still needed to manipulate SELinux policy).
         # tar is really missing on Oracle Linux 7!
-        return self.install_extra_packages(('policycoreutils-python', 'tar'), True)
+        return self.install_extra_packages(('policycoreutils-python-utils', 'tar'), True)
```
simathih, line 249: would this exist in other distros too?

schoonercg: The coverage seems to be fairly distributed but I believe the deciding factor is whether the system is using python2 or python3. 
https://pkgs.org/search/?q=policycoreutils-python-utils
https://pkgs.org/download/policycoreutils
because it is a direct replacement package in general, can we check for either package to satisfy the condition?

aswatt:
Using `policycoreutils-python-utils` can cause problems on Redhat 6 and other older distros that we still support, so replacing the name for all Redhat (and CentOS) distros isn't ideal. However, we did create a separate path for newer VMs (e.g. Redhat 8) to use the correct name, while keeping the previous name for older distros - PRs below.

- #1484 (PR for Redhat 8)
- #1589 (PR for CentOS 8 + others)

Closing this PR as the requested changes have already been merged, thank you!

inandakumar:
I still get an error pasted below. Do I need to upgrade any RPM?
```
ERROR:[Microsoft.Azure.Diagnostics.LinuxDiagnostic-3.0.141] Failed to set up mdsd dependencies: package policycoreutils-python is not installed
ERROR:Last metadata expiration check: 0:20:47 ago on Fri 24 Mar 2023 09:48:08 AM UTC.
ERROR:No match for argument: policycoreutils-python
ERROR:Error: Unable to find a match: policycoreutils-python
```

PR #1575, opened 2022-07-22
Extension auto rollback testing purpose only - should be reverted

NarineM:
Create dummy extension package that will always fail for debian VM. It is needed to test oms extension auto rollback feature and will be reverted right after pipeline creates extension package. This change will be deployed only in one canary region.


hestolz (approved):
sounds good
lets prepare a roll forward version as well to minimize time this is in place in canary. thanks!

PR #1578, opened 2022-07-25
[LAD 4.0] Add shim

hestolz:
Adapted from AMA/OMS shims.

### DIFF
--- a/Diagnostic/diagnostic.py
+++ b/Diagnostic/diagnostic.py
@@ -450,7 +450,7 @@ def start_daemon():
     raise an exception (often OSError)
     :return: None
     """
-    args = ['python', g_diagnostic_py_filepath, "-daemon"]
+    args = ['python2', g_diagnostic_py_filepath, "-daemon"]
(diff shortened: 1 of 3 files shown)


Diagnostic/diagnostic.py
```diff
@@ -450,7 +450,7 @@ def start_daemon():
     raise an exception (often OSError)
     :return: None
     """
-    args = ['python', g_diagnostic_py_filepath, "-daemon"]
+    args = ['python2', g_diagnostic_py_filepath, "-daemon"]
```
AaronRM, line 453: Any chance that a Python2 distribution _doesn't_ have a 'python2' command?

AaronRM (approved):

Diagnostic/diagnostic.py
hestolz: we use the same logic for AMA/OMS and for both of those we check using that, and haven't seen any issue. so i think it should be fine

PR #1582, opened 2022-08-01
add tst - check ama installation

peizhen-w:


AzureMonitorAgent/ama_tst/modules/error_codes.py
```diff
@@ -0,0 +1,30 @@
+# # General Errors
+NO_ERROR = 0
+USER_EXIT = 1
+# ERR_SUDO_PERMS = 100
+ERR_FOUND = 101
+# ERR_PYTHON_PKG = 154
+
+# Warnings
```
aswatt, line 8: I'd recommend only keeping in the warnings that you're currently using in the TST - you can treat warnings like errors that won't immediately block the rest of the tests from completing

AzureMonitorAgent/ama_tst/modules/errors.py
```diff
@@ -0,0 +1,103 @@
...
+# error info edited when error occurs
+error_info = []
+
+# list of all errors called when script ran
+err_summary = []
+
+
+
+# # set of all errors which are actually warnings
+warnings = set([WARN_FILE_PERMS, WARN_LOG_ERRS, WARN_LOG_WARNS, WARN_LARGE_FILES, \
```
aswatt, line 21: Same note regarding the warnings as above - you can keep the infrastructure, but I'd recommend removing any warnings you're not using

AzureMonitorAgent/ama_tst/modules/errors.py
```diff
@@ -0,0 +1,103 @@
...
+    ERR_OS_VER : "This version of {0} ({1}) is not supported. Please download {2}. To see all "\
+          "supported Operating Systems, please go to:\n"\
+          "\n   https://docs.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview#linux\n",
+    ERR_OS : "{0} is not a supported Operating System. To see all supported Operating "\
+          "Systems, please go to:\n"\
+          "\n   https://docs.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview#linux\n",
+    ERR_FINDING_OS : "Coudln't determine Operating System. To see all supported Operating "\
+          "Systems, please go to:\n"\
+          "\n   https://docs.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview#linux\n",
+    ERR_FREE_SPACE : "There isn't enough space in directory {0} to install OMS - there needs to be at least 500MB free, "\
```
aswatt, line 39: Nitpick - AMA not OMS

AzureMonitorAgent/ama_tst/modules/errors.py
```diff
@@ -0,0 +1,103 @@
...
+          "but {0} has {1}MB free. Please free up some space and try installing again.",
+    ERR_PKG_MANAGER : "This system does not have a supported package manager. Please install 'dpkg' or 'rpm' "\
+          "and run this troubleshooter again.",
+    ERR_MULTIPLE_AMA : "There are more than one AMA installed, please remove the extra AMA packages.",
```
aswatt, line 43: Nitpick wording - "There is more than one instance of AMA installed"

AzureMonitorAgent/ama_tst/modules/errors.py
```diff
@@ -0,0 +1,103 @@
...
+    ERR_AMA_INSTALL : "AMA package isn't installed correctly.",
+    ERR_SUBMODULE_INSTALL : "Submodule(s) {0} not installed correctly.",
+    ERR_LOG_DAEMON : "No logging daemon found. Please install rsyslog or syslog-ng.",
+    ERR_SYSLOG_USER : "Syslog user is not created successfully.",
+    ERR_OLD_AMA_VER : "You are currently running AMA Version {0}. This troubleshooter only "\
+          "supports versions 1.9 and newer. Please upgrade to the newest version. You can find "\
```
aswatt, line 49: Have you confirmed that the TST works all the way back to AMA version 1.9 (I believe current version out is 1.21)? If you don't want to test back that far, you should put the oldest version that you tested the TST on

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
```diff
@@ -0,0 +1,92 @@
+import re
+import requests
+from bs4 import BeautifulSoup
```
aswatt, line 3: I don't think bs4 works without installing it on the customer's VM - I tested it in python2 and python3, and both couldn't find the module. I asked Dmitry about it and he mentioned that we should try and stick with industry standard tooling, and it seems like this is mostly just for parsing the HTML file. Can you find another tool to parse the HTML file instead that doesn't require `pip install <module>` (or worst-case scenario, create simple infrastructure and parse the file manually)?

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
```diff
@@ -0,0 +1,92 @@
...
+def ask_update_old_version(ama_version, curr_ama_version):
+    print("--------------------------------------------------------------------------------")
+    print("You are currently running AMA Verion {0}. There is a newer version\n"\
+          "available which may fix your issue (version {1}).".format(ama_version, curr_ama_version))
+    answer = get_input("Do you want to update? (y/n)", (lambda x : x.lower() in ['y','yes','n','no']),\
+                       "Please type either 'y'/'yes' or 'n'/'no' to proceed.")
+    # user does want to update
+    if (answer.lower() in ['y', 'yes']):
+        print("--------------------------------------------------------------------------------")
+        print("Please head to the Github link below\n")
```
aswatt, line 52: I'd just remove the Github link part - all of the instructions on how to update can be found in the MSFT docs page

AzureMonitorAgent/ama_tst/modules/main.py
```diff
@@ -41,15 +44,19 @@ def run_troubleshooter():
     while (run_again):
         print("================================================================================\n"\
             # TODO: come up with scenarios
-              "(NOTE: troubleshooting scenarios are coming, currently only log collection is available)\n"\
+              "1: Installation failures. \n"\
+              "2: Onboarding failures. \n"\
```
aswatt, line 48: Nitpick - would remove the "Onboarding failures" line until the onboarding section has been completed, may be confusing to customers / CSS

AzureMonitorAgent/ama_tst/modules/main.py
```diff
@@ -60,7 +67,62 @@ def run_troubleshooter():
...
+            print("No errors were found.")
+        # user requested to exit
+        elif (success == USER_EXIT):
+            return
+        # error found
+        else:
+            print("Please review the errors found above.")
+
+        # if user ran single scenario, ask if they want to run again
+        if (issue in ['1','2']):
```
aswatt, line 114: Same note as above - I'd remove the '2' since there's no scenario 2 right now

aswatt (changes requested):
Overall really awesome job, thank you so much for making all these installation checks! Mostly just nitpicks, but the one major thing is the bs4 / BeautifulSoup package - can you change that to use something that doesn't require installation?

Thank you again for all of the work you've put into this!

AzureMonitorAgent/ama_tst/modules/error_codes.py
```diff
@@ -0,0 +1,30 @@
+# # General Errors
+NO_ERROR = 0
+USER_EXIT = 1
+# ERR_SUDO_PERMS = 100
```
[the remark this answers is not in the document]
peizhen-w: Do I really need to check file permissions here? We check if a user runs TST as a root at the beginning, so I assume they should not have permission issues later? https://github.com/Azure/azure-linux-extensions/blob/95e3999f6289f76d7fef0bacbae64ff617ac9d2e/AzureMonitorAgent/ama_tst/modules/main.py#L34

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
peizhen-w: I changed from bs4 to lxml. It can be used on Azure VM without installation. It should work with python 2.7 and after.

AzureMonitorAgent/ama_tst/modules/errors.py
peizhen-w: TST adds installation checks from AMA 1.21.0 so I believe this should be the good version to put on

AzureMonitorAgent/ama_tst/modules/error_codes.py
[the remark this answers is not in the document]
aswatt: That's actually valid - I think I had added in that error code in the OMS TST before adding in the `check_sudo()` equivalent at the beginning. In that case, I'd recommend removing it entirely and any references to it. Good catch!

AzureMonitorAgent/ama_tst/modules/error_codes.py
```diff
@@ -1,18 +1,12 @@
 # # General Errors
 NO_ERROR = 0
 USER_EXIT = 1
-# ERR_SUDO_PERMS = 100
+ERR_SUDO_PERMS = 100
```
aswatt, line 4: I mentioned above, but you're right that this error code isn't needed - feel free to remove it (and any references to it) completely from the code

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -190,6 +195,40 @@ def check_ama_installed(ama_vers):
     return (ama_exists, ama_unique)
 
 
+# openssl connect to specific endpoint
+def check_endpt_ssl(ssl_cmd, endpoint):
```
aswatt, line 199: Small nitpick, but I'd recommend leaving out any of the connection checks you're working on (i.e. the `check_endpt_ssl()` function, the internet connection errors, etc) - just for future clarity's sake. This way we can keep it clean that this PR only has all of the installation changes in it, and the next PR you make will be all of the onboarding / connectivity changes

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
```diff
@@ -1,22 +1,20 @@
 import re
 import requests
-from bs4 import BeautifulSoup
+from lxml import html
```
aswatt, line 3: Unfortunately `lxml` runs into the same issue we had with `bs4` where it needs a separate install. Could you try using [the `xml.dom.minidom` library](https://docs.python.org/3/library/xml.dom.minidom.html) instead? I've confirmed it exists without extra install on base Python for 2.7 and 3.6.9, but beyond that could you verify it does everything it needs to do?

aswatt (changes requested):
Overall looks good! Just a few notes to remove onboarding checks for future PR readability, and one more fix to the HTML parsing stuff

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
```diff
@@ -1,25 +1,28 @@
...
-        versions = root.xpath('//table/tbody/tr/td[4]/text()')
-
-        for version in versions:
-            version = re.sub('[A-Za-z ]+', '', version)
-            if (version == ''):
-                continue
-            if (comp_versions_ge(version, ama_version)):
-                return (version, None)
+    try:           
+        r = requests.get('https://docs.microsoft.com/en-us/azure/azure-monitor/agents/azure-monitor-agent-extension-versions')
```
aswatt, line 13: Small nitpick, but would recommend using `AMA_URL` here since you define it above

aswatt (approved):
Looks beautiful! Thank you again for all of the work you've put into this!

Left a small nitpick regarding using the AMA_URL variable, but approving since everything else looks good from my end (and so my review doesn't block it from merging)

AzureMonitorAgent/ama_tst/modules/main.py
```diff
@@ -60,7 +66,62 @@ def run_troubleshooter():
...
+            "  - Interactive Mode includes extra checks that require input")
+        mode = get_input("Do you want to run the troubleshooter in silent (s) or interactive (i) mode?",\
+                        (lambda x : x.lower() in ['s','silent','i','interactive','q','quit']),\
+                        "Please enter 's'/'silent' to run silent mode, 'i'/'interactive' to run \n"\
+                            "interactive mode, or 'q'/'quit' to quit.")
+        if (mode.lower() in ['q','quit']):
+            print("Exiting the troubleshooter...")
+            return
+        elif (mode.lower() in ['s','silent']):
+            print("Running troubleshooter in silent mode...")
```
hestolz, line 82: is there a way that the troubleshooter can be entirely silent mode -- that is, a single command like `./entrypoint.py --some-option --another-option` and the TST will run to completion? my reading of it currently is you still need to choose the run mode _after_ starting TST (rather than being able to specify runmode in the initial TST start command).

AzureMonitorAgent/ama_tst/modules/install/install.py
```diff
@@ -0,0 +1,97 @@
...
+
+def check_syslog_user():
+    with open('/etc/passwd', 'r') as fp:
+        for line in fp:
+            if line.startswith('syslog:'):
+                return NO_ERROR
+    return ERR_SYSLOG_USER
+
+# check all packages are installed
+def check_installation(interactive, err_codes=True, prev_success=NO_ERROR):
```
hestolz, line 37: thought for the future: python is also an AMA dependency. obviously if we are running the troubleshooter already, there is python on the machine since TST is written in python. might be worth having TST entrypoint be a shell script shim (like shim.sh calling agent.py) which makes sure there is python and lets the user know if its not found.

AzureMonitorAgent/ama_tst/modules/main.py
peizhen-w: Yes the current design is to choose silent/interactive mode after starting it. This follows the design of OMS troubleshooter, but it seems good to have a command line option to start it as a silent mode without extra input

AzureMonitorAgent/ama_tst/modules/install/install.py
peizhen-w: The entry point of AMA TST is ama_troubleshooter.sh. The script file checks if python is installed and displays error if python is not found

AzureMonitorAgent/ama_tst/modules/install/install.py
```diff
@@ -0,0 +1,97 @@
...
+    # check Package manager (dpkg/rpm)
+    print("Checking if machine has a supported package manager...")
+    checked_pkg_manager = check_pkg_manager()
+    if (is_error(checked_pkg_manager)):
+        return print_errors(checked_pkg_manager)
+    else:
+        success = print_errors(checked_pkg_manager)
+    
+    # check Package + submodule installation states
+    print("Checking if packages and submodules are installed correctly...")
```
hestolz, line 66: nit/pedantic: I would probably call these "subcomponents" rather than submodules

AzureMonitorAgent/ama_tst/modules/install/check_os.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        error_info.append((vm_dist,))
+        return ERR_OS
+
+
+def check_os():
+    if platform.machine() == 'x86_64':
+        # 32 bit or 64 bit
+        cpu_bits = find_vm_bits()
+        if (cpu_bits == None or (cpu_bits not in ['32-bit', '64-bit'])):
```
hestolz, line 98: We don't support 32bit systems

AzureMonitorAgent/ama_tst/modules/install/check_os.py
```diff
@@ -0,0 +1,107 @@
+import platform
+
+from error_codes import *
+from errors      import error_info
+from helpers     import find_vm_bits, find_vm_distro
+
+supported_dists_x86_64 = {'redhat' : ['7', '8'], # Rhel
```
hestolz, line 7: i assume here we will keep track of the latest supported distro list. however customers may run the TST for older versions of AMA. is there a way to warn the customer if they have installed an old AMA version on a new distro that wasn't supported at the time?

hestolz: same for old versions that didn't support aarch64

AzureMonitorAgent/ama_tst/modules/install/check_pkgs.py
```diff
@@ -0,0 +1,39 @@
...
+
+def check_packages():
+    # check azuremonitoragent rpm/dpkg
+    ama_vers = find_ama_version()
+    if (ama_vers == None):
+        return ERR_AMA_INSTALL
+    if (len(ama_vers) > 1):
+        return ERR_MULTIPLE_AMA
+    
+    # find submodule binaries
```
hestolz, line 17: again, nit pick: would probably call these subcomponents

AzureMonitorAgent/ama_tst/modules/install/install.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # check Syslog user created successfully
+    print("Checking if syslog user exists...")
+    checked_syslog_user = check_syslog_user()
+    if (is_error(checked_syslog_user)):
+        return print_errors(checked_syslog_user)
+    else:
+        success = print_errors(checked_syslog_user)
+    print("============================================")
+    return success
```
hestolz, line 97: Could we check whether `azuremonitoragent` service is enabled? Or is that out of scope/planned for a different module of the troubleshooter

AzureMonitorAgent/ama_tst/modules/error_codes.py
```diff
@@ -0,0 +1,24 @@
+# # General Errors
```
hestolz, line 1: how are these surfaced to the user of the TST; are they printed in human readable format, are they the TST exit code, etc. ? If they aren't reported in a human-readable way, is the mapping from code -> error description documented anywhere (other than this file) e.g. could it be part of the [.md](https://github.com/Azure/azure-linux-extensions/blob/master/AzureMonitorAgent/ama_tst/AMA-Troubleshooting-Tool.md)

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
```diff
@@ -0,0 +1,86 @@
...
+                cell = row.getElementsByTagName("td")[3]
+                version = cell.firstChild.nodeValue
+                version = re.sub('[A-Za-z ]+', '', version)
+                if (version == ''):
+                    continue
+                if (comp_versions_ge(curr_version, version)):
+                    return (None, None)
+                else:
+                    return (version, None)
+    except Exception as e:
```
hestolz, line 28: customers may not necessarily have VM configured to allow connection to `docs.microsoft.com`. If there's an easy way to check this, let's do so. If it is more involved (e.g. we will do connection checks as a separate module in the TST) let's ensure this scenario is described to the user e.g. 
```
Failed to get latest AMA version from Azure documentation at `docs.microsoft.com`: <insert exception here>. Please ensure your network configuration (e.g. firewall, NSG) is configured to allow connectivity to this domain.
```

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -14,6 +16,17 @@
 except:
     DEVNULL = open(os.devnull)
 
+SSL_CMD = "echo | openssl s_client -connect {0}:443 -brief"
+MSFT_DOC_URL = "docs.microsoft.com"
```
hestolz, line 20: seem unused — are these part of a future module/enhancement? if not pls remove

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -93,16 +111,77 @@ def find_package_manager():
                 pkg_manager = "rpm"
             except subprocess.CalledProcessError:
                 pass
-
+    general_info['PKG_MANAGER'] = pkg_manager
```
hestolz, line 114: `general_info` is a global variable (since declared outside function scope). if you want to modify it within function scope, you need to declare `global general_info` at the top of the function. See [this page](https://www.w3schools.com/python/python_variables_global.asp)

Here's a simple example

```python
d = { "key": "pre" }

print(d["key"])

def change():
  # uncomment the below to ensure d actually changes
  # global d
  d = { "key": "post" }

change()

print(d["key"])
```

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -93,16 +111,77 @@ def find_package_manager():
...
+    try:
+        rpm_info = subprocess.check_output(['rpm', '-qi', pkg], universal_newlines=True,\
+                                            stderr=subprocess.STDOUT)
+        if ("package {0} is not installed".format(pkg) in rpm_info):
+            # didn't find package
+            return None
+        rpm_lines = rpm_info.split('\n')
+        for line in rpm_lines:
+            # parse line
+            # note: parsing is weird bc rpm puts various information into two columns
```
hestolz, line 159: JFYI, the logic looks fine in any case, but the output looks different on different distros evidently, e.g. my centos7 machine:

```
[root@hestolz-centos-arc situhan]# rpm -qi azuremonitoragent
Name        : azuremonitoragent
Version     : 1.22.1
Release     : build.master.443
Architecture: x86_64
Install Date: Fri Sep  9 08:28:49 2022
Group       : Productivity/Monitoring
Size        : 112077757
License     : MICROSOFT SOFTWARE LICENSE TERMS
Signature   : RSA/SHA256, Fri Sep  2 20:29:43 2022, Key ID eb3e94adbe1229cf
Source RPM  : azuremonitoragent-1.22.1-build.master.443.src.rpm
Build Date  : Fri Sep  2 19:27:24 2022
Build Host  : 45ac5e19c8d9
Relocations : (not relocatable)
URL         : https://aka.ms/azuremonitor
Summary     : Azure Monitor Agent daemon
Description :
Azure Monitor Agent daemon
```

Lets sanity check that there aren't any issues parsing rpm or dpkg output depending on the version of the package manager itself

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -93,16 +111,77 @@ def find_package_manager():
...
+                if (name != pkg):
+                    # wrong package
+                    return None
+            if (parsed_line[0] == 'Version'):
+                # ['Version', ':', version, 'Vendor', ':', 'MSFT']
+                version = parsed_line[2]
+                general_info['{0}_VERSION'.format(pkg.upper())] = version
+                return version
+        return None
+    except subprocess.CalledProcessError:
```
hestolz, line 173: general comment, silently eating errors might not be the appropriate approach
while i understand not wanting to show the user too many technical details that may confuse them, if we don't have any exception output, if there are issues running the TST there is nothing the customer can share with us to help debug (and reproing the issue may be hard for us). it might be better to log these like `Internal exception while <description of function purpose>: <insert exception details>.`

AzureMonitorAgent/ama_tst/modules/install/check_pkgs.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+
+def check_packages():
+    # check azuremonitoragent rpm/dpkg
+    ama_vers = find_ama_version()
+    if (ama_vers == None):
+        return ERR_AMA_INSTALL
+    if (len(ama_vers) > 1):
+        return ERR_MULTIPLE_AMA
```
hestolz, line 15: Are we sure that multiple AMA folders in `/var/lib/waagent` is necessarily an error/indicative of an issue? Might be good to just check what folders/files are there in `/var/lib/waagent` after an upgrade of AMA versions -- the older folder or .zip may still be there.

AzureMonitorAgent/ama_tst/modules/install/check_os.py
peizhen-w: TST is bundled with AMA at this time, so I assume TST will always check os that is supported in this AMA version. Under which condition customers may run the latest TST for older versions of AMA?

AzureMonitorAgent/ama_tst/modules/install/install.py
peizhen-w: I believe this will be covered in Module 2 - onboarding

AzureMonitorAgent/ama_tst/modules/install/check_ama.py
peizhen-w: Yes, connection check will be covered in the next module. The case that users cannot connect to docs.microsoft.com will be included.

AzureMonitorAgent/ama_tst/modules/install/check_os.py
peizhen-w: I don't think our documentation or somewhere specifies the range of distros that one AMA version supports. I think a good way is to add a warning message that older AMA versions may not support but I can't tell if it actually supports or not in TST. Is simply adding a warning message an acceptable solution?

AzureMonitorAgent/ama_tst/modules/install/check_pkgs.py
peizhen-w: Ashley mentioned it's possible that users have multiple AMA folders in /var/lib/waagent, so I think it's necessary to have this change

AzureMonitorAgent/ama_tst/modules/main.py
peizhen-w: Can I add flag option in a separate PR?

AzureMonitorAgent/ama_tst/modules/helpers.py
peizhen-w: I checked a few distros and it looks like the latest rpm/dpkg output can be parsed in this way. If the output cannot be parsed it will be displayed in errors. For sanity check, is there any other checks I can do before parsing?

AzureMonitorAgent/ama_tst/modules/main.py
hestolz: Sure, no need to block this PR on this. Can be a future enhancement

AzureMonitorAgent/ama_tst/modules/install/install.py
hestolz: My mistake, thanks for clarifying

AzureMonitorAgent/ama_tst/modules/helpers.py
hestolz: I think its fine. If we checked on a few distros and its generally consistent and parseable and any errors would get surfaced should be sufficient.

AzureMonitorAgent/ama_tst/modules/errors.py
```diff
@@ -45,8 +45,8 @@
     ERR_PKG_MANAGER : "This system does not have a supported package manager. Please install 'dpkg' or 'rpm' "\
           "and run this troubleshooter again.",
     ERR_MULTIPLE_AMA : "There is more than one instance of AMA installed, please remove the extra AMA packages.",
-    ERR_AMA_INSTALL : "AMA package isn't installed correctly.",
-    ERR_SUBMODULE_INSTALL : "Submodule(s) {0} not installed correctly.",
+    ERR_AMA_INSTALL : "AMA package isn't installed correctly.\n\nError Details: \n{0}",
```
hestolz, line 48: Seems like in not all cases will the error details will be filled; e.g. `check_pkgs.py` line 13

peizhen-w: I checked the exceptions that should not be passed and display those error messages in the latest version.

AzureMonitorAgent/ama_tst/modules/install/check_os.py
peizhen-w: Since this is an unlikely scenario, can we keep this check as it is now? In the next part we check AMA version and suggest updating to the latest version to avoid any problems. I'm afraid we'll confuse users saying that current TST cannot check older AMA supported distros

hestolz (approved):

AzureMonitorAgent/ama_tst/modules/install/check_os.py
hestolz: yep agreed that it is pretty unlikely

PR #1585, opened 2022-08-07
Log Daemon reload status

simathih:


LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -309,7 +309,12 @@ def setup_me_service(is_lad, configFolder, monitoringAccount, metrics_ext_bin, m
             os.system(r"sed -i 's+%ME_MONITORING_ACCOUNT%+{1}+' {0}".format(me_service_path, monitoringAccount))
             daemon_reload_status = os.system("sudo systemctl daemon-reload")
             if daemon_reload_status != 0:
-                raise Exception("Unable to reload systemd after ME service file change. Failed to set up ME service.")
+                message = "Unable to reload systemd after ME service file change. Failed to set up ME service."
+                if HUtilObj is not None:
+                    HUtilObj.log(message)
```
hestolz, line 314: Do we necessarily want to relax this? If in current code we are throwing the exception, I'm assuming this causes an install/enable failure. How does this tie into our discussion of whether or not to strictly report install/enable failures in LAD vs. being lenient in order to not block provisioning? Same for AMA, since this is shared code.

simathih: for this specific error, I have confirmed that the daemon restarts fine regardless, we are just not sure why we get a non 0 code, so we can ignore this for now as there is no real impact

PR #1599, opened 2022-09-16
Billing logic changes to support selective disk backup

arisettisanjana:
As the Previous logic of extensions did not support for selective billing of disks, now as this feature has been supported the extensions logic should be excluding the disks which customer don't want the backup of and has to return the totalused size.

Logic
As the exclusion of disks by the customer is done by the LUN number
Have mapped the Lun number to the corresponding parent disk
Then got the partitions that are part of the parent disk that is/are to be excluded. (i.e Computation of excluded list)
Finally added the space used by the disk only if the device to be billed is not present in the excluded list.


VMBackup/main/common.py
```diff
@@ -42,7 +42,6 @@ class CommonVariables:
     statusBlobUploadError = 'statusBlobUploadError'
     TempStatusFileName = 'tempStatusFile.status'
     onlyLocalFilesystems = 'onlyLocalFilesystems'
-
```
iamashwani-who, line 45: There seems to be no change in the file. I think this file changes can be reverted.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -46,10 +49,21 @@ def __init__(self,patching, hutil, logger,para_parser):
             errMsg = 'Failed to serialize customSettings with error: %s, stack trace: %s' % (str(e), traceback.format_exc())
             self.logger.log(errMsg, True, 'Error')
             self.isOnlyOSDiskBackupEnabled = False
+        self.command="sudo lsscsi"
+        self.dict1={}
```
iamashwani-who, line 53: Use more descriptive variable name than dict1, lun1, ti, lines1.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -46,10 +49,21 @@ def __init__(self,patching, hutil, logger,para_parser):
...
+        self.lun1=[]
+        self.p=(os.popen(self.command).read())
```
iamashwani-who, line 55: should this be wrapped in a try-catch block in case we get a failure? Please check if we should continue to follow older logic in case of exception.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +79,28 @@ def device_list_for_billing(self):
...
         device_items = disk_util.get_device_items(None)
         for device_item in device_items :
-            # self.logger.log("Device name : {0} ".format(str(device_item.name)),True)
+            self.logger.log("Device name : {0} ".format(str(device_item.name)),True)
             if str(device_item.name).startswith("sd"):
                 devices_to_bill.append("/dev/{0}".format(str(device_item.name)))
+        self.logger.log("Initial billing items {0}".format(devices_to_bill))
+        self.logger.log("lines {0}".format(self.lines1))
+        for i in self.lines1:
+            lun=int(i[i.index(']')-1])
```
iamashwani-who, line 88: lun number be more than 1 digit. in that case, will this logic work?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +79,28 @@ def device_list_for_billing(self):
...
+                #list containing the elements present in the row of the cmd sudo lsscsi
+                self.dvc=self.d[len(self.d)-1]#storing the corresponding device name
+                if lun not in self.dict1.keys():
+                    self.dict1[lun]=[]
+                self.dict1[lun].append(self.dvc)
+        self.logger.log("Lun numbers to be excluded {0}".format(self.dict1.keys()))
+        self.logger.log("Disks to be excluded {0}".format(self.dict1.values()))
+        for k in self.dict1.keys():
+            for v in self.dict1[k]:
+                for z in devices_to_bill:
```
iamashwani-who, line 100: It might be better to take all values in self.dict1 into a hashtable and then loop items from devices_to_bill to check if it exists in the hashtable?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -212,19 +244,20 @@ def get_total_used_size(self):
...
                         if(mountpoint == '/'):
                             total_used = total_used + int(used)
                             self.logger.log("Adding only root device to size calculation. Device name : {0} used space in KB : {1} mount point : {2} fstype : {3}".format(device,used,mountpoint,fstype),True)
                             self.logger.log("Total Used Space: {0}".format(total_used),True)
                     else:
-                        self.logger.log("Adding Device name : {0} used space in KB : {1} mount point : {2} fstype : {3}".format(device,used,mountpoint,fstype),True)
-                        total_used = total_used + int(used) #return in KB
+                        if device not in self.ti :
+                            self.logger.log("Adding Device name : {0} used space in KB : {1} mount point : {2} fstype : {3}".format(device,used,mountpoint,fstype),True)
+                            total_used = total_used + int(used) #return in KB
```
iamashwani-who, line 256: add an else block to log that device was found in self.ti

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +85,33 @@ def device_list_for_billing(self):
...
         device_items = disk_util.get_device_items(None)
         for device_item in device_items :
-            # self.logger.log("Device name : {0} ".format(str(device_item.name)),True)
+            self.logger.log("Device name : {0} ".format(str(device_item.name)),True)
             if str(device_item.name).startswith("sd"):
                 devices_to_bill.append("/dev/{0}".format(str(device_item.name)))
+        self.logger.log("Initial billing items {0}".format(devices_to_bill))
+        self.logger.log("lsscsi_list {0}".format(self.lsscsi_list))
+        if(len(self.lsscsi_list)!=0):
+            for i in self.lsscsi_list:
```
ankitha-p, line 94: use meaningful variable names for indexing the list

ankitha-p: Also, add a comment of how an element in the list looks like as a sample, to be clear why we are finding : or ] and selecting substring.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +85,33 @@ def device_list_for_billing(self):
...
+                self.included=True
+                idxOfColon=i.rindex(':',0,i.index(']'))#to get the index of last ':'
+                idxOfColon+=1
+                lunNumbers=int(i[idxOfColon:i.index(']')])
+                self.temp=i.split()
```
ankitha-p, line 99: NIT: In general, it's a good practice to use a space before and after using binary operators such as '='. Across the code

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +85,33 @@ def device_list_for_billing(self):
...
+                #list containing the elements present in the row of the cmd sudo lsscsi
+                self.device=self.temp[len(self.temp)-1]#storing the corresponding device name
```
ankitha-p, line 101: Is it possible for the length of temp to be 0?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -140,6 +186,7 @@ def get_total_used_size(self):
             index = 1
             self.resource_disk= ResourceDiskUtil(patching = self.patching, logger = self.logger)
```
ankitha-p, line 187: Let's add a comment that ResourceDisks are excluded in billing as they are similar to temporary storage

VMBackup/main/Utils/SizeCalculation.py
arisettisanjana: As temp represents a row in lsscsi command it cannot be zero

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +85,47 @@ def device_list_for_billing(self):
...
+                    self.disksToBeExcluded.append(self.device_name)
+                self.logger.log("LUN Number {0}, disk {1}".format(lunNumber,self.device_name))   
+            self.logger.log("Disks to be excluded {0}".format(self.disksToBeExcluded))
+           
+            for disk in self.disksToBeExcluded:
+                for device in devices_to_bill:
+                    if disk in device:
+                        self.devicesToExclude.append(device)
+
+            for device in self.root_devices:
```
ankitha-p, line 124: Let's add a comment that, we are checking if OS disk is included in the Lun List, if yes, we remove it from devicesToExclude. (as Linux is by default assigning LUN 0 to root)

ankitha-p: Let's add a log statement for this as well.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +85,47 @@ def device_list_for_billing(self):
...
         device_items = disk_util.get_device_items(None)
         for device_item in device_items :
-            # self.logger.log("Device name : {0} ".format(str(device_item.name)),True)
             if str(device_item.name).startswith("sd"):
                 devices_to_bill.append("/dev/{0}".format(str(device_item.name)))
+            else:
+                self.logger.log("Not adding device {0} as it does not start with sd".format(str(device_item.name)))
+        self.logger.log("Initial billing items {0}".format(devices_to_bill))
+        
+        for file_system in self.file_systems_info:
```
ankitha-p, line 94: Let's add a comment to show how rows are there in file_systems_info: like you did for lsscsi in L101. It will create clarity asto why we are taking item at index 2 to add to root_devices.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -218,13 +279,16 @@ def get_total_used_size(self):
...
+                            self.logger.log("Adding Device name : {0} for billing used space in KB : {1} mount point : {2} fstype : {3}".format(device,used,mountpoint,fstype),True)
+                            total_used = total_used + int(used) #return in KB
+                        else:
+                            self.logger.log("Device {0} is not included for billing used space in KB : {1} mount point : {2} fstype : {3}".format(device,used,mountpoint,fstype),True)
+                            excluded_disks_used = excluded_disks_used + int(used)
                     if not (isKnownFs or fstype == '' or fstype == None):
                         total_used_unknown_fs = total_used_unknown_fs + int(used)
 
                 index = index + 1
-
```
ankitha-p, line 227: The new line can remain as it indicates a new block of code outside while loop.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -46,10 +47,29 @@ def __init__(self,patching, hutil, logger,para_parser):
...
+        # The command lsscsi is used for mapping the LUN numbers to the disk_names
+        self.command = "sudo lsscsi"
+        self.disksToBeExcluded = []
+        self.root_devices = []
+        self.root_mount_points = ['/' , '/boot/efi']
+        try:
+            self.output_lsscsi = (os.popen(self.command).read())
+            self.lsscsi_list = self.output_lsscsi.splitlines()
+        except Exception as e:
+            error_msg = " Failed to execute the command lsscsi because of error %s , stack trace: %s" % (str(e), traceback.format_exc())
```
iamashwani-who, line 59: remove space at beginning of error message.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -245,7 +309,10 @@ def get_total_used_size(self):
                 Utils.HandlerUtil.HandlerUtility.add_to_telemetery_data("ramDisksSize",str(total_used_ram_disks))
             if total_used_loop_device != 0 :
                 Utils.HandlerUtil.HandlerUtility.add_to_telemetery_data("loopDevicesSize",str(total_used_loop_device))
-            self.logger.log("Total used space in Bytes : {0}".format(total_used * 1024),True)
+            self.logger.log("TotalUsedSpaceAfterExcludeLUN in Bytes : {0}".format(total_used * 1024),True)
+            self.logger.log("TotalLUNExcludedUsedSpace in Bytes : {0}".format(excluded_disks_used *1024),True)
```
iamashwani-who, line 313: Can this information be logged in same format as the one used by windows extension?

iamashwani-who (dismissed):

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -65,9 +85,55 @@ def device_list_for_billing(self):
...
+                self.root_devices.append(file_system[0])
+        self.logger.log("root_devices {0}".format(str(self.root_devices)))
+        
+        self.logger.log("lsscsi_list {0}".format(self.lsscsi_list))
+        '''
+            Sample output of the lsscsi command 
+            [1:0:0:15]   disk    Msft     Virtual Disk     1.0   /dev/sda
+            [1:0:0:18]   disk    Msft     Virtual Disk     1.0   /dev/sdc
+        '''
+        if(len(self.lsscsi_list) != 0):
```
iamashwani-who, line 111: nit: this entire block can be potentially a separate function like parselsscsioutput.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -60,14 +86,89 @@ def get_loop_devices(self):
...
+    def disk_list_for_billing(self):
+        if(len(self.lsscsi_list) != 0):
+            for item in self.lsscsi_list:
+                idxOfColon = item.rindex(':',0,item.index(']'))# to get the index of last ':'
+                idxOfColon += 1
+                lunNumber = int(item[idxOfColon:item.index(']')])
+                # item_split is the list of elements present in the one row of the cmd sudo lsscsi
+                self.item_split = item.split()
+                #storing the corresponding device name from the list
+                self.device_name = self.item_split[len(self.item_split)-1]
```
rakrashba, line 98: using a property defined on `self` to do the work of a local variable will leave `self` dirty and using memory that should be freed at the end of the loop. Please try and use a local variable for this, instead of `self.device_name`, that goes out of scope at the end of the loop.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -46,10 +47,35 @@ def __init__(self,patching, hutil, logger,para_parser):
...
+        try:
+            self.output_lsscsi = (os.popen(self.command).read())
+            self.lsscsi_list = self.output_lsscsi.splitlines()
+        except Exception as e:
+            error_msg = "Failed to execute the command lsscsi because of error %s , stack trace: %s" % (str(e), traceback.format_exc())
+            self.logger.log(error_msg, True ,'Error')
+            self.output_lsscsi = ""
+            self.lsscsi_list = []
+        try:
+             self.output_lsblk = json.loads(os.popen("lsblk --json").read())
```
rakrashba, line 64: [SUGGESTION] - Since all thats used from the output of this command below ( lines 159-169) are the keys "name" and "mountpoint" - maybe use `lsblk --json --output name,mountpoint` instead. Should reduce the memory impact of storing the output of this by a few times. Include more keys when needed.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -60,14 +86,89 @@ def get_loop_devices(self):
...
+
+                for device in self.root_devices :
+                    if self.device_name in device :
+                        lunNumber = -1
+                        # Changing the Lun# of OS Disk to -1
+
+                if lunNumber in self.includedLunList :
+                    self.disksToBeIncluded.append(self.device_name)
+                self.logger.log("LUN Number {0}, disk {1}".format(lunNumber,self.device_name))   
+            self.logger.log("Disks to be included {0}".format(self.disksToBeIncluded))
```
vityagi, line 108: @arisettisanjana can you please share a sample extension trace file with these changes, will make it easier to correlate the command outputs and review

arisettisanjana: [Extension_PR_log.txt](https://github.com/Azure/azure-linux-extensions/files/9763942/Extension_PR_log.txt)

vityagi (approved):
Looks fine

iamashwani-who (dismissed):

vityagi (approved):
Looks fine

PR #1604, opened 2022-10-05
Fixes ExtensionUtils.run_command_get_output()

D1v38om83r:
ExtensionUtils.run_command_get_output() now decodes the output of commands into a utf-8 string.

Fixes VMAccess extension's inability to accept non latin characters as password.


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
pavanrachapudy, line 1: this file needed to be there?

Utils/extensionutils.py
```diff
@@ -200,7 +200,7 @@ def run_command_get_output(cmd, chk_err=True, log_cmd=True):
                 'CalledProcessError.  Error message is ' + str(e))
             return e.errno, str(e)
     # noinspection PyUnboundLocalVariable
-    return 0, output.decode('latin-1')
+    return 0, output.decode('utf-8')
```
pavanrachapudy, line 203: looks like other method(s) has this usage as well - run_send_stdin. I remember we were hitting that method for vmaccess. Also, in the current method too, it is still using latin-1? Shouldn't we handle these as well?

D1v38om83r: updated all output.decode to utf-8

.vscode/launch.json
D1v38om83r: makes running unit tests easier

jscalev (approved):

axelgMS:
thank you

PR #1607, opened 2022-10-10
AMA CIS hardening

hestolz:
Changes to ensure that Azure CIS-hardened image with AMA deployed still meets the pre-AMA benchmark score and does not cause the system to become out of compliance with the CIS benchmark. In this case, ME was listening on all interfaces, because if you don't specify the host, it will listen on 0.0.0.0. This was causing us to fail a benchmark rule requiring non-local ports to have iptables rules associated. With this change we ensure ME is only opening a port on localhost interface.

- Set `InfluxDbHost` to prevent CIS benchmark violation `3.4.3.2.3 Ensure iptables rules exist for all open ports`

### DIFF
--- a/Diagnostic/services/metrics-extension.service
+++ b/Diagnostic/services/metrics-extension.service
@@ -3,7 +3,7 @@ Description=Metrics Extension service for Linux Agent metrics sourcing
 After=network.target
 
 [Service]
-ExecStart=%ME_BIN% -TokenSource MSI -Input influxdb_udp -InfluxDbUdpPort %ME_INFLUX_PORT% -DataDirectory %ME_DATA_DIRECTORY% -LocalControlChannel -MonitoringAccount %ME_MONITORING_ACCOUNT% -LogLevel Error
+ExecStart=%ME_BIN% -TokenSource MSI -Input influxdb_udp -InfluxDbHost 127.0.0.1 -InfluxDbUdpPort %ME_INFLUX_PORT% -DataDirectory %ME_DATA_DIRECTORY% -LocalControlChannel -MonitoringAccount %ME_MONITORING_ACCOUNT% -LogLevel Error
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=on-failure
 RestartForceExitStatus=SIGPIPE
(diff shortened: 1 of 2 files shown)


Diagnostic/services/metrics-extension.service
```diff
@@ -3,7 +3,7 @@ Description=Metrics Extension service for Linux Agent metrics sourcing
 After=network.target
 
 [Service]
-ExecStart=%ME_BIN% -TokenSource MSI -Input influxdb_udp -InfluxDbUdpPort %ME_INFLUX_PORT% -DataDirectory %ME_DATA_DIRECTORY% -LocalControlChannel -MonitoringAccount %ME_MONITORING_ACCOUNT% -LogLevel Error
```
NarineM, line 6: What was the error happening before the fix ? Do we have quick tsg for the particular error to look for when investigating that is being resolved with this change? since this fix will make to only newer versions of AMA extension.

hestolz: Updated the description -- this PR is not addressing any AMA functionality issues on CIS-hardened images (none have been found). Rather, it is making changes to ensure that AMA installation does not cause the system to become out of compliance with the CIS benchmark. In this case, ME was listening on all interfaces, because if you don't specify the host, it will listen on 0.0.0.0. With this change now we ensure ME is only opening a port on localhost interface.

simathih (approved):

NarineM (approved):

Diagnostic/services/metrics-extension.service
NarineM: Thanks Henry for details. Signed off.

AzureMonitorAgent/agent.py
```diff
@@ -1146,8 +1146,22 @@ def find_package_manager(operation):
     for rpm_dist in rpm_set:
         if dist.startswith(rpm_dist):
             PackageManager = "rpm"
-            # Same as above.
+            # Same as above; need to force.
             PackageManagerOptions = "--force"
+
+            # If FIPS is enabled and RPM version is 4.14 or higher, we need to apply the below workaround
+            try:
```
NarineM, line 1153: I would suggest to have fips in seperate commit and PR as it is easy to revert or look back in history if issues happen with the changes. Also CIS benchmark related change seems unrelated.

hestolz: Sure, I have added as a separate PR #1611

simathih (approved):

PR #1611, opened 2022-10-19
AMA Install flag for FIPS RPM

hestolz:
Allow for FIPS+RHEL8 scenario

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -1146,8 +1146,22 @@ def find_package_manager(operation):
     for rpm_dist in rpm_set:
         if dist.startswith(rpm_dist):
             PackageManager = "rpm"
-            # Same as above.
+            # Same as above; need to force.
             PackageManagerOptions = "--force"
+
+            # If FIPS is enabled and RPM version is 4.14 or higher, we need to apply the below workaround
+            try:
+                fips_code, _ = run_command_and_log('grep -i "fips=1" /proc/cmdline 1>/dev/null 2>&1', check_error=False, log_cmd=False)
+                if fips_code == 0:
+                    rpm_code, rpm_output = run_command_and_log('rpm --version', check_error=False, log_cmd=False)
+                    if rpm_code == 0:
+                        rpm_version = (rpm_output.split(" ")[-1]).split(".")
+                        if int(rpm_version[0]) >= 4 and int(rpm_version[1]) >= 14:
+                            hutil_log_info("FIPS is enabled and RPM version is at least 4.14; applying --nodigest flag to allow for successful package install")
+                            PackageManagerOptions = "--force --nodigest --nofiledigest"
+            except Exception as e:
+                hutil_log_error("Unable to check for RPM 4.14 + FIPS scenario, skipping digest workaround: {0}".format(e))
+
             BundleFileName = BundleFileNameRpm
             break
 


simathih (approved):

AzureMonitorAgent/agent.py
```diff
@@ -1146,8 +1146,22 @@ def find_package_manager(operation):
...
+            # If FIPS is enabled and RPM version is 4.14 or higher, we need to apply the below workaround
+            try:
+                fips_code, _ = run_command_and_log('grep -i "fips=1" /proc/cmdline 1>/dev/null 2>&1', check_error=False, log_cmd=False)
+                if fips_code == 0:
+                    rpm_code, rpm_output = run_command_and_log('rpm --version', check_error=False, log_cmd=False)
+                    if rpm_code == 0:
+                        rpm_version = (rpm_output.split(" ")[-1]).split(".")
+                        if int(rpm_version[0]) >= 4 and int(rpm_version[1]) >= 14:
+                            hutil_log_info("FIPS is enabled and RPM version is at least 4.14; applying --nodigest flag to allow for successful package install")
+                            PackageManagerOptions = "--force --nodigest --nofiledigest"
```
NarineM, line 1161: No need to set --force. I believe it is only needed when already existing version of package needs to be rewritten or there are missing dependency packages on the machine but you still want to install main package.

hestolz: it is already the existing behavior (see above). so i am leaving as-is

AzureMonitorAgent/agent.py
```diff
@@ -1146,8 +1146,22 @@ def find_package_manager(operation):
...
+            # Same as above; need to force.
             PackageManagerOptions = "--force"
+
+            # If FIPS is enabled and RPM version is 4.14 or higher, we need to apply the below workaround
+            try:
+                fips_code, _ = run_command_and_log('grep -i "fips=1" /proc/cmdline 1>/dev/null 2>&1', check_error=False, log_cmd=False)
+                if fips_code == 0:
+                    rpm_code, rpm_output = run_command_and_log('rpm --version', check_error=False, log_cmd=False)
+                    if rpm_code == 0:
+                        rpm_version = (rpm_output.split(" ")[-1]).split(".")
```
NarineM, line 1158: If fips is enabled on rpm where version is smaller than 4.14, digest flag will be still needed to install the package. I think you can just remove the version check entirely. What are you trying to accomplish with the check?

AzureMonitorAgent/agent.py
```diff
@@ -1146,8 +1146,22 @@ def find_package_manager(operation):
     for rpm_dist in rpm_set:
         if dist.startswith(rpm_dist):
             PackageManager = "rpm"
-            # Same as above.
+            # Same as above; need to force.
             PackageManagerOptions = "--force"
+
+            # If FIPS is enabled and RPM version is 4.14 or higher, we need to apply the below workaround
+            try:
+                fips_code, _ = run_command_and_log('grep -i "fips=1" /proc/cmdline 1>/dev/null 2>&1', check_error=False, log_cmd=False)
```
NarineM, line 1154: does proc check work with older rpms? redhat 6 and 7? and newer onces like 8.*?

NarineM: I think it is good to confirm it returns the right enable flag for fips for each of rel families: 1 OS 6.*, 1 OS 7.* and 1 OS 8.*. I remember reading somewhere that the enabled flag for fips is not always returning the correct value.

AzureMonitorAgent/agent.py
hestolz: Per my understanding based on RPM forum posts and my personal testing RPM <4.14 does _not_ verify the digests, so we don't need the flags in that case. We should make scope of the flag as narrow as possible.

AzureMonitorAgent/agent.py
hestolz: Confirmed that RHEL7 with FIPS enabled has `fips=1` in `/proc/cmdline` (also confirmed that AMA will install on RHEL7 w/FIPS with no `--nodigest` flag). RHEL6 series doesn't matter since we don't support.

NarineM (approved):

AzureMonitorAgent/agent.py
NarineM: I still think that is less hassle and less time for installer if we skip digest validation just based on the enable flag. but signed off since you confirmed that with fips enabled rpm 4.14 do not check the digest during install.

AzureMonitorAgent/agent.py
NarineM: Do you want to hold on this change for now? Just realized that AMA does not support Centos 6 installation. In case of OMS upgrading rpm build to >4.14 means dropping Cent os/ rel 6 install which does not seem to be the case with AMA.

hestolz: I'm not sure how there's any interaction/relation between what we do with OMS and what we do with AMA here.

NarineM: In OMS we went with noDigest flag instead of creating FIPS compliant package for 2 reasons:
1.  Updating oms build system to centos 7 would mean dropping customers who are on redhat 6 or centos 6. In case of AMA , it is not an issue as these distros never been supported by AMA nor there are AMA customers that will break.
2.  OMS agent is on deprecation path

In case of AMA,  from FIPs compliance point of view, it might not be acceptable to customers to do same with new agent given #1 is not an issue for AMA.  Please follow up with PM and Dmitry to confirm they approve this decision. 

@Dmitry-Matveev

PR #1617, opened 2022-10-26
Billing fixes

arisettisanjana:
Updated the condition (Previously used device names) for including the disks to mount points.
As there was the difference in device names provided by df vs lsblk output, so using the mountpoints for comparison.


mearvind (approved):

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -166,8 +167,10 @@ def device_list_for_billing(self):
                                     if "mountpoint" in child.keys() and child["mountpoint"] != None :
                                         child["name"] = '/dev/' + child["name"]
                                         self.devicesToInclude.append(child["name"])
+                                        self.device_mount_points.append(child["mountpoint"])
```
iamashwani-who, line 170: Is there a possibility of chind["mountpoint"] having a null or blank value? If yes, then that can that cause duplicate records to be added in device_mount_points?

arisettisanjana: if "mountpoint" in child.keys() and child["mountpoint"] != None :
This condition possibly eliminates the cases where mountpoint is null or blank.
So, they will not be included in device_mount_points.

iamashwani-who (approved):

vityagi (approved):

PR #1618, opened 2022-10-26
Non blocking errors bug fix

arisettisanjana:
Added checks to avoid errors, for files that get MachineIdentity


vityagi (approved):
Looks fine

VMBackup/main/MachineIdentity.py
```diff
@@ -26,17 +26,21 @@ def __init__(self):
...
-        deployment = dom.getElementsByTagName("Role")
-        identity=deployment[0].getAttribute("guid")
+        identity = None
+        if os.path.exists("/var/lib/waagent/HostingEnvironmentConfig.xml"):
+            file = open("/var/lib/waagent/HostingEnvironmentConfig.xml",'r')
+            xmlText = file.read()
+            dom = xml.dom.minidom.parseString(xmlText)
+            deployment = dom.getElementsByTagName("Role")
+            identity=deployment[0].getAttribute("guid")
+            file.close()
```
iamashwani-who, line 36: file.close() should be in finally block to handle for cases when file open/read fails?

VMBackup/main/MachineIdentity.py
```diff
@@ -26,17 +26,21 @@ def __init__(self):
...
         return identity
 
     def save_identity(self):
         file = open(self.store_identity_file,'w')
         machine_identity = self.current_identity()
-        file.write(machine_identity)
+        if( machine_identity != None ):
+            file.write(machine_identity)
         file.close()
```
iamashwani-who, line 44: we can improve the current code to move file.close() to finally to handle error scenarios.

VMBackup/main/handle.py
```diff
@@ -241,15 +241,19 @@ def daemon():
...
+            else:
+                current_identity = mi.current_identity()
+                if(current_identity != stored_identity):
+                    current_seq_no = -1
+                    backup_logger.log("machine identity not same, set current_seq_no to " + str(current_seq_no) + " " + str(stored_identity) + " " + str(current_identity), True)
+                    hutil.set_last_seq(current_seq_no)
+                    mi.save_identity()
+        except Exception as e:
+            errMsg = "Unable to open file, error: %s, stack trace: %s" % (str(e), traceback.format_exc())
+            backup_logger.log(errMsg, True, 'Error')
```
iamashwani-who, line 256: Will this extra Exception handling block be required given that there is another exception handling block present immediately after.

iamashwani-who (approved):
Looks good.

rakrashba (changes requested):
`VMBackup/main/MachineIdentity.py` - `Line 30`
why `self.file` instead of `file`. It adds no extra benefit and it lives in memory as long as the class instance lives - using more memory than needed. `self.file` as I can see is also being used in two separate methods below for entirely different files. Why?
`Line 45` and `Line 58`

What if the class does need to store an attribute called `file` that DOES NOT change as long as the instance lives. The changes necessary because of the way this is implemented now will make that effort more tricky than it needs to be. 

Also, some one else working on it in the future might mistakenly assume that `self.file` is indeed a class level instance for a valid reason and waste time trying to understand why?

Please do not use the class level scope for variables that do not need class level scope.

This includes `self.file_pointer` in `VMBackup/main/Utils/HandlerUtil.py` as well - `Line 406`.

iamashwani-who (approved):
Approved

Issue #1612, opened 2022-10-23
DSC & RHEL 8.x & Python > 3.8 not working 


CoolZeroNL:
DSC extension is not compatible with Python 3.8 and up

```
2022/10/21 10:16:26 [Microsoft.OSTCExtensions.DSCForLinux-3.8.3] Config decoded correctly.
2022/10/21 10:16:26 Falling back to /etc/os-release distribution parsing
2022/10/21 10:16:26 [Microsoft.OSTCExtensions.DSCForLinux-3.8.3] Install,error,51,rhel; distro_version: 8.5 is not supported.
```

The function platform.dist &  platform.linux_distribution are been [removed in python 3.7](https://docs.python.org/3.5/library/platform.html#unix-platforms). So the DSC will use the [Fallback to use /etc/os-release.](https://github.com/Azure/azure-linux-extensions/blob/master/DSC/dsc.py#L158) This is working fine, only that RHEL is using a other name as is then is listed in the suppored_list.

```
# cat /etc/os-release 

NAME="Red Hat Enterprise Linux"
VERSION="8.6 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.6"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.6 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/8/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.6
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.6"
```
The ID in here is `rhel` and this is not listed in the supported_dists list.

Running the function separately shows also this:

```
# python3.6 tmp.py 
red hat-->red hat enterprise linux
['8', '6']

# python3.8 tmp.py 
Falling back to /etc/os-release distribution parsing
Unsupported OS :rhel; distro_version: 8.6
False
rhel
8.6

# python3.9 tmp.py 
Falling back to /etc/os-release distribution parsing
[...]

PR #1624, opened 2022-10-28
Supporting RHEL type in supported_dists to support Python > 3.7

CoolZeroNL:
RHEL 8 is using `rhel` as ID, so adding support for that in the list.

The function platform.dist & platform.linux_distribution are been [removed in python 3.7](https://docs.python.org/3.5/library/platform.html#unix-platforms). So the DSC will use the [Fallback to use /etc/os-release.](https://github.com/Azure/azure-linux-extensions/blob/master/DSC/dsc.py#L158) This is working fine, only that RHEL is using a other name as is then is listed in the suppored_list

```
# cat /etc/os-release 

NAME="Red Hat Enterprise Linux"
VERSION="8.6 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.6"
```

Issue related: https://github.com/Azure/azure-linux-extensions/issues/1612

### DIFF
--- a/DSC/dsc.py
+++ b/DSC/dsc.py
@@ -158,6 +158,7 @@ def check_supported_OS():
     supported_dists = {'redhat' : ['7', '8'], # CentOS
                        'centos' : ['7', '8'], # CentOS
                        'red hat' : ['7', '8'], # Redhat
+                       'rhel' : ['7', '8'], # Redhat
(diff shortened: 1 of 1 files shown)


CoolZeroNL:
@Bhargava-Chary-Chollaty can u review ?

PR #1629, opened 2022-11-15
Set http_proxy for CoreAgent and ME

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -746,6 +749,31 @@ def restart_launcher():
         check_kill_process('/opt/microsoft/azuremonitoragent/bin/fluent-bit')
         exit_code, output = run_command_and_log('systemctl restart azuremonitor-agentlauncher && systemctl enable azuremonitor-agentlauncher')
 
+def set_proxy(address, username, password):
+    """
+    # Set proxy http_proxy env var in dependent services
+    """
+    
+    http_proxy = address
+    address = address.replace("http://","")
```
NarineM, line 758: Do you want to add a try block here if address is not starting with http but rather with fqdn or hostname? It can cause extension crash , I think it is safer to add error handling to wrap default error stack.

simathih: as per AMA contract, it should be http but I will add try catch

NarineM (approved):

PR #1632, opened 2022-11-21
Guest agent snapshot task status error fix

arisettisanjana:
To fix the issue of:
Exception 'NoneType' object is not subscriptable  
self.isAnyDiskExcluded = para_parser.includedDisks[CommonVariables.isAnyDiskExcluded]
Adding try, catch blocks to avoid errors being thrown.
This scenario is being observed in BVT.


VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -66,11 +66,23 @@ def __init__(self,patching, hutil, logger,para_parser):
             self.logger.log(error_msg, True ,'Error')
             self.output_lsblk = {}
         self.devicesToInclude = [] #partitions to be included
-        self.isAnyDiskExcluded = para_parser.includedDisks[CommonVariables.isAnyDiskExcluded]
-        self.includedLunList = para_parser.includeLunList
         self.device_mount_points = []
+        self.isAnyDiskExcluded = False
+        try:
+            if(para_parser.includedDisks != None and para_parser.includedDisks != ''):
+                self.isAnyDiskExcluded = para_parser.includedDisks[CommonVariables.isAnyDiskExcluded]
```
iamashwani-who, line 73: if the key is not present, do we expect an exception? In any case, we can probably put a check if they key exists and the value is not None before assigning to isAnyDiskExcluded.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -66,11 +66,23 @@ def __init__(self,patching, hutil, logger,para_parser):
...
+                self.logger.log("isAnyDiskExcluded {0}".format(self.isAnyDiskExcluded))
+        except Exception as e:
+             error_msg = "Failed to serialize includedDisks because of error %s , stack trace: %s" % (str(e), traceback.format_exc())
+             self.logger.log(error_msg, True ,'Error')    
+        try:
+            if( para_parser.includeLunList != None and para_parser.includeLunList != ''):
+                self.includedLunList = para_parser.includeLunList
```
iamashwani-who, line 80: I don't think this block requires to be wrapped in a try-except block as we are checking None and Empty string.

PR #1635, opened 2022-11-28
Support other.conf files in Ubuntu 2210

norakoiralamsft:
Adding support for Ubuntu 2210, updating PasswordAuthentication to yes in 
/etc/ssh/sshd_config.d/50-cloud-init.conf

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -348,6 +348,13 @@ def _allow_password_auth():
     _set_sshd_config(config, "PasswordAuthentication", "yes")
     ext_utils.replace_file_with_contents_atomic(SshdConfigPath, "\n".join(config))
 
+    if isinstance(MyDistro, dist_utils.UbuntuDistro): #handle ubuntu 22.04 (sshd_config.d directory)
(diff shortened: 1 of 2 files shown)


VMAccess/vmaccess.py
```diff
@@ -360,9 +367,7 @@ def _set_sshd_config(config, name, val):
             # Match block must be put in the end of sshd config
             break
     if notfound:
-        if i is None:
-            i = 0
-        config.insert(i, "{0} {1}".format(name, val))
```
jscalev, line 365: Do we still need this if get_file_contents returns more than one file?

norakoiralamsft: The change was because .insert is used for arrays but config is a string so changed it to concat instead

jscalev (approved):

jscalev (approved):

ani-sinha:
@norakoiralamsft We need support for this for Centos/RHEL systems as well

ani-sinha:
> @norakoiralamsft We need support for this for Centos/RHEL systems as well

I sent a PR : https://github.com/Azure/azure-linux-extensions/pull/1774

PR #1638, opened 2022-12-02
Updated The billing logic to handle Unmanaged VM's and lsblk

arisettisanjana:
Had modified the lsblk  --json command as --json option is not available in all distros.
Also included a condition that helps in handling the case for unmanaged VM's.


VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -60,17 +60,24 @@ def __init__(self,patching, hutil, logger,para_parser):
             self.output_lsscsi = ""
             self.lsscsi_list = []
         try:
-             self.output_lsblk = json.loads(os.popen("lsblk --json --output name,mountpoint").read())
+            self.output_lsblk = os.popen("lsblk -n --list --output NAME,MOUNTPOINT").read().strip().split("\n")
```
iamashwani-who, line 63: Any specific reason for casing changes from name,mountpoint to NAME,MOUNTPOINT?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -137,38 +144,47 @@ def device_list_for_billing(self):
...
+                sdf
+                sdg
+        '''
+        for item in self.output_lsblk:
+            item_split = item.split()
+            if(len(item_split)==2):
+                device = item_split[0]
+                mount_point = item_split[1]
+            else:
+                mount_point = ""
```
iamashwani-who, line 179: can we log a statement that that item split did not return 2 items and hence skipping?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -137,38 +144,47 @@ def device_list_for_billing(self):
...
+                sdc1          /mnt/sdc1
+                sdc2          /mnt/sdc2
+                sdc3
+                sde
+                sde1          /mnt/sde1
+                sdf
+                sdg
+        '''
+        for item in self.output_lsblk:
+            item_split = item.split()
```
iamashwani-who, line 174: Should be reset the values of mount_point and device at the top of the loop?

VMBackup/main/Utils/SizeCalculation.py
arisettisanjana: might not need it.

iamashwani-who (approved):

rakrashba (changes requested):

PR #1640, opened 2022-12-06
AMA extension error handling and error reporting improvements

NarineM:
Changes in the PR:
1. Removed magic number 52.
2. Added errors from package manager corrupt  paths to be excluded from SLA
3. Removed stderr logging for each command line execution based on guest agent guidance on how this stream is used,. Should help with improving overall time execution of agent.py as path was being called for every command in each extension operation.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -119,6 +119,7 @@ UnsupportedOperatingSystem = 51
 IndeterminateOperatingSystem = 51
 MissingorInvalidParameterErrorCode = 53
 DPKGOrRPMLockedErrorCode = 56
+MissingDependency = 52
 
 # Settings
 GenevaConfigKey = "genevaConfiguration"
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -1424,19 +1424,17 @@ def run_command_and_log(cmd, check_error = True, log_cmd = True):
...
-            exit_code = 52
-
-    except:
-        hutil_log_info('Failed to write output to STDERR')
+    if "cannot open Packages database" in output:
+        # Install failures
+        # External issue. Package manager db is either corrupt or needs cleanup
+        # https://github.com/Azure/azure-marketplace/wiki/Extension-Build-Notes-Best-Practices#error-codes-and-messages-output-to-stderr
+        exit_code = MissingDependency
+        output += "Package manager database is in a bad state. Please recover package manager, db cache and try install again later."
```
simathih, line 1432: nit: comma is not needed right?

simathih (approved):

AzureMonitorAgent/agent.py
NarineM: yes, comma not needed.

simathih (approved):

PR #1642, opened 2022-12-08
Temporary fix for 'Kernel Too Old' issue in old Linux Distros

ankitha-p:
1. Added the safefreeze binary directly to the repo from older version (9188) to enable support for Old linux distros as we do not make changes to the safefreeze logic frequently. 
2. Updated the bin directory name in makefile, which will not overwrite the build pipeline changes.


VMBackup/main/safefreeze/Makefile
```diff
@@ -1,7 +1,9 @@
 CC := gcc
 SRCDIR := src
 LIBDIR := lib
-BINDIR := bin
+BINDIRNEW := binNew
+BINDIROLD := bin
+BINDIR := BINDIRNEW
```
rakrashba, line 6: this will probably need to be `BINDIR = $(BINDIRNEW)`

rakrashba (changes requested):

iamashwani-who (approved):

rakrashba (approved):
Looks good

PR #1650, opened 2022-12-19
fix python syntax warning in waagent

NarineM:
Fix python syntax warning found in AMA Install telemetry showing that agent is logging in standard error stream.
Example from AMA install telemetry:

[stderr]
Running scope as unit: install_52ef8942-a5b1-459d-b201-5ca1bfb45c60.scope
/var/lib/waagent/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent-1.24.2/waagent:2578: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if me.returncode is not 0 and chk_err is True and log_cmd:


Dmitry-Matveev (approved):

NarineM:
@nkuchta  can you review this change? I need owner approval to check in

D1v38om83r (approved):

PR #1651, opened 2022-12-19
Exit early on AMA install error

NarineM:


AzureMonitorAgent/agent.py
```diff
@@ -322,6 +322,9 @@ def install():
                                          retry_check = retry_if_dpkg_or_rpm_locked,
                                          final_check = final_check_if_dpkg_or_rpm_locked)
 
+    if exit_code != 0:
+        return exit_code, output
```
hestolz, line 326: there is logic below that seems to imply in rhel8.6 AARCH64 scenario the first install may report failure. @thperapp will this addition (early exit) break that scenario

NarineM: Do we need anything from copy_amacoreagent_binaries() for rel hat error case? I am thinking to move this line to be after rel check.

NarineM: Updated to exclude aarch64. Do not want to make the code more complicated, and  added check for aarch64.

thperapp (approved):

PR #1655, opened 2022-12-30
Billing with logical volumes

arisettisanjana:
Made the flow not to be dependent on lsscsi list when all disks are being included (no exclusion of disk).
Reporting the flag ( isSizeComputationFailed) as true, when the distro has problem in executing the lsscsi command, as the exclusion logic depends on it.


VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -132,7 +132,9 @@ def disk_list_for_billing(self):
                 self.logger.log("LUN Number {0}, disk {1}".format(lunNumber,device_name))   
             self.logger.log("Disks to be included {0}".format(self.disksToBeIncluded))
         else:
-            self.logger.log("There is some glitch in executing the command sudo lsscsi, So the lsscsi list is empty and the Billing will not be done ")
+            if(self.isAnyDiskExcluded == True):
+                self.size_calc_failed = True
+                self.logger.log("There is some glitch in executing the command sudo lsscsi, So the lsscsi list is empty and the Billing will not be done ")
```
iamashwani-who, line 137: Message should be "There is some glitch in executing the command 'sudo lsscsi' and therefore size calculation is marked as failed."

"Billing will not be done" is incorrect.

iamashwani-who, line 137: This message should be logged in else block of if(len(..) != 0)) instead of under if (self.isAnyDiskExcluded == True)?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -60,32 +111,153 @@ def get_loop_devices(self):
...
+                self.root_devices.append(file_system[0])
+        self.logger.log("root_devices {0}".format(str(self.root_devices)))
+        self.logger.log("lsscsi_list {0}".format(self.lsscsi_list))
+        '''
+            Sample output of the lsscsi command 
+            [1:0:0:15]   disk    Msft     Virtual Disk     1.0   /dev/sda
+            [1:0:0:18]   disk    Msft     Virtual Disk     1.0   /dev/sdc
+        '''
+
+        self.disksToBeIncluded = self.disk_list_for_billing()
```
iamashwani-who, line 198: this pattern seems inconsistent. In disk_list_for_billing, we utilize the returned value even though the returned value itself is self.disksToBeIncluded and we do not return any value in get_logicalVolumes_for_billing method.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -24,9 +24,10 @@
 class SizeCalculation(object):
```
iamashwani-who, line 24: Resolve the merge conflict in this file.

VMBackup/main/Utils/SizeCalculation.py
arisettisanjana: Actually this will only occur when the disks have been excluded and lsscsi got failed, so removed this check here and added at the time we call this function.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -132,8 +132,8 @@ def disk_list_for_billing(self):
                 self.logger.log("LUN Number {0}, disk {1}".format(lunNumber,device_name))   
             self.logger.log("Disks to be included {0}".format(self.disksToBeIncluded))
         else:
-            self.logger.log("There is some glitch in executing the command sudo lsscsi, So the lsscsi list is empty and the Billing will not be done ")
-        return self.disksToBeIncluded
+            self.size_calc_failed = True
+            self.logger.log("There is some glitch in executing the command 'sudo lsscsi' and therefore size calculation is marked as failed. ")
```
iamashwani-who, line 136: Remove extra space at the end of the sentence.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -403,7 +402,15 @@ def get_total_used_size(self):
                         excluded_disks_used = excluded_disks_used + int(used)
                     #Including only the disks which are asked to include (Here LunList can't be empty this case is handled at the CRP end)
                     else:
-                        if mountpoint in self.device_mount_points and device != resource_disk_device:
+                        if self.isAnyDiskExcluded == False and device != resource_disk_device:
+                            #No disk has been excluded So can include every disk
```
ankitha-p, line 406: If no disk is excluded, include every non resource disk device

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -396,13 +396,13 @@ def get_total_used_size(self):
                             total_used = total_used + int(used)
                             self.logger.log("Adding only root device to size calculation. Device name : {0} used space in KB : {1} mount point : {2} fstype : {3}".format(device,used,mountpoint,fstype),True)
                             self.logger.log("Total Used Space: {0}".format(total_used),True)
-                    #Handling a case where LunList is empty for UnmanagedVM's
-                    elif(self.LunListEmpty == True and device != resource_disk_device):
+                    #Handling a case where LunList is empty for UnmanagedVM's and failures if occurred( as we will billing for all the non resource disks)
+                    elif( (self.size_calc_failed == True or self.LunListEmpty == True) and device != resource_disk_device):
```
iamashwani-who, line 400: We can convert this into a helper function that returns if AllDisksToBeIncluded.

shridpant (approved):

iamashwani-who (approved):

ankitha-p (approved):

PR #1658, opened 2023-01-10
Modify AMA TST

peizhen-w:
- Add new modules (perf counter, CPU/memory) based on the design doc:
  1. Incorporate perf counter checks to previous modules 
  2. Check logrotate 
  3. Monitor CPU/memory usage in 5 minutes
  4. Print top 10 most used mem/cpu process to log collector

- Modify previous modules based on CSS feedback:
  1. Run 'all' mode in one line
  2. Print workspace info in log collector
  3. Bug fixes


AzureMonitorAgent/ama_tst/ama_troubleshooter.sh
```diff
@@ -37,5 +37,10 @@ fi
 
 echo "Starting AMA Troubleshooting Tool v.$TST_VERSION..."
```
hestolz, line 38: need to update version?

AzureMonitorAgent/ama_tst/modules/high_cpu_mem/high_cpu_mem.py
```diff
@@ -0,0 +1,25 @@
...
+
+    # check log rotation
+    print("Checking if log rotation is working correctly...")
+    checked_logrot = check_log_rotation()
+    if (is_error(checked_logrot)):
+        return print_errors(checked_logrot)
+    else:
+        success = print_errors(checked_logrot)
+
+    # check AMA CPU/memory usage
```
aswatt, line 19: Nitpick: I'd recommend printing a line like above saying "Checking AMA's CPU and memory usage..." or similar, so the user knows the check is occurring

AzureMonitorAgent/ama_tst/modules/install/check_pkgs.py
```diff
@@ -16,7 +16,13 @@ def check_packages():
         return ERR_MULTIPLE_AMA
     
     # find subcomponent binaries
-    subcomponents = ['mdsd', 'telegraf', 'agentlauncher', 'amacoreagent', 'MetricsExtension', 'fluent-bit']
+    subcomponents = ['mdsd', 'agentlauncher', 'amacoreagent', 'fluent-bit']
+    metrics_file = "/etc/opt/microsoft/azuremonitoragent/config-cache/metricCounters.json"
```
aswatt, line 20: Same thing as above regarding static strings

AzureMonitorAgent/ama_tst/modules/logcollector.py
```diff
@@ -11,6 +11,10 @@
 RPM_CMD = "rpm -qi azuremonitoragent"
 PS_CMD = "ps -ef | grep {0} | grep -v grep"
 OPENSSL_CMD = "echo | openssl s_client -connect {0}:443 -brief"
+PS_CMD_CPU = "ps aux --sort=-pcpu | head -10"
```
aswatt, line 14: Is the `=` supposed to be here?

If so, also, you can make these a bit cleaner by just having one line `PS_CMD = "ps aux --sort {0} | head -10"`, and then whenever you use the string you can use `PS_CMD.format("-rss")`, which will be identical to `PS_CMD_RSS` for example

aswatt (changes requested):
It looks great! Small nitpicks here and there, but nothing major, mostly style-related.

AzureMonitorAgent/ama_tst/modules/main.py
```diff
@@ -39,6 +116,13 @@ def run_troubleshooter():
     if (not check_sudo()):
         return
     
+    print(sys.argv)
+    if len(sys.argv) > 1 and sys.argv[1] == '-A':
```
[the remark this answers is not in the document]
aswatt: That's a good point - we should update our docs as well

AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -203,9 +204,66 @@ def run_cmd_output(cmd):
...
+
+def is_metrics_configured():
+    global general_info
+    if 'metrics' in general_info:
+        return general_info['metrics']
+    
+    metrics_file = "/etc/opt/microsoft/azuremonitoragent/config-cache/metricCounters.json"
+    with open(metrics_file) as f:
+        output = f.read(2)
+        if output != '[]':
```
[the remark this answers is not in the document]
peizhen-w: If no metrics is collecting, the content inside this file will be` '[]'`. I checked `if os.path.isfile(<filepath>)` somewhere else with a ERR message. Because the return type seems inconsistent, so I didn't put file existence check in helpers.py

AzureMonitorAgent/ama_tst/modules/main.py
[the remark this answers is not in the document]
peizhen-w: Added '--help' option and '-L' option to run log collector directly

AzureMonitorAgent/ama_tst/modules/main.py
```diff
@@ -116,12 +116,17 @@ def run_troubleshooter():
...
-    print(sys.argv)
+    # run all checks from command line
     if len(sys.argv) > 1 and sys.argv[1] == '-A':
         success = check_all(False)
         print_results(success)
         print_next_steps()
         return
+    
+    # run log collector from command line
+    if len(sys.argv) > 1 and sys.argv[1] == '-L':
```
hestolz, line 127: JFYI, can be in a future PR but I'd recommend moving to use [argparse](https://docs.python.org/3/library/argparse.html). It makes parsing cmd line arguments much cleaner.

AzureMonitorAgent/ama_tst/modules/helpers.py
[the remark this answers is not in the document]
aswatt: Sounds good! Thank you for clarifying!

AzureMonitorAgent/ama_tst/modules/logcollector.py
aswatt: Ah makes sense, have you validated that these commands work on their own? The OMS log collector isn't 100% perfect, so there might be typos - if the command works better as written, then yeah the way you organized it is best

aswatt (approved):
Lgtm! I'd still take a look over any unresolved comments, but approving so you're not blocked to merge.

aswatt (changes requested):
Oh wait, I just remembered - please rebuild the troubleshooter bundle with these changes!

AzureMonitorAgent/ama_tst/modules/logcollector.py
peizhen-w: Yes I tested the commands work for AMA

AzureMonitorAgent/ama_tst/modules/main.py
peizhen-w: I will include this change in the next PR, thanks for the advice!

aswatt (approved):

PR #1659, opened 2023-01-11
Don't change permissions on defaults file as it breaks rpm verification

simathih:
Refer to ICM 360024501


hestolz (approved):
LGTM. also checked myself, no impact on ability of mdsd to start up w/this perms change. Thanks Sidd 👌

PR #1661, opened 2023-01-17
Pankajjoshi/temp disk encryption m1

pankajosh:
draft commit - testing is in progress, 
feature: temp disk encryption for CVM,
             Encryption applied only of OS is encrypted in CVM.
             volume type = DATA/ALL is valid input. This commit is encrypting only resource disk. 
             encrypt-format-all is not required to encrypt resource disk.
             keyfile is placed at /var/lib/azure_disk_encryption_config, 
             crypttab is updated. 
             retaining mountpoint for resource disk in case of CVM. 
             HotFix suggested for IMDS -- This commit is removed and created a separate PR. 
                   1. IMDS recommended noproxy for IMDS call.
                   2. considering IMDS failure as non terminal.


VMEncryption/main/Common.py
```diff
@@ -94,6 +94,7 @@ class CommonVariables:
     default_mount_name = 'encrypted_disk'
     dev_mapper_root = '/dev/mapper/'
     osmapper_name = 'osencrypt'
+    cvmosmapper_name = 'cloudimg-rootfs'
```
pankajosh, line 97: remove cvmosmapper_name, as this is not used in code. 
using lablk to find device information.

VMEncryption/main/AbstractBekUtilImpl.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def __init__(self) -> None:
+        pass
+
+    def generate_passphrase(self):
+        if TestHooks.use_hard_code_passphrase:
+            return TestHooks.hard_code_passphrase
+        else:
+            with open("/dev/urandom", "rb") as _random_source:
+                bytes = _random_source.read(127)
```
ShridharParameshwarBhat, line 51: Can we keep the value `127` in a constant instead of hardcoding the value?

VMEncryption/main/AbstractBekUtilImpl.py
```diff
@@ -0,0 +1,95 @@
...
+                    f.write(passphrase)
+
+    @abstractmethod
+    def store_bek_passphrase(self, encryption_config, passphrase):
+        pass
+    @abstractmethod
+    def get_bek_passphrase_file(self, encryption_config):
+        pass
+    
+    def mount_bek_volume(self):
```
ShridharParameshwarBhat, line 75: Isn't munt_bek_volume an abstract method too? Why does this not have a decorator as `abstractmethod` ?

VMEncryption/main/AbstractBekUtilImpl.py
```diff
@@ -0,0 +1,95 @@
...
+        pass
+
+    def is_bek_volume_mounted_and_formatted(self):
```
ShridharParameshwarBhat, line 78: Similarly why is this not decorated as `abstractmethod`?

VMEncryption/main/BekUtil.py
```diff
@@ -16,151 +16,50 @@
...
+            security_type = imds_Stored_Results.get_security_type()
+        except Exception as ex:
+            logger.log("Failure in retriving security type from IMDS, Following BEK Volume Path.")
+        finally:
+            logger.log("BEK util path is followed according to VM security type {0}".format(security_type))
+                
+        if security_type != CommonVariables.ConfidentialVM:
+           self.bekUtilImpl = BekUtilVolumeImpl(disk_util,logger)
+        else:
+            self.bekUtilImpl = BekUtilFileImpl(disk_util,logger)
```
ShridharParameshwarBhat, line 41: I don't understand as to why this was refactored into two implementations. Are these implementations different depending on the security type? If this is the case I think we should have a base class as BekUtil and a subclass called CVMBekUtil that overrides any behavior for CVM feature.

VMEncryption/main/DiskUtil.py
```diff
@@ -1172,3 +1172,14 @@ def get_luks_header_size(self, device_path=None):
                 header_size = None
 
             return header_size
+            
+    def get_osmapper_name(self):
+        osmapper_name = CommonVariables.osmapper_name
+        try:
+            rootfs_device = list(filter(lambda d:d.mount_point=='/',self.get_device_items(None)))
+            if len(rootfs_device)!=0 and rootfs_device[0].type == 'crypt':
```
ShridharParameshwarBhat, line 1180: pythonic way is 

`if not len(rootfs_device)` instead of `if len(rootfs_device)!=0 `

ShridharParameshwarBhat: > pythonic way is
> 
> `if rootfs_device` instead of `if len(rootfs_device)!=0 `

VMEncryption/main/Common.py
pankajosh: removed

VMEncryption/main/BekUtil.py
pankajosh: Resolving based on our conversation during team meetings.

canfikret (changes requested):

VMEncryption/main/BekUtil.py
canfikret: For the record, this is for runtime polymorphism. Different implementation is chosen for CVM vs standard VM.

canfikret (approved):
Pls follow with a PR for AzureAutomation update. Thanks.

VMEncryption/main/BekUtilFileImpl.py
```diff
@@ -0,0 +1,69 @@
...
+    
+    def store_bek_passphrase(self, encryption_config, passphrase):
+        bek_filename = encryption_config.get_bek_filename()
+        try:
+            self.disk_util.make_sure_path_exists(self.keyfilePath)
+            self.store_passphrase(key_File_Path=self.keyfilePath,
+                                  bek_filename=bek_filename,
+                                  passphrase=passphrase)            
+        except Exception as e:
+            message = "Failed to store BEK in BEK VOLUME with error: {0}".format(traceback.format_exc())
```
vimish, line 42: This message needs to be updated. I believe the passphrase is not stored in BEK volume here.

VMEncryption/main/BekUtilFileImpl.py
```diff
@@ -0,0 +1,69 @@
...
+            self.logger.log(message)
+            raise e
+        else:
+            self.logger.log("Stored BEK in the BEK VOLUME successfully")
```
vimish, line 46: Same as above.

VMEncryption/main/AbstractBekUtilImpl.py
```diff
@@ -0,0 +1,96 @@
...
+       self.value = value
+
+    def __str__(self):
+       return(repr(self.value))
+
+class AbstractBekUtilImpl(ABC):
+    '''
+    This is an interface used for funcitonality implementation for BEK util class
+    '''
+    wrong_fs_msg = "BEK disk does not have vfat filesystem."
```
vimish, line 42: These messages are specific to BEK volume. I believe these should be moved to volume impl class

VMEncryption/main/BekUtilFileImpl.py
```diff
@@ -0,0 +1,69 @@
...
+import traceback
+from AbstractBekUtilImpl import AbstractBekUtilImpl
+import sys
+import os
+
+class BekUtilFileImpl(AbstractBekUtilImpl):
+    """
+    Utility class to store passphrase in a keyfilepath.
+    """
+    def __init__(self,diskutil,logger):
```
vimish, line 28: nit: disk_util

VMEncryption/main/BekUtilFileImpl.py
```diff
@@ -0,0 +1,69 @@
...
+            if os.path.exists(os.path.join(self.keyfilePath, bek_filename)):
+                return os.path.join(self.keyfilePath, bek_filename)
+
+            for filename in os.listdir(self.keyfilePath):
+                if bek_filename in filename:
+                    return os.path.join(self.keyfilePath, filename)
+
+        except Exception as e:
+            # use traceback to convert exception to string on both python2 and python3+
+            message = "Failed to get BEK from BEK VOLUME with error: {0}".format(traceback.format_exc())
```
vimish, line 66: Same as above.

VMEncryption/main/DiskUtil.py
```diff
@@ -1172,3 +1172,14 @@ def get_luks_header_size(self, device_path=None):
                 header_size = None
 
             return header_size
+            
+    def get_osmapper_name(self):
+        osmapper_name = CommonVariables.osmapper_name
+        try:
```
vimish, line 1178: Should this happen only in case of CVM? A corner case I can think of is if a customer have manually encrypted the OS volume

VMEncryption/main/handle.py
```diff
@@ -971,7 +1005,21 @@ def enable_encryption():
...
+                encryptResourceDiskforCVM = False
+                if security_Type==CommonVariables.ConfidentialVM:
+                    if confidential_encryption_tempdisk:
+                        encryptResourceDiskforCVM = True
+                    else: 
+                        message = "Resource disk not encrypted. It is confidential VMs, but PrivatePreview.ConfidentialEncryptionTempDisk is not set"
+                        logger.log(msg=message)
+                                
+                if encryptResourceDiskforCVM or \
+                    extension_parameter.command == CommonVariables.EnableEncryptionFormatAll:
```
vimish, line 1022: What happens if user specifies EFA for CVM by mistake and did not set the preview flag? Will the temp disk be encrypted?

VMEncryption/main/oscrypto/OSEncryptionState.py
```diff
@@ -54,7 +54,8 @@ def __init__(self, state_name, context):
                                                             disk_util=self.disk_util)
 
         self.bek_util = BekUtil(disk_util=self.disk_util,
-                                logger=self.context.logger)
+                                logger=self.context.logger,
+                                encryption_environment=self.context.encryption_environment)
```
vimish, line 58: nit: Probably encryption_env parameter could have been made optional to avoid any change in OS encryption path. No need to make the change now but something to consider for future changes.

VMEncryption/main/test/test_resource_disk_util.py
```diff
@@ -39,7 +39,7 @@ def setUp(self):
         self.mock_crypt_mount_config_util = mock.create_autospec(CryptMountConfigUtil)
         self.mock_passphrase_filename = "mock_passphrase_filename"
         mock_public_settings = {}
-        self.resource_disk = ResourceDiskUtil(self.logger, self.mock_disk_util, self.mock_crypt_mount_config_util, self.mock_passphrase_filename, mock_public_settings, ["ubuntu", "16"])
+        self.resource_disk = ResourceDiskUtil(self.logger, self.mock_disk_util, self.mock_crypt_mount_config_util, self.mock_passphrase_filename, mock_public_settings, ["ubuntu", "16"],False)
```
vimish, line 42: Will it be possible to add unit test for CVM scenarios to? Not a blocker for this PR but if possible then add in next PR.

VMEncryption/main/handle.py
```diff
@@ -1011,15 +1011,18 @@ def enable_encryption():
                 public_settings = get_public_settings()
                 confidential_encryption_tempdisk = public_settings.get("PrivatePreview.ConfidentialEncryptionTempDisk")
                 encryptResourceDiskforCVM = False
+                encryptResourceDisk = False
```
vimish, line 1014: Two flags seems redundant here. Can we just use encryptResourceDisk flag and set it to true either for CVM or EFA?

vimish (approved):

VMEncryption/main/AbstractBekUtilImpl.py
bamuir: If you could relay responses to this quesion on here @pankajosh that would be great, as I am also curious on this now and see it is marked resolved

VMEncryption/main/DiskUtil.py
```diff
@@ -1172,3 +1172,14 @@ def get_luks_header_size(self, device_path=None):
                 header_size = None
 
             return header_size
+            
+    def get_osmapper_name(self):
```
bamuir, line 1176: Can you add some comments on what the osmapper is?

PR #1665, opened 2023-01-25
Bamuir/ade mhsm support

bamuir:
Adding support for ManagedHSM on ADE for Linux. This introduces the concept of a KeyStoreType, which when set to ManagedHSM, causes the extension to send the new setting version (5.0) to host.

### DIFF
--- a/VMEncryption/main/Common.py
+++ b/VMEncryption/main/Common.py
@@ -114,7 +114,8 @@ class CommonVariables:
     MigrateValue = 'Migrate'
 
     """
-    value for KeyStoreTypeKey could be MHSM
+    value for KeyStoreTypeKey could be ManagedHSM
+    This is a value for ManagedHSM testing, and should be removed before public release
     """
     KeyStoreTypeKey = 'KeyStoreType'
     KeyStoreTypeManagedHSM = 'ManagedHSM'
--- a/VMEncryption/main/Common.py
+++ b/VMEncryption/main/Common.py
@@ -114,6 +114,12 @@ class CommonVariables:
     MigrateValue = 'Migrate'
 
     """
+    value for KeyStoreTypeKey could be MHSM
+    """
+    KeyStoreTypeKey = 'KeyStoreType'
+    KeyStoreTypeManagedHSM = 'ManagedHSM'
+
+    """
     value for VolumeType could be Data
     """
     VolumeTypeKey = 'VolumeType'
--- a/VMEncryption/main/EncryptionSettingsUtil.py
+++ b/VMEncryption/main/EncryptionSettingsUtil.py
@@ -186,24 +186,34 @@ class EncryptionSettingsUtil(object):
                     })
         return array
 
-    def get_settings_data(self, protector_name, kv_url, kv_id, kek_url, kek_kv_id, kek_algorithm, extra_device_items, disk_util, crypt_mount_config_util, key_store_type):
+    def get_settings_data(self, protector_name, kv_url, kv_id, kek_url, kek_kv_id, kek_algorithm, extra_device_items, disk_util, crypt_mount_config_util, key_store_type, keystoretype_flag_exists):
         """ returns encryption settings object in format required by wire server """
 
         cutil = CheckUtil(self.logger)
+
+        # Perform algorithm check regaurdless of ManagedHSM or Keyvault
+        if kek_algorithm not in CommonVariables.encryption_algorithms:
+                    if kek_algorithm:
+                        raise Exception("The KEK encryption algorithm requested was not recognized")
+                    else:
+                        kek_algorithm = CommonVariables.default_encryption_algorithm
+                        self.logger.log("No KEK algorithm specified, defaulting to {0}".format(kek_algorithm))
 
         # ManagedHSM Checks
         if key_store_type.lower() == CommonVariables.KeyStoreTypeManagedHSM.lower():
             # validate mhsm parameters prior to creating the encryption settings object
             if not kv_url.strip() or not kv_id.strip():
-                raise Exception("KeyVaultUrl or KeyVaultResourceId are not empty. Please remove them for use with 'KeyStoreType':'ManagedHSM'.")
+                raise Exception("KeyvaultUrl or KeyvaultresourceId are not empty. 'KeyStoreType' parameter is not supported for keyvault currently, please remove Key Vault p")
             self.logger.log("get_settings_data: KeyVault Url is empty, validating KeyEncryptionKeyKVURL and KeyEncryptionKeyKVId for ManagedHSM")
-            cutil.check_mhsm_url(kek_url, "A ManagedHSM URL was specified, but it was invalid for ManagedHSM. ")
-            cutil.check_mhsm_id(kek_kv_id, "A ManagedHSM ID is required, but is missing or invalid. ")
+            cutil.check_mhsm_url(kek_url, "A ManagedHSM URL is specified, but it is invalid for ManagedHSM.")
+            cutil.check_mhsm_id(kek_kv_id, "A ManagedHSM ID is required, but is missing or invalid.")
             cutil.check_mhsm_name(kek_kv_id, kek_url, "A ManagedHSM ID and ManagedHSM URL were provided, but their ManagedHSM names did not match")
             self._DISK_ENCRYPTION_KEY_SOURCE = CommonVariables.KeyStoreTypeManagedHSM
         elif key_store_type:
             raise Exception("The expected flag name and value to enable ManagedHSM is 'KeyStoreType':'{0}'. " +
             "Please correct the flag name and value and retry enabling ManagedHSM".format(CommonVariables.KeyStoreTypeManagedHSM))
+        elif not key_store_type and keystoretype_flag_exists:
+            raise Exception("The flag to enable ManagedHSM is empty. Remove the flag for use with KeyVault, or correct the value to 'KeyStoreType':'{0}'".format(CommonVariables.KeyStoreTypeManagedHSM))
         # Keyvault Checks
         else:
             # validate key vault parameters prior to creating the encryption settings object
@@ -214,12 +224,6 @@ class EncryptionSettingsUtil(object):
                 cutil.check_kv_id(kek_kv_id, "A KEK URL was specified, but its KEK KeyVault ID was missing or invalid")
                 cutil.check_kek_url(kek_url, "A KEK URL was specified, but it was invalid")
                 cutil.check_kek_name(kek_kv_id, kek_url, "A KEK ID and KEK URL were provided, but their key vault names did not match")
-                if kek_algorithm not in CommonVariables.encryption_algorithms:
-                    if kek_algorithm:
-                        raise Exception("The KEK encryption algorithm requested was not recognized")
-                    else:
-                        kek_algorithm = CommonVariables.default_encryption_algorithm
-                        self.logger.log("No KEK algorithm specified, defaulting to {0}".format(kek_algorithm))
             else:
                 if kek_kv_id:
                     raise Exception("The KEK KeyVault ID was specified but the KEK URL was missing")
--- a/VMEncryption/main/EncryptionSettingsUtil.py
+++ b/VMEncryption/main/EncryptionSettingsUtil.py
@@ -202,8 +202,8 @@ class EncryptionSettingsUtil(object):
         # ManagedHSM Checks
         if key_store_type and key_store_type.lower() == CommonVariables.KeyStoreTypeManagedHSM.lower():
             # validate mhsm parameters prior to creating the encryption settings object
-            if not kv_url.strip() or not kv_id.strip():
-                raise Exception("KeyvaultUrl or KeyvaultresourceId are not empty. 'KeyStoreType' parameter is not supported for keyvault currently, please remove Key Vault p")
+            if kv_url or kv_id:
+                raise Exception("KeyvaultUrl or KeyvaultresourceId are not empty, and 'KeyStoreType' parameter is set to ManagedHSM. Please remove KeyvaultUrl KeyvaultresourceId for ManagedHSM.")
             self.logger.log("get_settings_data: KeyVault Url is empty, validating KeyEncryptionKeyKVURL and KeyEncryptionKeyKVId for ManagedHSM")
             cutil.check_mhsm_url(kek_url, "A ManagedHSM URL is specified, but it is invalid for ManagedHSM.")
             cutil.check_mhsm_id(kek_kv_id, "A ManagedHSM ID is required, but is missing or invalid.")
--- a/VMEncryption/main/EncryptionSettingsUtil.py
+++ b/VMEncryption/main/EncryptionSettingsUtil.py
@@ -200,7 +200,7 @@ class EncryptionSettingsUtil(object):
                         self.logger.log("No KEK algorithm specified, defaulting to {0}".format(kek_algorithm))
 
         # ManagedHSM Checks
-        if key_store_type.lower() == CommonVariables.KeyStoreTypeManagedHSM.lower():
+        if key_store_type and key_store_type.lower() == CommonVariables.KeyStoreTypeManagedHSM.lower():
             # validate mhsm parameters prior to creating the encryption settings object
             if not kv_url.strip() or not kv_id.strip():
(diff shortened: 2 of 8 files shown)


bamuir:
- [ ] Add Unit Tests if possible
- [ ] Test e2e with Shridhar's Host Agent changes

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -182,30 +186,47 @@ def _dict_to_name_value_array(self, values):
...
-                raise Exception("The KEK KeyVault ID was specified but the KEK URL was missing")
+            # validate key vault parameters prior to creating the encryption settings object
+            cutil.check_kv_id(kv_id, "A KeyVault ID is required, but is missing or invalid")
+            cutil.check_kv_url(kv_url, "A KeyVault URL is required, but is missing or invalid")
+            cutil.check_kv_name(kv_id, kv_url, "A KeyVault ID and KeyVault URL were provided, but their key vault names did not match")
+            if kek_url:
+                cutil.check_kv_id(kek_kv_id, "A KEK URL was specified, but its KEK KeyVault ID was missing or invalid")
+                cutil.check_kek_url(kek_url, "A KEK URL was specified, but it was invalid")
+                cutil.check_kek_name(kek_kv_id, kek_url, "A KEK ID and KEK URL were provided, but their key vault names did not match")
+                if kek_algorithm not in CommonVariables.encryption_algorithms:
```
bamuir, line 217: Vishal: Perform algorithm check on MHSM as well

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -182,30 +186,47 @@ def _dict_to_name_value_array(self, values):
                     })
         return array
 
-    def get_settings_data(self, protector_name, kv_url, kv_id, kek_url, kek_kv_id, kek_algorithm, extra_device_items, disk_util, crypt_mount_config_util):
+    def get_settings_data(self, protector_name, kv_url, kv_id, kek_url, kek_kv_id, kek_algorithm, extra_device_items, disk_util, crypt_mount_config_util, key_store_type):
```
bamuir, line 189: Double check all calls to this for the added parameter

VMEncryption/main/Common.py
```diff
@@ -113,6 +113,12 @@ class CommonVariables:
     MigrateKey = 'MigrateFlag'
     MigrateValue = 'Migrate'
 
+    """
+    value for KeyStoreTypeKey could be MHSM
```
ShridharParameshwarBhat, line 117: value for KeyStoreTypeKey could be `ManagedHSM`  (instead of `MHSM)` just to convey the right string value in the comment.

ShridharParameshwarBhat: Probably you can also add an comment saying that , the parameter would accept `Keyvault` as the value in future. But this is optional.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -182,30 +186,47 @@ def _dict_to_name_value_array(self, values):
...
-                    raise Exception("The KEK encryption algorithm requested was not recognized")
-                else:
-                    kek_algorithm = CommonVariables.default_encryption_algorithm
-                    self.logger.log("No KEK algorithm specified, defaulting to {0}".format(kek_algorithm))
+
+        # ManagedHSM Checks
+        if key_store_type.lower() == CommonVariables.KeyStoreTypeManagedHSM.lower():
+            # validate mhsm parameters prior to creating the encryption settings object
+            if not kv_url.strip() or not kv_id.strip():
+                raise Exception("KeyVaultUrl or KeyVaultResourceId are not empty. Please remove them for use with 'KeyStoreType':'ManagedHSM'.")
```
ShridharParameshwarBhat, line 198: "KeyStoreType":"ManagedHSM" only holds true for ARM template as customers have to provide key value parameters in json file. This codepath can be hit for powershell scenario too.

So it would be good to just mention in the log that `KeyvaultUrl or KeyvaultresourceId are not empty. 'KeyStoreType' parameter is not supported for keyvault currently.`

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -182,30 +186,47 @@ def _dict_to_name_value_array(self, values):
...
+            self.logger.log("get_settings_data: KeyVault Url is empty, validating KeyEncryptionKeyKVURL and KeyEncryptionKeyKVId for ManagedHSM")
+            cutil.check_mhsm_url(kek_url, "A ManagedHSM URL was specified, but it was invalid for ManagedHSM. ")
+            cutil.check_mhsm_id(kek_kv_id, "A ManagedHSM ID is required, but is missing or invalid. ")
```
ShridharParameshwarBhat, line 201: We can make usage of `was` or `is` consistent . Either both of them as `is` or `was`. I prefer `is`. And trim the whitespace on the end of the lines too.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -182,30 +186,47 @@ def _dict_to_name_value_array(self, values):
...
+            cutil.check_mhsm_name(kek_kv_id, kek_url, "A ManagedHSM ID and ManagedHSM URL were provided, but their ManagedHSM names did not match")
+            self._DISK_ENCRYPTION_KEY_SOURCE = CommonVariables.KeyStoreTypeManagedHSM
+        elif key_store_type:
+            raise Exception("The expected flag name and value to enable ManagedHSM is 'KeyStoreType':'{0}'. " +
```
ShridharParameshwarBhat, line 205: Lets say the customer passes "`KeyStoreType":""` , in the template , So for this case the code doesn't raise the exception because it is empty and the condition `elif key_store_type:` returns `false`.

You should raise the exception when the key `KeyStoreType` exists in the `elif` case.

VMEncryption/main/check_util.py
```diff
@@ -99,6 +99,14 @@ def check_kek_url(self, test_kek_url, message):
             raise Exception('\n' + message + '\nActual: ' + test_kek_url + '\nExpected: ' + expected + "\n")
         return
 
+    def check_mhsm_url(self, test_mhsm_url, message):
+        """basic sanity check of the MHSM url"""
+        expected = "https://keyvault-name}.{vault-endpoint}"
```
ShridharParameshwarBhat, line 104: The expected string here needs update. It should refer to managedHSM and not keyvault

VMEncryption/main/check_util.py
```diff
@@ -99,6 +99,14 @@ def check_kek_url(self, test_kek_url, message):
...
+        pattern = re.compile(r'https://(.)+\\.(managedhsm)\\.(.)+(:443)?\\/keys/[^\\/]+\\/[0-9a-zA-Z]+$', re.IGNORECASE)
+        if not (test_mhsm_url and pattern.match(test_mhsm_url)):
+            raise Exception('\n' + message + '\nActual: ' + test_mhsm_url + '\nExpected: ' + expected + "\n")
```
ShridharParameshwarBhat, line 107: Why `test_mhsm_url and pattern.match(test_mhsm_url)` , can't we just have `pattern.match(test_mhsm_url)`?

VMEncryption/main/check_util.py
```diff
@@ -107,6 +115,23 @@ def check_kv_id(self, test_kv_id, message):
             raise Exception('\n' + message + '\nActual: ' + test_kv_id + '\nExpected: ' + expected + "\n")
         return
 
+    def check_mhsm_id(self, test_mhsm_id, message):
```
ShridharParameshwarBhat, line 118: You can try to replicate the `check_kv_id` function to check_mhsm_id` function with few updates. Both essentially do the same thing with a little difference. Good to have similar implementation.

VMEncryption/main/EncryptionSettingsUtil.py
bamuir: There do not appear to be any additional tests using this function, double checked

VMEncryption/main/EncryptionSettingsUtil.py
bamuir: Not sure I am following. If the customer passes "KeyStoreType":"" , correct, the exception would not be hit. It will only be hit if key_store_type exists and is not "ManagedHSM", so for example, "  " or "KeyVault" would throw. 

If the customer passes "KeyStoreType":"" with a ManagedHSM URL, I expect the KeyVault checks would fail. If we want it to fail automatically on this, I would need an "elif not key_store_type:" that throws an exception saying something like "Key Store Flag passed with null/empty value is not allowed, remove the flag or pass the value of 'ManagedHSM'"

VMEncryption/main/check_util.py
bamuir: I followed other check conventions of kek and kv_url checks

VMEncryption/main/check_util.py
bamuir: I think I missed a commit here, this looks unfinished. I will revisit these checks and implement tests, I must have forgotten a save

VMEncryption/main/Common.py
bamuir: As per discussion with Jack, I will add a comment that this flag is for testing purposes only and will be removed before public MHSM release

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -182,30 +186,51 @@ def _dict_to_name_value_array(self, values):
...
+            cutil.check_mhsm_url(kek_url, "A ManagedHSM URL is specified, but it is invalid for ManagedHSM.")
+            cutil.check_mhsm_id(kek_kv_id, "A ManagedHSM ID is required, but is missing or invalid.")
+            cutil.check_mhsm_name(kek_kv_id, kek_url, "A ManagedHSM ID and ManagedHSM URL were provided, but their ManagedHSM names did not match")
+            self._DISK_ENCRYPTION_KEY_SOURCE = CommonVariables.KeyStoreTypeManagedHSM
+        elif key_store_type:
+            raise Exception("The expected flag name and value to enable ManagedHSM is 'KeyStoreType':'{0}'. " +
+            "Please correct the flag name and value and retry enabling ManagedHSM".format(CommonVariables.KeyStoreTypeManagedHSM))
+        elif not key_store_type and keystoretype_flag_exists:
+            raise Exception("The flag to enable ManagedHSM is empty. Remove the flag for use with KeyVault, or correct the value to 'KeyStoreType':'{0}'".format(CommonVariables.KeyStoreTypeManagedHSM))
+        # Keyvault Checks
```
ShridharParameshwarBhat, line 217: I think you can minimize the two elifs to one elif which makes the logic slightly complicated.

You have these two elifs

```
 elif key_store_type:
            raise Exception("The expected flag name and value to enable ManagedHSM is 'KeyStoreType':'{0}'. " +
            "Please correct the flag name and value and retry enabling ManagedHSM".format(CommonVariables.KeyStoreTypeManagedHSM))
        elif not key_store_type and keystoretype_flag_exists:
            raise Exception("The flag to enable ManagedHSM is empty. Remove the flag for use with KeyVault, or correct the value to 'KeyStoreType':'{0}'".format(CommonVariables.KeyStoreTypeManagedHSM))```
```			

You can just have one elif.

```
elif keystoretype_flag_exists:
	raise Exception(<Message: The expected flag name and value to enable ManagedHSM is  'KeyStoreType':'{0}' , Please correct the flag name " +
                        "and value and retry enabling ManagedHSM>)
```

VMEncryption/main/test/test_check_util.py
```diff
@@ -57,6 +57,22 @@ def test_is_kv_id(self):
...
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroups/{rgname}/providers/Microsoft.KeyVault////////", "")
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroupssss/{rgname}/providers/Microsoft.KeyVault/vaults/{vaultname}", "")
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroups/{rgname}/providers/Microsoft.KeyVault/vaults/{vaultname}", "")
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroupssss/{rgname}/providers/Microsoft.KeyVault/managedHSM/{vaultname}", "")
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroups/{rgname}/providers/Microsoft.KeyVault/managedHSM/{vaultname}", "")
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroups/{rgname}/providers/Microsoft.ManagedHSM/managedHSM/{vaultname}", "")
+        self.assertRaises(Exception, self.cutil.check_mhsm_id, "/subscriptions/{subid}/resourceGroups/{rgname}/providers/Microsoft.ManagedHSM/vaults/{vaultname}", "")
+
```
ShridharParameshwarBhat, line 75: Why can't we just have some test names for subid (guid of subscriptionId) , rgname (something like : testrg) and valutname (like : testkv) instead of generic names such as {subid} , {rgname} and {vaultname} ? The tests may become slightly confusing because we expect the proper names for these parameters and not the one surrounded with `{` and `}`.

VMEncryption/main/test/test_check_util.py
```diff
@@ -57,6 +57,22 @@ def test_is_kv_id(self):
...
     def test_is_kv_url(self):
```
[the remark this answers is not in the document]
bamuir: @ShridharParameshwarBhat can you provide me with an example of the MHSM URL / any other parameters that I can test outside the MHSM ID which I have?

VMEncryption/main/test/test_check_util.py
bamuir: I copied from the kv_id checks. My guess is that that is left over sometimes from templates/docs, so it is a test to see if bad parameters are caught? This test expects failure. There are also other tests without the {}

ShridharParameshwarBhat: Oh got it. Then nevermind . I did not see that the its a negative test.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -209,12 +209,10 @@ def get_settings_data(self, protector_name, kv_url, kv_id, kek_url, kek_kv_id, k
             cutil.check_mhsm_id(kek_kv_id, "A ManagedHSM ID is required, but is missing or invalid.")
             cutil.check_mhsm_name(kek_kv_id, kek_url, "A ManagedHSM ID and ManagedHSM URL were provided, but their ManagedHSM names did not match")
             self._DISK_ENCRYPTION_KEY_SOURCE = CommonVariables.KeyStoreTypeManagedHSM
-        elif key_store_type:
+        elif keystoretype_flag_exists:
             raise Exception("The expected flag name and value to enable ManagedHSM is 'KeyStoreType':'{0}'. " +
```
ShridharParameshwarBhat, line 213: Looks good now . But do test the scenarios. 














Looks good to me now. But do test for all the scenarios that you could think of. Below were some of the test cases which I documented for testing my change for new parameter.


Enable with keyvault

Enable with keyvault + kek

Enable with kek and no keyvault

Enable with "KeyStoreType" set to "" - with keyvault

Enable with "KeyStoreType" set to "" - with Keyvault + kek

Enable with "KeyStoreType" set to "random" - with keyvault

Enable with "keysToReTyPe" set to "" - with keyvault

Enable with "keysToReTyPe" set to "random" - with keyvault

Enable with "keysToReTyPe" set to "ManagedHSM" - with keyvault

Enable with "keysToReTyPe" set to "manageDHsM" - with keyvault

Enable with "KeyStoreType" set to "managedhsm" - with Keyvault

Enable with "KeyStoreType" set to "managedhsm" - with Kek URL instead of managed hsm

Enable with "KeyStoreType" set to "managedhsm" - with invalid mhsm 

Enable with "KeyStoreType" set to "managedhsm" - with valid mhsm

bamuir:
Latest PR: https://github.com/Azure/azure-linux-extensions/pull/1682

PR #1667, opened 2023-01-30
Add traceFlags settings to enable mdsd tracing in AMA via public settings

AaronRM:
Add traceFlags settings to enable mdsd tracing in AMA via public settings.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -417,13 +417,18 @@ def enable():
         ("azuremonitoragentmgr", False)
     ])
 
+    # Set traceFlags in publicSettings to enable mdsd tracing. For example, the EventIngest flag can be enabled via "traceFlags": "0x2"
+    flags = ""
+    if public_settings is not None and "traceFlags" in public_settings:
+        flags = "-T {} ".format(public_settings.get("traceFlags"))
+
     # Use an Ordered Dictionary to ensure MDSD_OPTIONS (and other dependent variables) are written after their dependencies
     default_configs = OrderedDict([
         ("MDSD_CONFIG_DIR", "/etc/opt/microsoft/azuremonitoragent"),
         ("MDSD_LOG_DIR", "/var/opt/microsoft/azuremonitoragent/log"),
         ("MDSD_ROLE_PREFIX", "/run/azuremonitoragent/default"),
         ("MDSD_SPOOL_DIRECTORY", "/var/opt/microsoft/azuremonitoragent"),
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -417,13 +417,18 @@ def enable():
         ("azuremonitoragentmgr", False)
     ])
 
+    # Set traceFlags in publicSettings to enable mdsd tracing. For example, the EventIngest flag can be enabled via "traceFlags": "0x2"
+    flags = ""
+    if public_settings is not None and "traceFlags" in public_settings:
+        flags = "-T {} ".format(public_settings.get("traceFlags"))
```
simathih, line 423: shouldn't it be {0}?

AzureMonitorAgent/agent.py
```diff
@@ -417,13 +417,18 @@ def enable():
...
+
     # Use an Ordered Dictionary to ensure MDSD_OPTIONS (and other dependent variables) are written after their dependencies
     default_configs = OrderedDict([
         ("MDSD_CONFIG_DIR", "/etc/opt/microsoft/azuremonitoragent"),
         ("MDSD_LOG_DIR", "/var/opt/microsoft/azuremonitoragent/log"),
         ("MDSD_ROLE_PREFIX", "/run/azuremonitoragent/default"),
         ("MDSD_SPOOL_DIRECTORY", "/var/opt/microsoft/azuremonitoragent"),
-        ("MDSD_OPTIONS", "\"-A -c /etc/opt/microsoft/azuremonitoragent/mdsd.xml -d -r $MDSD_ROLE_PREFIX -S $MDSD_SPOOL_DIRECTORY/eh -L $MDSD_SPOOL_DIRECTORY/events\""),
+        ("MDSD_OPTIONS", "\"{}-A -c /etc/opt/microsoft/azuremonitoragent/mdsd.xml -d -r $MDSD_ROLE_PREFIX -S $MDSD_SPOOL_DIRECTORY/eh -L $MDSD_SPOOL_DIRECTORY/events\"".format(flags)),
```
simathih, line 431: shouldn't it be {0}?

AzureMonitorAgent/agent.py
AaronRM: [The index is not necessary on Python 3.1+](https://docs.python.org/3/library/string.html#formatstrings). "The positional argument specifiers can be omitted for [str.format()](https://docs.python.org/3/library/stdtypes.html#str.format), so '{} {}'.format(a, b) is equivalent to '{0} {1}'.format(a, b)."

AzureMonitorAgent/agent.py
AaronRM: See previous comment.

simathih (approved):

NarineM (approved):

PR #1672, opened 2023-02-03
TST Log Collector Updates

aswatt:
- Collect rsyslog journalctl updates for last 30 days
- Add azEnvironment, resourceId, and location values taking from IMDS metadata to amalinux.out
- Add `systemctl status rsyslog \ syslog-ng` output to amalinux.out
- A bit of code cleanup


peizhen-w (approved):
lgtm

PR #1674, opened 2023-02-07
Add Amazon 2 Linux Support

aswatt:
Verified that AMA works on Amazon 2 Linux VMs installed via Arc, adding official support. In addition, currently if the OS isn't supported, the error message seen by the cx will be "The OS has neither rpm or dpkg" - added the VM support check in a few common unintuitive failure locations to make it clearer when the issue is due to an unsupported OS.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -374,6 +374,7 @@ def uninstall():
     Note: uninstall operation times out from WAAgent at 5 minutes
     """
 
+    exit_if_vm_not_supported('Uninstall')
     find_package_manager("Uninstall")
     if PackageManager == "dpkg":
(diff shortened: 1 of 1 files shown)


simathih (dismissed):

AzureMonitorAgent/agent.py
```diff
@@ -1190,6 +1190,7 @@ def find_package_manager(operation):
             break
 
     if PackageManager == "":
+        exit_if_vm_not_supported(operation)
```
hestolz, line 1193: seems maybe better to alter `install()` and `uninstall()` so that we call `exit_if_vm_not_supported` before `find_package_manager` ?

aswatt: Sounds good - I modified the check to be before `find_package_manager` in both options

PR #1675, opened 2023-02-08
Add RHEL9 Support

aswatt:
Adding support for RHEL 9 in AMA.


peizhen-w:
Could you please add new distros to AMA TST under `install/check_os.py` if that's convenient for you?

aswatt:
> Could you please add new distros to AMA TST under `install/check_os.py` if that's convenient for you?

Added the RHEL 9 support change to the AMA TST as well! I also included Amazon Linux 2, as it was recently included as well (#1674) but not included in that PR.

peizhen-w (approved):

PR #1683, opened 2023-02-28
MDC Recommendation updates for unsupported scenarios

Jyotsna-Anand:
Added updates to mark VMs unsupported by ADE and HBE as Compliant. 

Previously, these VMs were reported as Non-Compliant along with a reason for why the machine is not supported. Currently, MDC's custom logic to detect unsupported scenarios does not categorize the VMs correctly. The updates are a temporary work around to unblock for Public Preview. 

Also, updated "Host Based Encryption" feature name to "EncryptionAtHost" to align with customer facing documentation.

### DIFF
--- a/VMEncryption/mdc-recommendation/helper-scripts/MdcHandler.py
+++ b/VMEncryption/mdc-recommendation/helper-scripts/MdcHandler.py
@@ -76,9 +76,14 @@ def detect_encryption_status(imds_metadata):
     confidential_vm_status = VerifyVMSupport.is_confidential_vm(imds_metadata)
 
     if (confidential_vm_status is True):
-        is_vm_compliant = False
-        policy_reasons['code'] = "ConfidentialVMNotSupported"
+        is_vm_compliant = True
+        policy_reasons['code'] = "Compliant"
         policy_reasons['phrase'] = "Confidential VMs are not supported by this recommendation."
+
+        ##TO DO: As a temporary fix to report Confidential VMs as compliant, the below block has been commented. The  machine will be marked as compliant.
+        #is_vm_compliant = False
+        #policy_reasons['code'] = "ConfidentialVMNotSupported"
+        #policy_reasons['phrase'] = "Confidential VMs are not supported by this recommendation."
     else:
         is_vm_hbe_encrypted = check_hbe_encryption(imds_metadata)
         is_vm_ade_encrypted = check_ade_encryption(imds_metadata)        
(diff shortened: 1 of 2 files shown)


VMEncryption/mdc-recommendation/helper-scripts/MdcHandler.py
```diff
@@ -76,17 +76,22 @@ def detect_encryption_status(imds_metadata):
     confidential_vm_status = VerifyVMSupport.is_confidential_vm(imds_metadata)
 
     if (confidential_vm_status is True):
-        is_vm_compliant = False
-        policy_reasons['code'] = "ConfidentialVMNotSupported"
+        is_vm_compliant = True
+        policy_reasons['code'] = "Compliant"
         policy_reasons['phrase'] = "Confidential VMs are not supported by this recommendation."
```
vimish, line 81: Should this reason be empty?

VMEncryption/mdc-recommendation/helper-scripts/MdcHandler.py
```diff
@@ -95,14 +100,19 @@ def detect_encryption_status(imds_metadata):
             vm_support_dict = VerifyVMSupport.check_vm_supportability(imds_metadata)
 
             if vm_support_dict['is_vm_not_supported'] is True:
-                is_vm_compliant = False
-                policy_reasons['code'] = vm_support_dict['code']
+                is_vm_compliant = True
+                policy_reasons['code'] = "Compliant"
                 policy_reasons['phrase'] = vm_support_dict['phrase']
```
vimish, line 105: Same as above.

VMEncryption/mdc-recommendation/helper-scripts/VerifyVMSupport.py
```diff
@@ -1,164 +1,164 @@
...
+        reasons_code = "UltraDisksVirtualMachine"
+        reasons_phrase_str += "VM is using ultra disks which is currently not supported by Azure Disk Encryption and EncryptionAtHost."
+
+    if verify_vm_os_sku_completeness(imds_metadata) is False:
+        vm_support_dict['is_vm_not_supported'] = True
+        vm_support_dict['code'] = "UnknownOsVirtualMachine"
+        vm_support_dict['phrase'] = "VM OS details are missing in IMDS metadata for verification"
+        return vm_support_dict
+
+    else:
```
vimish, line 144: Is this else condition required? The function will anyway return if the above if condition is True.

Jyotsna-Anand: Agree, removed the else condition in latest commit.

VMEncryption/mdc-recommendation/helper-scripts/MdcHandler.py
Jyotsna-Anand: For the Linux GC package, we always display the reason for compliance or non-compliance since it's difficult and time consuming to debug the native binary. For example, for compliant VMs, we display why they are compliant (whether ADE or HBE is being used). Similarly, I thought it might be better to display the reason for being an unsupported scenario too. 

I have appended the reason for NotSupported scenarios in my latest commit to indicate that the VM is being marked as compliant since it is not supported.

vimish (approved):

PR #1684, opened 2023-03-09
Handling Negative Size reported to CRP

arisettisanjana:
There are certain distros like FortiOS where the size computation logic returns negative value due to the system commands used not being properly supported. 
As negative used space is not a valid value and the Caller is expecting a value greater than or equal to zero, marking the sizeComputationFailed flag to true and return the used space as zero

Changes made:
In SizeClaculation.py 

WorkItem :
https://msazure.visualstudio.com/One/_workitems/edit/17388438


VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -352,6 +352,11 @@ def get_total_used_size(self):
                         isKnownFs = True
                         break
 
+                if int(used) < 0 :
```
iamashwani-who, line 355: Can this change be added after line 336 instead? Any reason for running additional steps after calculating value of used variable?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -337,6 +337,12 @@ def get_total_used_size(self):
                 fstype = ''
                 isNetworkFs = False
                 isKnownFs = False
+
+                if int(used) < 0 :
+                    self.logger.log("The used space is negative, so marking the size computation as failed and returning zero")
+                    self.size_calc_failed = True
+                    return 0,self.size_calc_failed
```
ankitha-p, line 344: Nit: Please use whitespace after the comma ([Reference](https://peps.python.org/pep-0008/#:~:text=Immediately%20before%20a%20comma%2C%20semicolon%2C%20or%20colon%3A))

ankitha-p (approved):

PR #1688, opened 2023-03-23
Install/uninstall syslog local configurations based on syslog DCR

NarineM:
Create syslog watcher to monitor DCR changes for 3P to generate local syslog configs if syslog is enabled in DCR
No change for 1P, always generate local syslog configs as before.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -1195,8 +1195,8 @@ def generate_localsyslog_configs():
     Install local syslog configuration files if not present and restart syslog
     """
     if os.path.exists('/etc/rsyslog.d/') and not os.path.exists('/etc/rsyslog.d/10-azuremonitoragent.conf'):
-        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/05-azuremonitoragent-loadomuxsock.conf","/etc/rsyslog.d/05-azuremonitoragent-loadomuxsock.conf")
-        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/10-azuremonitoragent.conf","/etc/rsyslog.d/10-azuremonitoragent.conf")
+        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/rsyslogconf/05-azuremonitoragent-loadomuxsock.conf","/etc/rsyslog.d/05-azuremonitoragent-loadomuxsock.conf")
+        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/rsyslogconf/10-azuremonitoragent.conf","/etc/rsyslog.d/10-azuremonitoragent.conf")
         os.chmod('/etc/rsyslog.d/05-azuremonitoragent-loadomuxsock.conf', stat.S_IRGRP | stat.S_IRUSR | stat.S_IWUSR | stat.S_IROTH)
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -1087,6 +1149,82 @@ def metrics_watcher(hutil_error, hutil_log):
...
+            time.sleep(sleepTime)
+
+def generate_localsyslog_configs():
+    """
+    Install local syslog configuration files if not present and restart syslog
+    """
+    if os.path.exists('/etc/rsyslog.d/') and not os.path.exists('/etc/rsyslog.d/10-azuremonitoragent.conf'):
+        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/05-azuremonitoragent-loadomuxsock.conf","/etc/rsyslog.d/05-azuremonitoragent-loadomuxsock.conf")
+        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/10-azuremonitoragent.conf","/etc/rsyslog.d/10-azuremonitoragent.conf")
+        os.chmod('/etc/rsyslog.d/05-azuremonitoragent-loadomuxsock.conf', stat.S_IRGRP | stat.S_IRUSR | stat.S_IWUSR | stat.S_IROTH)
```
simathih, line 1200: would this be fine with both python 2 and 3?

AzureMonitorAgent/agent.py
```diff
@@ -1087,6 +1149,82 @@ def metrics_watcher(hutil_error, hutil_log):
...
+        os.chmod('/etc/rsyslog.d/10-azuremonitoragent.conf', stat.S_IRGRP | stat.S_IRUSR | stat.S_IWUSR | stat.S_IROTH)
+        run_command_and_log(get_service_command("rsyslog", "restart"))
+
+    if os.path.exists('/etc/syslog-ng/syslog-ng.conf') and not os.path.exists('/etc/syslog-ng/conf.d/azuremonitoragent.conf'):
+        syslog_ng_confpath = os.path.join('/etc/syslog-ng/', 'conf.d')
+        os.makedir(syslog_ng_confpath)
+        copyfile("/etc/opt/microsoft/azuremonitoragent/syslog/azuremonitoragent.conf","/etc/syslog-ng/conf.d/azuremonitoragent.conf")
```
simathih, line 1207: indent seems off here right?

simathih: also, do we need to keep syslog-ng config files separately under /etc/opt/microsoft/azuremonitoragent? essentially to avoid cx accidentally manually copying syslog-ng config into rsyslog

simathih (changes requested):

AzureMonitorAgent/agent.py
NarineM: I ll fix the indent in next iteration. thanks!

simathih (dismissed):

simathih (approved):

simathih:
I just realized this restarts rsyslog every 30s: if syslog_enabled:
                # place syslog local configs
                syslog_enabled  = False
                generate_localsyslog_configs()
            else:
                # remove syslog local configs
                remove_localsyslog_configs()

simathih:
Please generate config only if the marker file changed

PR #1695, opened 2023-04-04
Add Check to AMA Log Collector

aswatt:
Adds check to run `du -h -d 1` on AMA events folder. Also cleans up subprocess module import, as it isn't used.


AzureMonitorAgent/ama_tst/modules/logcollector.py
```diff
@@ -234,6 +234,18 @@ def create_outfile(output_dirpath, logs_date, pkg_manager):
             outfile.write("--------------------------------------------------------------------------------\n")
         outfile.write("--------------------------------------------------------------------------------\n")
 
+        # du output on events folder
+        for flag in ["", "--apparent-size"]:
+            du_full_cmd = DU_CMD.format(flag)
+            outfile.write("Output of command: {0}\n".format(du_full_cmd))
+            outfile.write("========================================\n")
+            outfile.write(helpers.run_cmd_output(du_full_cmd))
+            outfile.write("--------------------------------------------------------------------------------\n")
```
AaronRM, line 243: Why not match the starting line format ======= ?

AzureMonitorAgent/ama_tst/modules/logcollector.py
```diff
@@ -234,6 +234,18 @@ def create_outfile(output_dirpath, logs_date, pkg_manager):
             outfile.write("--------------------------------------------------------------------------------\n")
         outfile.write("--------------------------------------------------------------------------------\n")
 
+        # du output on events folder
+        for flag in ["", "--apparent-size"]:
+            du_full_cmd = DU_CMD.format(flag)
+            outfile.write("Output of command: {0}\n".format(du_full_cmd))
+            outfile.write("========================================\n")
+            outfile.write(helpers.run_cmd_output(du_full_cmd))
```
AaronRM, line 242: Do we need to append a newline after the cmd output?

AaronRM (approved):

AzureMonitorAgent/ama_tst/modules/logcollector.py
aswatt: The `=` line separates the command and its output, while the `-` line separates different commands; example below:

```
Output of command: ls -l
========================================
-rw-rw-r-- 1 aswatt aswatt 21275 Apr  4 16:22 ama_tst.tgz
--------------------------------------------------------------------------------
Output of command: pwd
========================================
/home/aswatt
--------------------------------------------------------------------------------```

I was trying to make it the most legible, but lmk if you think there's a better way to format this!

AzureMonitorAgent/ama_tst/modules/logcollector.py
aswatt: The `run_cmd_output()` function automatically adds a newline already

PR #1697, opened 2023-04-10
Removing temp files that are being created in extensions flow.

arisettisanjana:
So currently the present code at https://github.com/Azure/azure-linux-extensions/blob/20a56bd98a63953561a3066a0b58fba615e705d9/VMBackup/main/Utils/HandlerUtil.py#:~:text=f%20%3D%20tempfile.NamedTemporaryFile(delete%3DFalse) temporary files are being created to read the protected settings, once they have been read we are not removing or cleaning them.
In general they are cleaned up when reboot happens or the customer's VM cleans up the /tmp files every 10 days.
To improve this we are adding the logic to clean/remove them as soon as they read the protected settings.
Changes made in VMBackup/main/Utils/HandlerUtil.py


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -221,6 +221,12 @@ def _parse_config(self, ctxt):
                     self.error('JSON exception decoding ' + cleartxt)
                 handlerSettings['protectedSettings'] = jctxt
                 self.log('Config decoded correctly.')
+                # cleaning/removing the temp files created
+                try:
+                    if os.path.isfile(f.name):
+                        os.remove(f.name)
+                except Exception as e:
+                    self.error('Failed to remove the temporary file that was created for reading the protected settings' + e)
```
iamashwani-who, line 229: We can limit the error message to ('Failed to remove the temporary file ' + e)

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -221,6 +221,12 @@ def _parse_config(self, ctxt):
                     self.error('JSON exception decoding ' + cleartxt)
                 handlerSettings['protectedSettings'] = jctxt
                 self.log('Config decoded correctly.')
+                # cleaning/removing the temp files created
+                try:
+                    if os.path.isfile(f.name):
+                        os.remove(f.name)
```
iamashwani-who, line 227: it should be path than file name, correct?

arisettisanjana: the name attribute stores the absolute path.

iamashwani-who (approved):

VMBackup/main/Utils/HandlerUtil.py
ankitha-p, line 227: are we sure this file is not referenced anywhere else during the execution cycle of the code.

arisettisanjana: yes

rakrashba (approved):

PR #1698, opened 2023-04-10
Added more logs to debug issues

ankitha-p:
If there is any issue or extension gets stuck, currently it is difficult to pin-point the issue or line of code. adding more logs can help with this. hence added more log statements.


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -636,7 +640,7 @@ def write_to_status_file(self, stat_rept_file):
         try:
             tempStatusFile =  os.path.join(self._context._status_dir, CommonVariables.TempStatusFileName)
             if self._context._status_file:
-                with open(tempStatusFile,'w+') as f:
+                with open(tempStatusFile,'w') as f:
```
iamashwani-who, line 643: why this change?

ankitha-p: w+ opens the file in both read and write mode. Since we are only performing the write operation, opening in read mode is not required.

VMBackup/main/blobwriter.py
```diff
@@ -107,7 +107,9 @@ def WritePageBlob(self, message, blobUri, blobProperties):
                     http_util = HttpUtil(self.hutil)
                     sasuri_obj = urlparse.urlparse(blobUri + '&comp=page')
                     # Get Blob-properties to know content-length
-                    blobContentLength = int(blobProperties.contentLength)
+                    blobContentLength = 0
+                    if(blobProperties is not None):
+                        blobContentLength = int(blobProperties.contentLength)
                     self.hutil.log("WritePageBlob: contentLength:"+str(blobContentLength))
```
iamashwani-who, line 113: Is there an different handling that is required for blobProperties being NULL instead of letting the flow go through?

VMBackup/main/blobwriter.py
```diff
@@ -176,7 +178,9 @@ def ClearPageBlob(self, blobUri, blobProperties):
                     http_util = HttpUtil(self.hutil)
                     sasuri_obj = urlparse.urlparse(blobUri + '&comp=page')
                     # Get Blob-properties to know content-length
-                    contentLength = int(blobProperties.contentLength)
```
iamashwani-who, line 179: Is there an different handling that is required for blobProperties being NULL instead of letting the flow go through?

ankitha-p: If blobProperties is null. accessing contentLength property throws exception. Without knowing the contentLenght of the current blob, put_page_clear() from a definitive index is not possible. hence that logic should be skipped which is being taken care by setting the default value of contentLength = 0.

VMBackup/main/blobwriter.py
ankitha-p: skipping the blobWrite if the blob properties are null. this is the current behavior. the change I made will achieve the same without logging an exception.

rakrashba (approved):

VMBackup/main/backuplogger.py
```diff
@@ -48,7 +49,9 @@ def enforce_local_flag(self, enforced_local):
 
     """description of class"""
     def log(self, msg, local=False, level='Info'):
+        #self.hutil.log("Started backuplogger log function")
```
arisettisanjana, line 52: nit: we can remove this commented log

PR #1703, opened 2023-04-18
Fixed Bug when LUN list is empty for TVM,DD

arisettisanjana:
Issue description:
The current code while taking a snapshot for DD, TVM etc scenarios when LUN list is empty, and some disks are being excluded it calls Guest to take the snapshot. But these scenarios are handled by Host to take snapshot.
Now as we don't have the LUN no's of which disks to take snapshot, Host will be taking the snapshot of all disks that results in inconsistency with the one CRP expects.

Fix description:
So, will be failing the Extension in these scenarios saying there is a mismatch in the inputs.

Changes made in the following files:
VMBackup/main/ExtensionErrorCodeHelper.py, VMBackup/main/common.py to add the new error code
VMBackup/main/freezesnapshotter.py added checks to handle the above scenarios
VMBackup/main/handle.py to exit the extension with the error


VMBackup/main/ExtensionErrorCodeHelper.py
```diff
@@ -42,7 +42,8 @@ class ExtensionErrorCodeEnum():
     FailedPrepostPluginConfigPermissionError = 317
     FailedPrepostPluginConfigOwnershipError = 318
     FailedGuestAgentInvokedCommandTooLate = 402
-    
+    FailedInputMismatch = 405
```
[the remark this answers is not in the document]
rakrashba: Are these error codes supposed to corelate to http status codes? If they do a simple 400 should work.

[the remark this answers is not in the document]
arisettisanjana: As we are failing the extension specifically because there is a mismatch between the expected and received parameters, I think this error code would be able to point that.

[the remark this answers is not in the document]
iamashwani-who: Input mismatch can happen due to multiple conditions. The name should specifically mention the condition that is failing.

VMBackup/main/freezesnapshotter.py
```diff
@@ -129,8 +129,24 @@ def __init__(self, logger, hutil , freezer, g_fsfreeze_on, para_parser, takeCras
...
+                        errMsg = 'VmgsBlob is included, so the host must create the snapshot. IsAnyDiskExcluded is true, but, the included LUN list is empty in the extension input, '\
+                                    'which is not allowed for host DoSnapshot. Thus, failing the extension run.'
+                        self.logger.log(errMsg, True, 'Error')
+                        self.hutil.SetExtErrorCode(ExtensionErrorCodeHelper.ExtensionErrorCodeEnum.FailedInputMismatch)
+                    elif( CommonVariables.isAnyWADiskIncluded in para_parser.includedDisks.keys() and para_parser.includedDisks[CommonVariables.isAnyWADiskIncluded] == True):
+                        errMsg = 'WADisk is included, so the host must create the snapshot. IsAnyDiskExcluded is true, but, the included LUN list is empty in the extension input, '\
+                                    'which is not allowed for host DoSnapshot. Thus, failing the extension run.'
+                        self.logger.log(errMsg, True, 'Error')
+                        self.hutil.SetExtErrorCode(ExtensionErrorCodeHelper.ExtensionErrorCodeEnum.FailedInputMismatch)
+                    else:
```
iamashwani-who, line 147: These 3 conditions should return different error codes so that they can be differentiated.

VMBackup/main/freezesnapshotter.py
```diff
@@ -129,8 +129,24 @@ def __init__(self, logger, hutil , freezer, g_fsfreeze_on, para_parser, takeCras
...
+                    elif( CommonVariables.isVmgsBlobIncluded in para_parser.includedDisks.keys() and para_parser.includedDisks[CommonVariables.isVmgsBlobIncluded] == True):
+                        errMsg = 'VmgsBlob is included, so the host must create the snapshot. IsAnyDiskExcluded is true, but, the included LUN list is empty in the extension input, '\
+                                    'which is not allowed for host DoSnapshot. Thus, failing the extension run.'
+                        self.logger.log(errMsg, True, 'Error')
+                        self.hutil.SetExtErrorCode(ExtensionErrorCodeHelper.ExtensionErrorCodeEnum.FailedInputMismatch)
+                    elif( CommonVariables.isAnyWADiskIncluded in para_parser.includedDisks.keys() and para_parser.includedDisks[CommonVariables.isAnyWADiskIncluded] == True):
+                        errMsg = 'WADisk is included, so the host must create the snapshot. IsAnyDiskExcluded is true, but, the included LUN list is empty in the extension input, '\
+                                    'which is not allowed for host DoSnapshot. Thus, failing the extension run.'
```
iamashwani-who, line 144: "IsAnyDiskExcluded is true, but, the included LUN list is empty in the extension input,"
Please describe the reasoning for this in comments. It is not clear.

VMBackup/main/ExtensionErrorCodeHelper.py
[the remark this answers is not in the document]
ankitha-p: Hey @arisettisanjana , I still think using 'mismatch' in the error code can be a misnomer. A good example of mismatch would be-> to have one taskId in input but different taskId in output for the same run. This would be a mismatch. 
This DataDiskLunList field being empty when atleast one disk is excluded is a case of poor input creation by the requestor. 
Some suggestions: 'FailedInvalidDataDiskLunList' or 'FailedInconsistentDataDiskLunList'

ankitha-p, line 45: Usually the error code number ranges are grouped based on categories like GuestAgent Error codes, fsfreeze errorcodes, workload failures etc. Let's also make sure the codes are consistent across windows and Linux. Looks like this file has error codes between 11-20 for incorrect input like errors. You can look at adding the new error code within this range.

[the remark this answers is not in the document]
arisettisanjana: sure

arisettisanjana: sure will add a code between that

VMBackup/main/freezesnapshotter.py
arisettisanjana: sure will add that

VMBackup/main/freezesnapshotter.py
arisettisanjana: Correct me if i am wrong 
I think by the error code we can come to know that there was a failure because of Lun list was not passed and Logs we can come to know that there was that particular type of disk while taking the snapshot hence failing it.

VMBackup/main/handle.py
```diff
@@ -322,7 +327,11 @@ def daemon():
             canTakeCrashConsistentSnapshot = can_take_crash_consistent_snapshot(para_parser)
             temp_g_fsfreeze_on = True
             freeze_snap_shotter = FreezeSnapshotter(backup_logger, hutil, freezer, temp_g_fsfreeze_on, para_parser, canTakeCrashConsistentSnapshot)
-
+            if (hutil.ExtErrorCode == ExtensionErrorCodeHelper.ExtensionErrorCodeEnum.FailedInvalidDataDiskLunList):
+                temp_result = CommonVariables.FailedInvalidDataDiskLunList
+                temp_status = 'error'
+                error_msg = "Input mismatch CRP did not send the LUN No's of disks to be included"
```
iamashwani-who, line 333: this is not a mismatch but an invalid input case, right?

Invalid input. IsAnyDiskExcluded is marked as true but input LUN list is empty. It is not allowed if VM has Direct Drives or if VM has Write Accelerated disks or if VM is a TVM/CVM.

rakrashba (changes requested):

iamashwani-who (approved):

ankitha-p (approved):

PR #1714, opened 2023-05-11
Handle Linux Hardening where sudo on root is blocked

simathih:


LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -94,6 +94,8 @@ def stop_metrics_service(is_lad):
 
         if os.path.isfile(metrics_service_path):
             code = os.system("sudo systemctl stop {0}".format(metrics_service_name))
+            if code != 0:
+                code = os.system("systemctl stop {0}".format(metrics_service_name))
```
hestolz, line 98: do we even need sudo? I do not see it used in `agent.py`

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -352,8 +356,10 @@ def start_metrics(is_lad):
     if metrics_utils.is_systemd():
         service_restart_status = os.system("sudo systemctl restart {0}".format(metrics_service_name))
         if service_restart_status != 0:
-            log_messages += "Unable to start {0}. Failed to start ME service.".format(metrics_service_name)
-            return False, log_messages
+            service_restart_status = os.system("systemctl restart {0}".format(metrics_service_name))
+            if service_restart_status != 0:
+                log_messages += "Unable to start {0}. Failed to start ME service.".format(metrics_service_name)
```
NarineM, line 361: Looks good, can we append the error message to say "Unable to start ......" + ". Check if the system has some custom hardening or policies preventing systemctl command"

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
simathih: this code is shared with LAD, I need to check if LAD needs sudo

simathih: Nikhil added the sudo, should we just remove it and see if it breaks anything? :) I checked LAD code and that should be running
 as root as well so sudo shouldn't be needed

NarineM: They can have the opposite when the user is not the root, in this case it should run with sudo.

simathih: user is always root, Guest agent invokes our extension code with root

simathih: let me remove sudo and I feel we should be fine

NarineM: waagent may be, but we are using shim to apply manual workarounds. Is this code also called from shim?

simathih: manual workaround can do sudo shim.sh instead...that is for internal use only

NarineM: ok, but we will need to check with CSS tsg and update the steps . I am almost certain we have shim in troubleshooting.

NarineM (approved):

PR #1715, opened 2023-05-14
Use tcp port for syslog if configured

simathih:


AaronRM (dismissed):

NarineM (approved):

AzureMonitorAgent/agent.py
```diff
@@ -1192,34 +1194,112 @@ def generate_localsyslog_configs():
...
-        run_command_and_log(get_service_command("syslog-ng", "restart"))
-        hutil_log_info("Installed local syslog configuration files and restarted syslog")
+    public_settings, _ = get_settings()
+    syslog_port = ''
+    if os.path.isfile(AMASyslogPortFilePath):
+        f = open(AMASyslogPortFilePath, "r")
+        syslog_port = f.read()
+        f.close()
+
+    if public_settings is not None and "useOmfwd" in public_settings and syslog_port != '':
```
Dmitry-Matveev, line 1205: We should be very careful with publicly facing settings / contracts (e.g. "useOmfwd") as these will be set in stone once shipped or documented (and changing / removing them is a breaking change). Have we agreed on naming internally with PM?

simathih: I am not sure if we want to document this as we only want to keep it behind a flag only for a release or two but I will ask PM team. I have opened another PR to fix the name: https://github.com/Azure/azure-linux-extensions/pull/1720

PR #1717, opened 2023-05-15
VMAccess support for Flatcar distro

mayankdaruka-msft:
- Add default linux agent config filename for CoreOS (incl. Flatcar) distros to correctly retrieve Provisioning.PasswordCryptId and Provisioning.PasswordCryptSaltLength values
- Provide default values for PasswordCryptId and PasswordCryptSaltLength in case config file does not exist at provided location
- New flag in parse() function in ovfutils to get rid of dependency on Provisioning.DecodeCustomData attribute within VMAccess

### DIFF
--- a/Utils/distroutils.py
+++ b/Utils/distroutils.py
@@ -10,9 +10,15 @@ import Utils.extensionutils as ext_utils
 import Utils.constants as constants
 
 
-def get_my_distro(os_name, config):
+def get_my_distro(config, os_name=None):
     if 'FreeBSD' in platform.system():
         return FreeBSDDistro(config)
+    
+    if os_name is None:
+        if os.path.isfile(constants.os_release):
+            os_name = ext_utils.get_line_starting_with("NAME", constants.os_release)
+        elif os.path.isfile(constants.system_release):
+            os_name = ext_utils.get_file_contents(constants.system_release)
 
     if os_name is not None:
         if re.search("fedora", os_name, re.IGNORECASE):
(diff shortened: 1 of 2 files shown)


Utils/distroutils.py
```diff
@@ -10,16 +10,10 @@
 import Utils.constants as constants
 
 
-def get_my_distro(config):
+def get_my_distro(os_name, config):
```
D1v38om83r, line 13: Overload this function instead of changing the signature. Other extensions might be making use of Utils.py, and they can't catch the changes at compile time because Python

Utils/ovfutils.py
```diff
@@ -130,18 +130,20 @@ def parse(xml_text, configuration, is_deprovision=False):
...
+                if len(cd_section) > 0:
+                    ovf_env.CustomData = get_node_text_data(cd_section[0])
+                    if len(ovf_env.CustomData) > 0:
+                        ext_utils.set_file_contents(constants.LibDir + '/CustomData', bytearray(
+                            translate_custom_data(ovf_env.CustomData, configuration)))
+                        logger.log('Wrote ' + constants.LibDir + '/CustomData')
+                    else:
+                        logger.error('<CustomData> contains no data!')
+            except Exception as e:
+                logger.error(str(e) + ' occured creating ' + constants.LibDir + '/CustomData')
```
D1v38om83r, line 145: when is write_custom_data expected to be false?

D1v38om83r (changes requested):

D1v38om83r (approved):

PR #1724, opened 2023-05-23
fix syslog conf generation for auto config

NarineM:


AzureMonitorAgent/agent.py
```diff
@@ -464,8 +464,12 @@ def enable():
             log_and_exit("Enable", MissingorInvalidParameterErrorCode, 'Mixing genevaConfiguration or azureMonitorConfiguration with other configuration schemas is not allowed')
 
         if geneva_configuration and geneva_configuration.get("enable") == True:
-            hutil_log_info("Detected Geneva+ mode; azuremonitoragentmgr service will be started to handle Geneva tenants")
             ensure["azuremonitoragentmgr"] = True
+            if azure_monitor_configuration and azure_monitor_configuration.get("enable") == True:
```
hestolz, line 468: this log change is unnecessary, we log wheter AMCS mode is also enabled below on original line 471. please revert to the original, and only add the `generate_localsyslog_configs()` to this block  (block starting line 466)

hestolz (changes requested):

AzureMonitorAgent/agent.py
```diff
@@ -478,6 +482,8 @@ def enable():
     elif public_settings is not None and public_settings.get("GCS_AUTO_CONFIG") == True:
         hutil_log_info("Detected Auto-Config mode; azuremonitoragentmgr service will be started to handle Geneva tenants")
         ensure["azuremonitoragentmgr"] = True
+        # generate local syslog configuration files as in auto config syslog is not driven from DCR
+        generate_localsyslog_configs()
```
hestolz, line 486: correct, keep this for autoconfig

AzureMonitorAgent/agent.py
```diff
@@ -466,6 +466,8 @@ def enable():
         if geneva_configuration and geneva_configuration.get("enable") == True:
             hutil_log_info("Detected Geneva+ mode; azuremonitoragentmgr service will be started to handle Geneva tenants")
             ensure["azuremonitoragentmgr"] = True
+            if not azure_monitor_configuration or (azure_monitor_configuration and azure_monitor_configuration.get("enable") != True):
```
hestolz, line 469: We don't need this condition; if a customer is using geneva AND amcs we still should generate the omuxsock module file

simathih (approved):

hestolz (approved):

PR #1725, opened 2023-05-29
Case insensitivity of Settings

arisettisanjana:
Need for change:
Case Insensitive validation is required at extensions for the settings passed by CRP.
Changes made:
parameterparser.py

### DIFF
--- a/VMBackup/main/parameterparser.py
+++ b/VMBackup/main/parameterparser.py
@@ -36,6 +36,7 @@ class ParameterParser(object):
         self.includedDisks = None
         self.dynamicConfigsFromCRP = None
         self.wellKnownSettingFlags = {CommonVariables.isSnapshotTtlEnabled: False, CommonVariables.useMccfToFetchDsasForAllDisks: False, CommonVariables.useMccfForLad: False}
+        self.settingkeys = {"issnapshotttlenabled": CommonVariables.isSnapshotTtlEnabled, "usemccftofetchdsasforalldisks": CommonVariables.useMccfToFetchDsasForAllDisks, "usemccfforlad": CommonVariables.useMccfForLad}
         self.includeLunList = []    #To be shared with HP
 
         """
@@ -112,9 +113,9 @@ class ParameterParser(object):
                 for config in self.dynamicConfigsFromCRP:
                     if CommonVariables.key in config and CommonVariables.value in config:
                         found = False
-                        for flag in self.wellKnownSettingFlags:
-                            if((config[CommonVariables.key]).lower() == flag.lower()):
-                                self.wellKnownSettingFlags[flag] = config[CommonVariables.value]
+                        for flag in self.settingkeys:
+                            if((config[CommonVariables.key]).lower() == flag):
+                                self.wellKnownSettingFlags[self.settingkeys[flag]] = config[CommonVariables.value]
                                 found = True
                                 break
                         if(found != True):
--- a/VMBackup/main/parameterparser.py
+++ b/VMBackup/main/parameterparser.py
@@ -36,7 +36,7 @@ class ParameterParser(object):
         self.includedDisks = None
         self.dynamicConfigsFromCRP = None
         self.wellKnownSettingFlags = {CommonVariables.isSnapshotTtlEnabled: False, CommonVariables.useMccfToFetchDsasForAllDisks: False, CommonVariables.useMccfForLad: False}
-        self.settingkeys = {"issnapshotttlenabled": CommonVariables.isSnapshotTtlEnabled, "usemccftofetchdsasforalldisks": CommonVariables.useMccfToFetchDsasForAllDisks, "usemccfforlad": CommonVariables.useMccfForLad}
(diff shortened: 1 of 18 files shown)


VMBackup/main/parameterparser.py
```diff
@@ -111,7 +111,7 @@ def __init__(self, protected_settings, public_settings, backup_logger):
                 backup_logger.log("settings received " + str(self.dynamicConfigsFromCRP), True)
                 for config in self.dynamicConfigsFromCRP:
                     if CommonVariables.key in config and CommonVariables.value in config:
-                        if(config[CommonVariables.key] in self.wellKnownSettingFlags):
+                        if(config[CommonVariables.key].lower() in self.wellKnownSettingFlags.lower()):
```
iamashwani-who, line 114: why not keep all setting names as toLower in wellKnownSettingFlags instead of performing lower() operation for each key?

arisettisanjana: The setting names while sending to BHS should be in camel case format like isSnapshotTtlEnabled.

iamashwani-who: Then line 115 is going to have a problem because we are picking key value as config["key"] which will not match as it is not lowercase?

VMBackup/main/parameterparser.py
```diff
@@ -111,10 +111,14 @@ def __init__(self, protected_settings, public_settings, backup_logger):
...
-                        if(config[CommonVariables.key] in self.wellKnownSettingFlags):
-                            self.wellKnownSettingFlags[config[CommonVariables.key]] = config[CommonVariables.value]
-                        else:
-                            backup_logger.log("The received " + str(config[CommonVariables.key]) + " is not an expected setting name.", True)
+                        found = False
+                        for flag in self.wellKnownSettingFlags:
+                            if((config[CommonVariables.key]).lower() == flag.lower()):
+                                self.wellKnownSettingFlags[flag] = config[CommonVariables.value]
+                                found = True
+                                break
```
iamashwani-who, line 119: It might be more efficient to have another small Dictionary with lowercase name as key and actual key as value. Then you can simply look up if the lowercase key passed from CRP exists and if yes then you can assign the value by using the value as the key.

arisettisanjana: Yep, repeated conversion of the keys(flag) into lower can be avoided at the cost of space.

VMBackup/main/parameterparser.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, protected_settings, public_settings, backup_logger):
         self.includedDisks = None
         self.dynamicConfigsFromCRP = None
         self.wellKnownSettingFlags = {CommonVariables.isSnapshotTtlEnabled: False, CommonVariables.useMccfToFetchDsasForAllDisks: False, CommonVariables.useMccfForLad: False}
+        self.settingkeys = {"issnapshotttlenabled": CommonVariables.isSnapshotTtlEnabled, "usemccftofetchdsasforalldisks": CommonVariables.useMccfToFetchDsasForAllDisks, "usemccfforlad": CommonVariables.useMccfForLad}
```
iamashwani-who, line 39: can be named as settingKeyMapping

VMBackup/main/parameterparser.py
```diff
@@ -111,10 +112,14 @@ def __init__(self, protected_settings, public_settings, backup_logger):
                 backup_logger.log("settings received " + str(self.dynamicConfigsFromCRP), True)
                 for config in self.dynamicConfigsFromCRP:
                     if CommonVariables.key in config and CommonVariables.value in config:
-                        if(config[CommonVariables.key] in self.wellKnownSettingFlags):
-                            self.wellKnownSettingFlags[config[CommonVariables.key]] = config[CommonVariables.value]
-                        else:
-                            backup_logger.log("The received " + str(config[CommonVariables.key]) + " is not an expected setting name.", True)
+                        found = False
```
iamashwani-who, line 115: This loop is not required anymore as we can use the settingKeyMapping dictionary for direct lookup

iamashwani-who (approved):

VMBackup/main/parameterparser.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, protected_settings, public_settings, backup_logger):
         self.includedDisks = None
         self.dynamicConfigsFromCRP = None
         self.wellKnownSettingFlags = {CommonVariables.isSnapshotTtlEnabled: False, CommonVariables.useMccfToFetchDsasForAllDisks: False, CommonVariables.useMccfForLad: False}
+        self.settingKeysMapping= {"issnapshotttlenabled": CommonVariables.isSnapshotTtlEnabled, "usemccftofetchdsasforalldisks": CommonVariables.useMccfToFetchDsasForAllDisks, "usemccfforlad": CommonVariables.useMccfForLad}
```
rakrashba, line 39: 1. instead of hardcoding the names in lowercase here maybe do something like
```python
self.settingKeyMappings = {}
self.settingKeyMappings[CommonVariables.isSnapshotTtlEnabled.lower()] = CommonVariables.isSnapshotTtlEnabled
# ...
# and so on
```
That way the any change to the value associated with the key needs to be done at one place, in case the key itself does not change.

2. Also review if this value needs to be a class level attribute. Maybe use a local variable if this should/need not be accessible outside this class.

rakrashba (changes requested):

iamashwani-who (approved):

ankitha-p (approved):

PR #1732, opened 2023-06-09
Fix proxy setting schema

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -688,7 +688,7 @@ def handle_mcs_config(public_settings, protected_settings, default_configs):
         else:
             log_and_exit("Enable", MissingorInvalidParameterErrorCode, 'Parameter "address" is required in proxy public setting')
 
-        if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == "true":
+        if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == true:
```
Dmitry-Matveev, line 691: It looks like whichever way we choose, we may need to change [documentation here](https://learn.microsoft.com/en-us/azure/azure-monitor/agents/azure-monitor-agent-data-collection-endpoint?tabs=PowerShellWindowsArc#proxy-configuration). Examples for ARC Windows and Linux are with "true", examples for VM Windows and Linux are with boolean true.

Dmitry-Matveev (approved):

PR #1733, opened 2023-06-09
 Merge LAd 4.1.10 release branch into master to baseline master

NarineM:
Lad 4.0 release main branch: https://github.com/Azure/azure-linux-extensions/tree/release/lad4.0
Last LAD 4 release 4.1.10 tag: https://github.com/Azure/azure-linux-extensions/releases/tag/lad-4.1.10
 LAd 4.1.10 release branch codebase: https://github.com/Azure/azure-linux-extensions/tree/narinem-lad-4.1.10

### DIFF
--- a/Diagnostic/HandlerManifest.json
+++ b/Diagnostic/HandlerManifest.json
@@ -2,11 +2,11 @@
   {
     "version": 1.0,
     "handlerManifest": {
-      "installCommand": "./shim.sh -install",
-      "uninstallCommand": "./shim.sh -uninstall",
-      "updateCommand": "./shim.sh -update",
-      "enableCommand": "./shim.sh -enable",
-      "disableCommand": "./shim.sh -disable",
+      "disableCommand": "shim.sh -disable",
+      "enableCommand": "shim.sh -enable",
+      "installCommand": "shim.sh -install",
+      "uninstallCommand": "shim.sh -uninstall",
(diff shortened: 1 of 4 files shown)


AaronRM:
[LAD 4 requires with Python 2](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/diagnostics-linux?tabs=azcli#prerequisites), so not clear we should be including Python 3 as a fallback in shim.sh. Check with @simathih who [made the changes in master](#1447) whether we should be keeping the Python 3 fallback.

AaronRM (approved):

NarineM:
I see latest commit from Henry (July 20220) changing shim to only allow python2 : https://github.com/Azure/azure-linux-extensions/commit/e9ec1e59b2e973cbfae984993ef9a370e062e51e#diff-db8ac4ee7c5c68c2099db90a6de2f18c06eae2217541914ad5ab4ee335efb16e
Note that this is what ships in Lad 4 releases.

 2 years older commit by Sidd for python 3 was merged with this commit: https://github.com/Azure/azure-linux-extensions/commit/062212a7b3cb24b663c9c3bd143a937555b788ca

Given the dates of these commits and the fact that we know about python 2 dependency from ICMs, I think we should keep Henry's change (same change also in this PR).
@simathih @hestolz  please let me know otherwise.

simathih (approved):

PR #1734, opened 2023-06-09
Support Arc-Enabled Servers proxy configuration file

simathih:
per https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/233012/VM-Extension-Authoring-for-Arc?anchor=4.-proxy-support


AzureMonitorAgent/agent.py
```diff
@@ -697,7 +698,20 @@ def handle_mcs_config(public_settings, protected_settings, default_configs):
...
+    # is this Arc? If so, check for proxy     
+    if os.path.isfile(ArcSettingsFile):
+        f = open(ArcSettingsFile, "r")
+        data = f.read()
+
+        if (data != ''):
+            json_data = json.loads(data)
+            if "proxy.url" in json_data:
+                url = json_data["proxy.url"]
+                if url != '':
```
NarineM, line 711: Need error handling when json_data is null: 
if json_data is not None and "proxy.url" in json_data:

and error handling when proxy.url is specified but empty or invalid similar to line#698. If ignoring invalid url is intentional we would not need error handling but we will likely need it to be logged

simathih: yeah, the only validation we do is to check if its non-empty, cx has to ensure config is correct

NarineM (approved):

PR #1736, opened 2023-06-23
Fix az_resource_id for flexible VMSS

simathih:


LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -750,6 +750,13 @@ def handle_config(config_data, me_url, mdsd_url, is_lad):
     if "virtualMachineScaleSets" in az_resource_id: 
         az_resource_id = "/".join(az_resource_id.split("/")[:-2])
 
+    virtual_machine_name = ""
```
NarineM, line 753: Can you add similar to previous section comment that this is needed specifically for ME ? and what ME expect the id to be in case of flexible after parsing? Example of resource id after parsing that ME can handle

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -750,6 +750,13 @@ def handle_config(config_data, me_url, mdsd_url, is_lad):
     if "virtualMachineScaleSets" in az_resource_id:
```
NarineM, line 750: I think this comment needs to be changed, it gives impression it is about vmss id, but should have been vmss instance resource id or examples are wrong:
...... If the instance is VMSS then trim the last two values from the resource id ie - "/virtualMachines/0"
   Since ME expects the resource id in a particular format. For egs -
    IMDS returned ID - /subscriptions/<sub-id>/resourceGroups/<rg_name>/providers/Microsoft.Compute/virtualMachineScaleSets/<VMSSName>/virtualMachines/0
  ME expected ID- /subscriptions/<sub-id>/resourceGroups/<rg_name>/providers/Microsoft.Compute/virtualMachineScaleSets/<VMSSName>......................
    
 VMSS resource ID only is formatted to contain VMSS name, when VMSS instance resource ID is the VMSS child instance ID, which has /0/ ending depending on the instance name. Is ME only taking VMSS ID ? if yes, why we require setting for LAD to point to VMSS instance resource ID which contains /0/ at the end and not directly VMSS ID?

NarineM (changes requested):

LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
simathih: updated the comments

NarineM (approved):

PR #1750, opened 2023-07-06
Using Boolean True instead of true

### LINKED ISSUE #1749: Azure Monitor Agent Linux failing to install 1.27.2
A recent change created an issue with the agent on linux. This should be:
`if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == True:` and not `if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == true:`

Here is the error:

```
2023/07/06 16:45:20 ERROR:    if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == true:
2023/07/06 16:45:20 ERROR:NameError: name 'true' is not defined
```

https://github.com/Azure/azure-linux-extensions/blob/ac28d63cfa7991559563ee9fa228f527079240c2/AzureMonitorAgent/agent.py#L692C96-L692C101

Introduced by: https://github.com/Azure/azure-linux-extensions/pull/1732

lacj01-od:
Fixed typo in Boolean value: The lowercase *true* is not a valid identifier in python. This will use the proper *True* instead.

Fixes #1749

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -689,7 +689,7 @@ def handle_mcs_config(public_settings, protected_settings, default_configs):
         else:
             log_and_exit("Enable", MissingorInvalidParameterErrorCode, 'Parameter "address" is required in proxy public setting')
 
-        if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == true:
+        if "auth" in public_settings.get("proxy") and public_settings.get("proxy").get("auth") == True:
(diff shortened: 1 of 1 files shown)


simathih (approved):

kaovd:
Hi - I just hit this issue in prod - as someone who hasn't had insight into how this repository and release trains work, while I don't mean to bug is there an ETA when is this going to come downstream? 1.27.2 is still currently the latest.
@simathih

simathih:
we will start rollout of 1.27.3 this week

PR #1753, opened 2023-07-07
stop start issue fix for EnableEncryptionFormatAll

pankajosh:
fix for [Incident 384350526](https://icmcdn.akamaized.net/imp/v3/incidents/details/384350526/home) : Investigate latency Issues with Compute Instance LROs.
resource disk re-encryption fix for stop start use case.

### DIFF
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -751,7 +751,11 @@ def enable():
         encryption_config = EncryptionConfig(encryption_environment=encryption_environment, logger=logger)
         if encryption_config.config_file_exists():
             existing_volume_type = encryption_config.get_volume_type()
-
+        #log to capture lsblk before encryption view.
+        disk_util.log_lsblk_output()
+        if public_settings.get(CommonVariables.EncryptionEncryptionOperationKey) == CommonVariables.EnableEncryptionFormatAll:
+            #in case of stop start unmount /mnt to avoid resource disk encryption issues.
+            disk_util.umount('/mnt')
         is_migrate_operation = False
         if CommonVariables.MigrateKey in public_settings:
             if public_settings.get(CommonVariables.MigrateKey) == CommonVariables.MigrateValue:


VMEncryption/main/handle.py
```diff
@@ -751,7 +751,11 @@ def enable():
         encryption_config = EncryptionConfig(encryption_environment=encryption_environment, logger=logger)
         if encryption_config.config_file_exists():
             existing_volume_type = encryption_config.get_volume_type()
-
+        #log to capture lsblk before encryption view.
+        disk_util.log_lsblk_output()
+        if public_settings.get(CommonVariables.EncryptionEncryptionOperationKey) == CommonVariables.EnableEncryptionFormatAll:
```
canfikret, line 756: Is the incident reported for EncryptFormalAll scenarios only?

canfikret (changes requested):

VMEncryption/main/handle.py
pankajosh: yes, in standard VM resource disk encryption is supported only for  EncryptFormalAll scenario. issue has been seen on stop start.

bamuir: Is the resource disk still getting encrypted if it is unmounted?

pankajosh: yes, resource disk got encrypted if EFA is there, irrespective of disk mount. 
I had explained the issue in the below page. 
https://microsoft.sharepoint.com/teams/AzureSecurityCompliance/Security/_layouts/15/Doc.aspx?sourcedoc={c820c6d1-ba1e-45d8-bd74-5e103f5ac176}&action=edit&wd=target%28LiveSite.one%7Ccd0e0f5b-0d53-4726-81f4-85b86d4cfa01%2FAnalysis%20384350526%20Investigate%20latency%20Issues%20with%20Compute%20Instance%20LROs%7Cd38eeb58-1889-41cb-8e92-21bbb8cef4ff%2F%29&wdorigin=703

VMEncryption/main/handle.py
```diff
@@ -751,7 +751,11 @@ def enable():
...
+            #in case of stop start unmount /mnt to avoid resource disk encryption issues.
+            disk_util.umount('/mnt')
```
canfikret, line 758: where will the resource disk get mounted again?

Have you considered creating a separate mount point for BEK which is not under /mnt? For example, /azure_bek_disk ?

canfikret (changes requested):

VMEncryption/main/handle.py
canfikret: Perhaps, we should consider mounting BEK volume on /bekvolume

bamuir: @canfikret yeah, we have been wanting to move it. For new VM's, it is an easy change, but for existing VM's, snapshots/backups/etc we would need to think about a solution that wont break those.

PR #1760, opened 2023-07-17
FlagToConfiguresudo

arisettisanjana:
Need:
Since some customers encounter issues when running commands with sudo privileges, and the existing logic runs a few commands with sudo privileges.
Changes:
We have introduced a new flag that customers can set in the /etc/azure/vmbackup.conf file to avoid running the commands with sudo privileges.
By default, the behavior remains unchanged, and the commands are executed with sudo access if the flag is off.

Changes have been made in SizeCalculation.py
to read the setting from the config file. This allows for more flexibility and control over the command execution.


iamashwani-who:
I think this setting should apply to all places where sudo is used, and not be limited to exclude disk size calculations only.

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -68,11 +69,15 @@ def __init__(self,patching, hutil, logger,para_parser):
             self.logger.log("As the LunList is empty including all disks")
     
     def get_lsscsi_list(self):
-        command = "sudo lsscsi"
+        if (self.sudo_off == 1):
```
iamashwani-who, line 72: nit: it could be written as following too.
command = "lsscsi"
if(self.sudo_off == 0)
     command = "sudo " + command

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -177,6 +186,10 @@ def device_list_for_billing(self):
             Since root devices are at mount points '/' and '/boot/efi' we use file_system_info to find the root_devices based on the mount points.
         '''
 
+        # check if user off the sudo usage in commands
+        self.sudo_off = self.hutil.get_intvalue_from_configfile("sudo_off", self.sudo_off)
```
iamashwani-who, line 190: does this use default value (0?) if conf does not have the setting?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -128,7 +137,7 @@ def disk_list_for_billing(self):
             self.logger.log("Disks to be included {0}".format(self.disksToBeIncluded))
         else:
             self.size_calc_failed = True
-            self.logger.log("There is some glitch in executing the command 'sudo lsscsi' and therefore size calculation is marked as failed.")
+            self.logger.log("There is some glitch in executing the command 'lsscsi' and therefore size calculation is marked as failed.")
```
iamashwani-who, line 140: command name is pv, not lsscsi, right?

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -68,11 +69,15 @@ def __init__(self,patching, hutil, logger,para_parser):
...
+        if (self.sudo_off == 1):
+            command = "lsscsi"
+        else:
+            command = "sudo lsscsi"
         try:
+            self.logger.log("executing command  {0}".format(command))
             self.lsscsi_list = (os.popen(command).read()).splitlines()
         except Exception as e:
-            error_msg = "Failed to execute the command lsscsi because of error %s , stack trace: %s" % (str(e), traceback.format_exc())
+            error_msg = "Failed to execute the command %s because of error %s , stack trace: %s" % (command, str(e), traceback.format_exc())
```
rakrashba, line 80: nit: maybe quotes around the `%s` placeholders: `\"%s\"`

VMBackup/main/Utils/SizeCalculation.py
```diff
@@ -57,6 +57,7 @@ def __init__(self,patching, hutil, logger,para_parser):
         self.isAnyDiskExcluded = False
         self.LunListEmpty = False
         self.logicalVolume_to_bill = []
+        self.sudo_off = 0 # run the commands using sudo
```
rakrashba, line 60: Suggestion:

Since we are including this config - we can do a better default initialization and check before even allowing toggle of this config. Checking if sudo is installed can be done with:
```bash
command -v sudo
```
The above will output an empty string if `sudo` is not found in PATH - which almost always means that it isn't installed.
One problem that could arise though is if the command `command` itself is not available - I have never faced this in practice however. LISA should be able to help us verify this if we decide to go ahead with this.

VMBackup/main/Utils/SizeCalculation.py
arisettisanjana: Yes

VMBackup/main/Utils/SizeCalculation.py
arisettisanjana: since our extension runs with sudo access, checking for the existence of sudo within the extension itself may not be necessary. 
However, in LISA it might be a good idea to include the check for sudo existence.

iamashwani-who (approved):

rakrashba (approved):

PR #1765, opened 2023-07-20
remove python-future from oms extension

NarineM:
This PR covers removal of 3rd party urllib dependency PythonCharmers that triggered s360 security bugs. We ll' be using python2 standard libs such as urllib and urlparse APIs moving forward.
Unit tested by mocking extension calls with distros running python v3.8.10, python 2.7.5.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "VMEncryption/transitions"]
 	path = VMEncryption/transitions
 	url = https://github.com/tyarkoni/transitions.git
-[submodule "OmsAgent/ext/python-future"]
-	path = OmsAgent/ext/python-future
-	url = https://github.com/PythonCharmers/python-future
 [submodule "AzureMonitorAgent/ext/python-future"]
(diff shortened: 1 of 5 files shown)


simathih:
is the usage impact only in the extension code for OMS? how will we test impact across various python versions? could this affect our install/enable telemetry?

simathih (changes requested):
how will we test impact across various python versions?

NarineM:
@simathih it was tested with 2 major versions of python (See the PR descriptions). We will need to run E2E tests across distro matrix in Canary as this change impacts extension code which is hard to test via mock testing. Yes, it can potentially impact SLA not only telemetry, as it is used for example to reitrieve resource info from metadata API: https://github.com/Azure/azure-linux-extensions/blob/master/OmsAgent/omsagent.py#L570
It is large chunk of change and needs to be tested E2E.

simathih (approved):

jscalev (approved):

PR #1766, opened 2023-07-20
Updates to Linux Dual pass repo for SBOM compliance

Jyotsna-Anand:
- Deleted OpenCensusAgent folder and go.sum, go.mod files to meet SBOM compliance
- These files are not used by Azure Disk Encryption and are not part of the "ade-singlepass-dev" branch
- The files in dual pass dev branch are being removed as OpenCensusAgent triggers CG alerts for a lot of untrusted packages being used.


canfikret (approved):
I don't think any team uses it. Teams should use OpenTelemetry instead, which is the next evolution of it.

PR #1768, opened 2023-07-27
VMAccess support for restoring SSH config

mayankdaruka-msft:
Currently, when a user runs the reset_ssh action with VMAccess, it copies a stock resource file from the internal resources directory and replaces sshd_config with that file. The issue, for some of our customers, with this is that their VMs are left in a state that differs from their original configuration.

These changes fix the above issue and make related improvements by:
- Altering the backup logic to create a new backup only if ChallengeResponseAuthentication or PasswordAuthentication have been modified, or if no backup exists
- Create a new backup in /var/cache/vmaccess, instead of the internal resources directory
- Having a new parameter, restore_backup_ssh, in addition to reset_ssh
- If restore_backup_ssh == True, restore the backup in /var/cache/vmaccess to sshd_config, allowing our customers to preserve their original configuration that complies with their security requirements, rather than using the VMAccess default config files

Work Item: https://msazure.visualstudio.com/One/_sprints/taskboard/CPlat%20Runtime%20Extensions/One/Gallium/CY23Q3/2Wk/2Wk09%20(Jul%2016%20-%20Jul%2029)?workitem=13384139

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -388,7 +388,7 @@ def _backup_and_update_sshd_config(hutil, attr_name, attr_value):
 
     hutil.log(f"Setting {attr_name} to {attr_value} in sshd_config.")
     
-    _backup_sshd_config()
+    _backup_sshd_config(hutil)
     _set_sshd_config(config, attr_name, attr_value)
     ext_utils.replace_file_with_contents_atomic(SshdConfigPath, "\n".join(config))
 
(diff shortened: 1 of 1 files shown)


VMAccess/vmaccess.py
```diff
@@ -182,23 +182,12 @@ def enable():
...
     name = "ChallengeResponseAuthentication"
-    config = ext_utils.get_file_contents(SshdConfigPath).split("\n")
-    for i in range(0, len(config)):
-        if config[i].startswith(name) and "no" in config[i].lower():
-            ext_utils.add_extension_event(name=hutil.get_name(), op="sshd", is_success=True,
-                                          message="ChallengeResponseAuthentication no")
-            return
-
-    ext_utils.add_extension_event(name=hutil.get_name(), op="sshd", is_success=True,
-                                  message="ChallengeResponseAuthentication yes")
```
jscalev, line 193: Why are the events no longer being raised?

jscalev (dismissed):

jscalev (approved):

PR #1770, opened 2023-07-31
Enable Syslog TCP port for 3P distros

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -1224,6 +1224,29 @@ def generate_localsyslog_configs():
         if features is not None and "useSyslogTcp" in features:            
             useSyslogTcp = features.get("useSyslogTcp")    
     
+    # always use syslog tcp port, unless 
+    # - the distro is Red Hat based and doesn't have semanage
+    #   these distros seem to have SELinux on by default and we shouldn't be installing semanage ourselves
+    distro, version = find_vm_distro('Enable')
+    for name in ['ubuntu', 'debian']:
```
NarineM, line 1231: Instead of being dependent on distro type is there a way to check se system setting instead? We will also have cases when it is one of the distroes where selinux is not enabled by default but it is enabled by customer and where the port will fail to be enabled.

AzureMonitorAgent/agent.py
```diff
@@ -1224,6 +1224,29 @@ def generate_localsyslog_configs():
...
+    for name in ['suse', 'sles', 'opensuse']:
+        if distro.startswith(name):
+            useSyslogTcp = True   
+          
+    for name in ['centos', 'redhat', 'red hat', 'rhel', 'oracle', 'ol', 'cbl-mariner', 'mariner', 'rocky', 'alma', 'amzn']:
+        if distro.startswith(name):
+            path = shutil.which("semanage") 
+
+            if path is None:
+                hutil_log_info("semanage not found, cannot let TCP Port through for syslog")
```
NarineM, line 1244: I think it would be helpful to tell also that by default syslog events will not flow in the case the tool is not present at the time of ama install, unless we implement fall back to socket path if the tool is not found or ask the customer to manually install it , then try the ama again. Can you add this logic in extension enable also? In that case we will not ask customers to uninstall/-reinstall AMA but only re run enable again after the tool is installed? Either way we need better message here to capture the impact on cx.

NarineM (changes requested):

NarineM (changes requested):

AzureMonitorAgent/agent.py
NarineM: If the logic is in enable() api, I think waagent subsequent calls will make the agent auto recover as long as cx installs or has pre-req tools.

NarineM (approved):

PR #1771, opened 2023-08-07
WaagentBugFix

arisettisanjana:
Need:
The method RunGetOutput is a wrapper for subprocess.check_output which is used for running a command and returning the CompletedProcess object.
Irrespective of the type of the output (CompletedProcess object) being returned, we are trying to decode it.
If the output is of type string, the decode step results in following error - AttributeError: 'std_output' object has no attribute 'decode'
Change:
Added a check to ensure that the output is decoded only if it is of type bytes.


ankitha-p (approved):

VMBackup/main/WaagentLib.py
```diff
@@ -2255,13 +2255,23 @@ def RunGetOutput(cmd, chk_err=True, log_cmd=True):
...
+            output = output.decode('latin-1')
     except subprocess.CalledProcessError as e:
         if chk_err and log_cmd:
             Error('CalledProcessError.  Error Code is ' + str(e.returncode))
             Error('CalledProcessError.  Command string was ' + e.cmd)
-            Error('CalledProcessError.  Command result was ' + (e.output[:-1]).decode('latin-1'))
-        return e.returncode, e.output.decode('latin-1')
-    return 0, output.decode('latin-1')
+            if isinstance(e.output[:-1], bytes):
+                Error('CalledProcessError.  Command result was ' + (e.output[:-1]).decode('latin-1'))
```
ankitha-p, line 2265: my earlier question was, if we check in the try block itself, do we even need to check twice in the except block.

iamashwani-who (approved):

VMBackup/main/WaagentLib.py
arisettisanjana: Ya, like if there is an exception raised even before checking for this condition then we would be needing them.

PR #1773, opened 2023-08-16
Replace python-future with standard python in AMA and LAD

NarineM:
This PR covers removal of 3rd party urllib dependency PythonCharmers that triggered s360 security bugs. With this change we will switch to standard python.
Unit tested by mocking extension calls with distros running python v3.8.10, python 2.7.5.


simathih (changes requested):
how will we test impact across various python versions?

simathih:
it has potential of breaking install/enable. we need to be elaborate with python version testing

NarineM:
Added python version info that was used to test the change to this PR description. Needs to be well tested in Canary E2E as we can not rely only on testing done so far with mocking extension calls.

simathih (approved):

simathih:
Ensure E2E pipeline and Canary E2E pass

D1v38om83r (dismissed):

D1v38om83r (approved):

simathih (approved):

PR #1774, opened 2023-08-16
Set password auth in 50-cloud-init.conf for CentOS and RHEL systems

ani-sinha:
Azure 'reset password' feature overrides the /etc/ssh/sshd_config file and sets the PasswordAuthentication to 'yes'. Unfortunately, cloud-init versions 22.3 and newer sets 'PasswordAuthentication' to 'no' in the /etc/ssh/sshd_config.d/50-cloud-init.conf which overrides the setting in /etc/ssh/sshd_config file. VMAccess handles this for Ubuntu by setting PasswordAuthentication to 'yes' additionally in /etc/ssh/sshd_config.d/50-cloud-init.conf file. This change extends the same method to include CentOS and RHEL systems.

Fixes: https://github.com/canonical/cloud-init/issues/4335

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -382,13 +382,12 @@ def _allow_password_auth():
     _set_sshd_config(config, "PasswordAuthentication", "yes")
     ext_utils.replace_file_with_contents_atomic(SshdConfigPath, "\n".join(config))
 
-    if isinstance(MyDistro, dist_utils.UbuntuDistro): #handle ubuntu 22.04 (sshd_config.d directory)
-        cloudInitConfigPath = "/etc/ssh/sshd_config.d/50-cloud-init.conf"
-        config = ext_utils.get_file_contents(cloudInitConfigPath)
-        if config is not None: #other versions of ubuntu don't contain this file
-            config = config.split("\n")
-            _set_sshd_config(config, "PasswordAuthentication", "yes")
-            ext_utils.replace_file_with_contents_atomic(cloudInitConfigPath, "\n".join(config))
+    cloudInitConfigPath = "/etc/ssh/sshd_config.d/50-cloud-init.conf"
+    config = ext_utils.get_file_contents(cloudInitConfigPath)
+    if config is not None:
+        config = config.split("\n")
+        _set_sshd_config(config, "PasswordAuthentication", "yes")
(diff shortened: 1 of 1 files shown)


VMAccess/vmaccess.py
```diff
@@ -382,10 +382,12 @@ def _allow_password_auth():
     _set_sshd_config(config, "PasswordAuthentication", "yes")
     ext_utils.replace_file_with_contents_atomic(SshdConfigPath, "\n".join(config))
 
-    if isinstance(MyDistro, dist_utils.UbuntuDistro): #handle ubuntu 22.04 (sshd_config.d directory)
+    if isinstance(MyDistro, dist_utils.UbuntuDistro) or \ #handle ubuntu 22.04 (sshd_config.d directory)
```
cjp256, line 385: I think it would be best if we just removed the distro gates so we don't have to add them all over time.

If 50-cloud-init.conf exists, I cannot think of a case where we don't want to update it.  That is, unless vmaccess switches to using its own high-priority config file (e.g. 10-azure-vmaccess.conf) for distros using multi-part configs.

cjp256 (changes requested):

VMAccess/vmaccess.py
anhvoms, line 385: I would suggest doing a sshd -T and verifies the output to check that Password authentication is indeed enabled after the restart of ssh. If the password authentication is still set to no, an error should be returned back to the user. 
This is the only way to ensure that the change did take effect as intended as there might be other .conf files that affect sshd behavior, not just 50-cloud-init.conf.

ani-sinha:
I do not know how to do this. I do t know enough of this tool. Can one of
you please help fix this for CentOS a s RHEL systems? Thanks.

VMAccess/vmaccess.py
ani-sinha: Fixed this.

ani-sinha: I do not know how to do this.

ani-sinha:
One more Ping ...

anhvoms:
@nkuchta can you help review this PR? Thanks

jscalev (approved):

ani-sinha:
Please merge!

ani-sinha:
Ping once more ...

PR #1781, opened 2023-08-23
Don't send TTL header as part of guest based snapshot call

arisettisanjana:
Need:The TTL flag was previously added as part of the headers when taking a snapshot via Guest. However, the XStore service only accepts TTL as part of the header when it receives Sas Uri's encoded using system keys. Currently it can be achieved through Backup Host Server only. Hence, the snapshotting flow initiated by guest agent should not try to specify TTL.

Change:
Modifying the guestsnapshotter.py to removing ttl as part of header even when TTL lag is enabled.


VMBackup/main/guestsnapshotter.py
```diff
@@ -95,7 +95,8 @@ def snapshot(self, sasuri, sasuri_index, settings, meta_data, snapshot_result_er
                         value = meta['Value']
                         headers["x-ms-meta-" + key] = value
                 if(CommonVariables.isSnapshotTtlEnabled in settings and settings[CommonVariables.isSnapshotTtlEnabled]):
-                    headers[CommonVariables.snapshotTtlHeader] = '0'
+                    self.logger.log("Not passing the TTL header via Guest path though it is enabled")
+                    #headers[CommonVariables.snapshotTtlHeader] = '0'
```
kirankumarp-azure, line 99: In general, let's not have commented code in our codebase. The only times I've found commented code to be useful is when it somehow helps local machine setup/testing.

For any other context, the code should be either there or not. Not in between.

kirankumarp-azure:
PR Title "Don't send TTL header as part of guest based snapshot call" would explain what we're doing better, IMO.

kirankumarp-azure (approved):

iamashwani-who (approved):
Looks good to me.

PR #1792, opened 2023-09-01
In GCS scenarios, ensure that AMACoreAgent is running

AaronRM:
This reverts commit 065757b47b69f4116e376c10f66693f8f483b0f5.


simathih (approved):
why was it reverted and being reapplied? I am slightly confused now

AaronRM:
> why was it reverted and being reapplied? I am slightly confused now

Discussed offline on Friday. I was going for a more targeted way of starting the azuremonitor-coreagent service, but ultimately decided it was safer to enable it for all AMA-1P scenarios at this point (once we got the AMACA log verbosity issue addressed in the latest build).

PR #1800, opened 2023-09-19
Sarisetti/bug fix enum

arisettisanjana:
The previous release uses enum module to map the log levels.
Need:
As using enum was not compatible with older python versions.
Changes include:
moving LogSeverity.json file to inside main folder
and also modified the enum specific logic to ensure it is compatible with both versions.


VMBackup/main/Utils/EventLoggerUtil.py
```diff
@@ -79,7 +79,14 @@ def update_properties(self, task_id):
 
     def severity(self, severity_level):
         level = 0
```
iamashwani-who, line 81: default level should be Warning, right?

iamashwani-who (approved):

ankitha-p (approved):

PR #1801, opened 2023-09-19
Sarisetti/bug fix waagent

arisettisanjana:
Need:
The extension code employs the "is not" operator for comparisons within an if block, which is triggering a warning.

Change:
have updated it to use the "!=" operator, which enhances readability and aligns with common coding practices.


ankitha-p (approved):

iamashwani-who:
Please publish test results on env with Python 2.7 and Python 3+.

iamashwani-who (approved):

cjp256:
$ cat x.py
foo=None

if foo != None:
    print("foo not None")


$ pylint x.py
************* Module x
<snip>
x.py:3:3: C0121: Comparison 'foo != None' should be 'foo is not None' (singleton-comparison)

arisettisanjana:
In pythin as True, False and None are considered as literals and still the good practice is to have is operator for comparision.
Discarding the PR.

PR #1809, opened 2023-09-29
Tests if execution of semanage is required. 

lmalmeida:
Always executing _semanage port -a_ it creates additonal audit entries and load on Microsoft Defender.
Doing a test if is required allows reducing system load

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -1241,8 +1241,12 @@ def generate_localsyslog_configs():
             if check_semanage != 0:            
                 hutil_log_info("semanage not found, cannot let TCP Port through for syslog")
             elif syslog_port != '':
-                # allow the syslog port in SELinux
-                run_command_and_log('semanage port -a -t syslogd_port_t -p tcp ' + syslog_port)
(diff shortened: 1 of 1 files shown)


simathih:
is this really needed? semanage doesn't allow you to add duplicates in the first place: 
[~]# semanage port -a -t syslogd_port_t -p tcp 28330
ValueError: Port tcp/28330 already defined

simathih:
it doesn't create additional entries

simathih (approved):

lmalmeida:
> is this really needed? semanage doesn't allow you to add duplicates in the first place: [~]# semanage port -a -t syslogd_port_t -p tcp 28330 ValueError: Port tcp/28330 already defined

It shouldn't be required, but I'm seeing additional audit log entries on RHEL 8.8 and impact on the performance of Microsoft Defender.

simathih:
> > is this really needed? semanage doesn't allow you to add duplicates in the first place: [~]# semanage port -a -t syslogd_port_t -p tcp 28330 ValueError: Port tcp/28330 already defined
> 
> It shouldn't be required, but I'm seeing additional audit log entries on RHEL 8.8 and impact on the performance of Microsoft Defender.

got it, this should be included in AMA Linux 1.29

PR #1818, opened 2023-10-15
set rsyslog worker threads and syslog-ng flow control

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -1291,7 +1301,13 @@ def generate_localsyslog_configs():
                 if portSetting not in f.read():
                     portUpdated = True
 
-            if portUpdated == True:
+            flowControl = 'flow-control'
+            setFlowControl = False
+            with open('/etc/syslog-ng/conf.d/azuremonitoragent-tcp.conf') as f:
+                if flowControl not in f.read():
```
NarineM, line 1307: why we check the content of config file for some properties?

simathih: we won't update the existing config on the cx machine otherwise, these are new settings which require a config deployment

NarineM: ok, is not that the issue that we need to address with upgrade bug? Uninstall already removes these configs but can you confirm? we have a work item to fix AMA upgrade to make sure configs are overwritten with what comes in ama install package. I think we should not add custom handling for some settings only till the upgrade bug is addressed.

PR #1820, opened 2023-10-16
Add ASAP SKU support.

vimish:


VMEncryption/main/DiskUtil.py
```diff
@@ -887,10 +890,15 @@ def get_azure_data_disk_controller_and_lun_numbers(self, dev_items_real_paths):
         """
         Return the controller ids and lun numbers for data disks that show up in the dev_items
         """
+        list_devices = []
+        is_os_nvme, root_device_path_nvme = self.is_os_disk_nvme()
 
+        if is_os_nvme:
```
pankajosh, line 896: Is it true to say that if OS disk is NVMe then other all disks must be NVMe.

vimish: Yes. For the new ASAP SKU's both OS and Data disk will be NVMe. Resource and BEK disk are SCSI.

VMEncryption/main/EncryptionSettingsUtil.py
```diff
@@ -247,10 +247,16 @@ def get_settings_data(self, protector_name, kv_url, kv_id, kek_url, kek_kv_id, k
...
+        is_os_nvme, root_device_path_nvme = disk_util.is_os_disk_nvme()
         if os.path.exists(root_device_path):
             root_device_path = os.path.realpath(root_device_path)
+        elif os.path.exists(root_device_path_scsi):
+            root_device_path = os.path.realpath(root_device_path_scsi)
+        elif is_os_nvme and os.path.exists(root_device_path_nvme):
+            root_device_path = os.path.realpath(root_device_path_nvme)
         else:
-            root_device_path = os.path.realpath(os.path.join(CommonVariables.azure_symlinks_dir, "scsi0/lun0"))
+            self.logger.log("Cannot locate root device")
```
bamuir, line 259: If the root device is not found, and we are no longer defaulting to "root_device_path = os.path.realpath(os.path.join(CommonVariables.azure_symlinks_dir, "scsi0/lun0"))", how does the error come across on line 260 if root_device_path is None? 

I can't think of how to test this right now, but it seems there might be a TypeError when root device is not found. Wondering if we want to throw a more specific error ?

bamuir (approved):
Looks good to me. Would we be able to measure adoption / usage currently using the VM size on our existing tables?

VMEncryption/main/DiskUtil.py
```diff
@@ -1203,6 +1291,18 @@ def get_azure_devices(self):
...
+    
+    def get_azure_nvme_os_devices(self):
+        devices = []
+        is_os_nvme, os_block_device = self.is_os_disk_nvme()
+        if is_os_nvme:
+            proc_comm = ProcessCommunicator()
+            self.command_executor.ExecuteInBash('ls {0}*'.format(os_block_device), True, communicator=proc_comm)
+            os_devices = re.split('[ \n]', proc_comm.stdout)
+            for os_device in os_devices:
+                if os_device.startswith(os_block_device) and os.path.exists(os_device):
```
Jyotsna-Anand, line 1303: Is there a case when os_device can be null? Should there be a check to ensure os_device is not null, before invoking these other methods on os_device?

Jyotsna-Anand (approved):

VMEncryption/main/EncryptionSettingsUtil.py
vimish: The default to scsi0/lun0 was breaking in NVMe VM's as it did not exist. An error was thrown when trying to stamp the disks. I believe the same error will be thrown if we are not able to locate the root device.

But, I get the point. I will look into a better error handling probably in the next PR.

VMEncryption/main/DiskUtil.py
vimish: If OS is determined to be NVMe then we will always have an OS device.
Also, I am planning to do a review of error handling for this new path. I will add more checks then.

PR #1822, opened 2023-10-18
Fixing missing log statements in guest agent log files by forcing dispose before process exit

arisettisanjana:
Changes:
1) Calling dispose method to push the remaining logs in the event Queue, before we exit.
2) Included the entire code of  deamon method inside try catch block, to get logs before we abruptly terminate.
3) Also cleaning up of the temp folder, created to store the events initially.


VMBackup/main/Utils/EventLoggerUtil.py
```diff
@@ -122,8 +122,8 @@ def log_event(self, message):
         try:
             if self.current_message_len + len(message) > LoggingConstants.MaxMessageLengthPerEvent:
                 self.event_queue.put(Event("Info",
-                                           self.current_message, LoggingConstants.DefaultEventTaskName,
-                                           self.operation_id, self.extension_version).convertToDictionary())
+                                        self.current_message, LoggingConstants.DefaultEventTaskName,
```
iamashwani-who, line 125: no change. We can revert the lines 125 and 126.

VMBackup/main/Utils/EventLoggerUtil.py
```diff
@@ -268,6 +268,9 @@ def _dispose(self, disposing):
                             logger.log("Warning: Unable to process events before termination of extension. Exception: {0}" .format(str(ex)))
                 self.disposed = True
                 print("Information: Event Logger has terminated")
+                print("Clearing the temp directory")
+                if os.path.exists(self.temporary_directory):
+                    os.rmdir(self.temporary_directory)
```
iamashwani-who, line 273: I think this can be moved to a helper method with a try-catch to suppress failure?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -253,6 +255,8 @@ def do_parse_context(self, operation, seqNo):
             self.log("waagent new path is used")
         if not _context:
             self.log("maybe no new settings file found")
+            if(self.eventlogger is not None):
+                self.eventlogger.dispose()
```
iamashwani-who, line 259: this can be moved to a helper method with a try-catch to suppress failure?

iamashwani-who (approved):

VMBackup/main/Utils/HandlerUtil.py
arisettisanjana: discussed offline.

VMBackup/main/Utils/EventLoggerUtil.py
```diff
@@ -245,6 +245,13 @@ def _send_event_file_to_event_directory(self, file_path, events_folder):
             FileHelpers.deleteFile(file_path)
             print("Information: Event reporting has paused due to reaching maximum capacity in the Event directory. Reporting will resume once space is available. Events for this iteration will not be reported.")
 
+    def clear_temp_directory(self, directory_path):
+        try:
+            if os.path.exists(directory_path):
+                os.rmdir(directory_path)
```
rakrashba, line 251: [os.rmdir](https://docs.python.org/3/library/os.html#os.rmdir) only works if the folder does not have any subfolders or files inside it - basically when it's empty. It will throw an exception. Use it if and only if you are sure that the directory is empty. Otherwise please look at something like [shutil.rmtree](https://docs.python.org/3/library/shutil.html#shutil.rmtree).

rakrashba (changes requested):

VMBackup/main/Utils/EventLoggerUtil.py
arisettisanjana: So even the intention behind using os.rmdir was to clear an empty directory as this method would be called only at the end where all the files from temp directory would be moved to event folder and then we would be clearing out the temp folder. 
But yes it would be better to have that check, will update it.

rakrashba (approved):

iamashwani-who (approved):

PR #1824, opened 2023-10-19
[enhancement] Diagnostic script to monitor system level data during a restore point operation

rakrashba:
During snapshot operations on Linux we often come across bugs that are pretty tricky to debug without any knowledge about the system state at that point in time. This PR will introduce a small app written in go that will do exactly that.

The data this app will capture can be considered customer data - especially when strace is on - hence this app will not be shipped with normal extensions code.

The README has extensive directions on how to build and run this both manually and as part of a Restore Point operation.


.gitignore
```diff
@@ -97,6 +97,8 @@ VMBackup/MANIFEST
 VMBackup/*.pyproj.user
 VMBackup/*.suo
 VMBackup/main/safefreeze/bin/*
+# VMBackup debughelper builds
+VMBackup/debughelper/msft_snap_monit
```
iamashwani-who, line 101: is this the output path for binary compilation?

VMBackup/main/handle.py
```diff
@@ -582,7 +621,8 @@ def daemon():
         backup_logger.commit_to_local()
         if(eventlogger is not None): 
             eventlogger.dispose()
-
+    if monitor_process is not None:
+        monitor_process.terminate()
```
iamashwani-who, line 625: line 588 uses os.kill while we use process.terminate here. Why is the difference?

.gitignore
rakrashba: Yes. The binary file is by default created in the working directory itself. The name of the binary is defined by the module name which can be found in the first line of `VMBackup/debughelper/go.mod`

iamashwani-who (dismissed):

vityagi (approved):
Approved

iamashwani-who (approved):

ankitha-p (approved):

rakrashba:
Merging.

PR #1832, opened 2023-10-26
VMEncryption: For confidential VM temp disk enc, do not attempt secure key release if KEK is not provided

canfikret:
This is to unblock tdx CVM until GA is fixed.

### DIFF
--- a/VMEncryption/main/handle.py
+++ b/VMEncryption/main/handle.py
@@ -2116,7 +2116,7 @@ def daemon_encrypt():
 
         if not confidential_encryption_datadisk_flag:
             msg = "Currently for this CVM, Data disk encryption is not supported. {0}".format(msg)
-            logger.log(msg=msg,level=CommonVariables.ErrorLevel)
+            logger.log(msg=msg,level=CommonVariables.WarningLevel)
             return
         hutil.do_status_report(operation="EnableEncryption",
                        status=CommonVariables.extension_success_status,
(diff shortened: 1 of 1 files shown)


VMEncryption/main/handle.py
```diff
@@ -797,7 +797,9 @@ def enable():
         encryption_operation = public_settings.get(CommonVariables.EncryptionEncryptionOperationKey)
         check_release_rsa_or_ec_key = False
         if security_Type == CommonVariables.ConfidentialVM:
-            check_release_rsa_or_ec_key = True
+            if CommonVariables.KeyEncryptionKeyURLKey in public_settings:
```
pankajosh, line 800: are we saying KEK is not required if temp disk encryption is attempted, whereas if data disk encryption is attempted then KEK is required and SKR must be attempted.

canfikret: Correct.

pankajosh: currently I do not see the precheck for 
ConfidentialEncryptionDataDisk is true then KEK should be present, 
or we can include this check this condition for SKR. 
if ConfidentialEncryptionDataDisk and KEK 
    do SKR

canfikret: We can add that check for ConfidentialEncryptionDataDisk true scenario. However, note that this flag will go away when data disk enc goes public preview.

pankajosh: yes, I agree with your comments.

pankajosh (approved):

PR #1835, opened 2023-11-02
do not SKR in CVM for temp disk encryption if KEK is not present.

pankajosh:
for CVM, do not SKR for temp disk encryption if KEK is not present.


pankajosh:
wrong branch name created.

PR #1837, opened 2023-11-02
do not SKR in CVM for temp disk encryption if KEK is not present.

pankajosh:


VMEncryption/ReleaseNotes.txt
```diff
@@ -1,3 +1,11 @@
+1.4.0.5
+- For CVM temp disk encryption, do not attempt secure key release if KEK is not provided.
+
+1.4.0.4
+- Add support for boost SKU's
```
canfikret, line 5: let's clarify this is "Azure boost SKUs"

pankajosh: I reference it from 
https://github.com/Azure/azure-linux-extensions/blob/8ee39d86f5b41819390403a7c33e7236f0033fe8/VMEncryption/ReleaseNotes.txt
this commit is used for release 1.4.0.4

PR #1840, opened 2023-11-06
Ensure 1P syslog config is always present in case of 1P/3P combined mode

simathih:
when we added syslog watcher thread in 1.26 (https://github.com/Azure/azure-linux-extensions/commit/d6a32c812eb4edd25d29590d44881dc5fbfb583e), if there is mixed mode in next generation config and customer has all the following:

1. genevaConfiguration is enabled in the next generation schema
2. azureMonitorConfiguration is enabled in the next generation schema
3. 3P DCR has no syslog configured and hence no marker file is generated

then, in enable(), we first deploy syslog config but they are wiped in the watcher thread because syslog is not enabled as per 3P DCR even though 1P mode needs it


NarineM (dismissed):

AzureMonitorAgent/agent.py
```diff
@@ -2132,4 +2170,4 @@ def get_error_message(self, operation):
                                                                    self.error_code)
 
 if __name__ == '__main__' :
-    main()
+    main()
```
NarineM, line 2173: Just noticed this after signing off. Looks accidental change (line #2173)

simathih: fixed this

NarineM (approved):

PR #1843, opened 2023-11-08
Fix Telegraf not starting errors

aswatt:
If configuration had no metrics or perf counters set up to be collected, telegraf would (rightly) not be installed, but LAD's code would repeatedly try to start up telegraf every 30 seconds (and log the failure to extension.log). This fixes this bug to check if telegraf should be enabled based on the configuration.

Specifically, if there is a metrics section but no `scheduledTransferPeriod` value (and no perf counters section), our previous behavior was to try and set up metrics for PT1H and start telegraf, which wouldn't succeed as telegraf was never installed initially. This change lines us up with [our current public docs](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/diagnostics-linux?toc=%2Fazure%2Fazure-monitor%2Ftoc.json&tabs=azcli#metrics), which clarify that at least one `scheduledTransferPeriod` should be specified.


AaronRM (approved):
:shipit:

simathih (approved):

PR #1851, opened 2023-11-21
Adding new error code FailedSafeFreezeBinaryNotFound

arisettisanjana:
Requirement:
Currently, in the VM Snapshot Linux extension, its run can fail with generic fsfreezefailedtimeout error code in cases when the requisite fsfreeze binary file is not present in the bin/ folder. This scenario can occur when the target file has been manually deleted or deleted by an AntiVirus program that does not recognize the file. We require a different error code to distinguish such failures to enable better debugging when file is missing. Introducing FailedSafeFreezeBinaryNotFound error code to cover for such scenarios.

Changes made in:
fsfreeze.py - before IO freeze is performed added a check whether the binary exists and then proceed accordingly.
handle.py - checks the flag and exits if file is missing
added error code in common.py, ExtensionErrorCodeHelper.py


VMBackup/main/handle.py
```diff
@@ -277,6 +277,8 @@ def daemon():
             backup_logger.log(errMsg, True, 'Error')
 
         freezer = FsFreezer(patching= MyPatching, logger = backup_logger, hutil = hutil)
+        backup_logger.log(freezer.file_exists, True, 'Info')
```
iamashwani-who, line 280: I think it might be better to not wait till daemon process to check file existence. This can be added as a basic validation just like we do for python install?

VMBackup/main/fsfreezer.py
```diff
@@ -128,7 +129,12 @@ def __init__(self, patching, logger, hutil):
         else:
             self.logger.log("isArm64Machine : " + str(self.isArm64Machine) + " Using x64 safefreeze binary")
             self.safeFreezeFolderPath = "safefreeze/bin/safefreeze"
-
+        
+        temp = os.path.join(os.path.dirname(os.path.realpath(__file__)), self.safeFreezeFolderPath)
```
iamashwani-who, line 133: probably better to expose it as a helper method that can be called from other places to validate if file exists.

VMBackup/main/common.py
```diff
@@ -95,6 +95,7 @@ class CommonVariables:
     error_http_failure = 15
     FailedHandlerGuestAgentCertificateNotFound = 16
     #error_upload_status_blob = 16
+    FailedSafeFreezeBinaryNotFound = 18
```
ankitha-p, line 98: this code is reserved for FailedHandlerSnapshotBitlocker in windows. Please re-check and assign a non-conflicting error number. Better to be in the range [121-130]

arisettisanjana: yeah have checked it, modified to 19 will push it along with the other changes in new iteration.
Would also require modification at the CRP side to include this error cde.

VMBackup/main/handle.py
arisettisanjana: As discussed offline adding it in a way similar to python validation would not be throwing an error, which would not be that useful in this case as it would be doing the validation in handle.sh file. So, going ahead with this.

VMBackup/main/fsfreezer.py
```diff
@@ -147,7 +150,14 @@ def __init__(self, patching, logger, hutil):
         self.getLockRetry = 0
         self.maxGetLockRetry = 5
         self.safeFreezelockFile = None
-
+    
+    def isFileExists(self, relative_path):
```
iamashwani-who, line 154: CheckIfFileExists or DoesFileExists

VMBackup/main/fsfreezer.py
```diff
@@ -147,7 +150,14 @@ def __init__(self, patching, logger, hutil):
...
+        full_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), relative_path)
+        self.logger.log("path "+ str(full_path))
+        self.logger.log("path exists " + str(os.path.exists(full_path)))
```
iamashwani-who, line 157: we can introduce a local variable to keep the value instead of calling os.path.exists multiple times.

self.file_exists = os.path.exists(full_path)
self.logger.log("path " + str(full_path) + ". Found the file: " + fileExists)

VMBackup/main/handle.py
```diff
@@ -342,6 +344,17 @@ def daemon():
                 temp_result=CommonVariables.FailedHandlerGuestAgentCertificateNotFound
                 temp_status= 'error'
                 exit_with_commit_log(temp_status, temp_result,error_msg, para_parser)
+            
+            if(freezer.file_exists == False):
+                backup_logger.log("inside if",True, "Info")
```
iamashwani-who, line 349: why is this log statement required?

VMBackup/main/handle.py
```diff
@@ -342,6 +344,17 @@ def daemon():
...
+                file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), freezer.safeFreezeFolderPath) 
+                error_msg = "safefreeze binary is missing in the following path " + str(file_path)
+                backup_logger.log("setting the error code", True, "Info")
```
iamashwani-who, line 352: same here. why is this log statement required if the final error code is logged elsewhere?

iamashwani-who (approved):

VMBackup/main/handle.py
ankitha-p, line 280: Please add some text as to what variable is being printed. For example {FreezeFileExists: true}. Currently only variable value is printed.

VMBackup/main/fsfreezer.py
```diff
@@ -128,7 +129,9 @@ def __init__(self, patching, logger, hutil):
         else:
             self.logger.log("isArm64Machine : " + str(self.isArm64Machine) + " Using x64 safefreeze binary")
             self.safeFreezeFolderPath = "safefreeze/bin/safefreeze"
-
+
```
ankitha-p, line 132: Please add a log mentioning that the presence of safeFreezeFolder path will be checked.

VMBackup/main/fsfreezer.py
```diff
@@ -111,7 +111,8 @@ def __init__(self, patching, logger, hutil):
         self.hutil = hutil
         self.safeFreezeFolderPath = "safefreeze/bin/safefreeze"
         self.isArm64Machine = False
-
+        self.file_exists = True
```
ankitha-p, line 114: can we re-name variable to specifically what file_exists? safeFreezeBinExists? or instead you could add a comment mentioning what file we are referring to

iamashwani-who (approved):

PR #1866, opened 2024-01-03
Changes needed for snapshots with the new Madari based path using xDiskSvc

rakrashba:
## Things covered by this PR

- start a long running process seamlessly without affecting existing VM Pipeline based code path
- automatic creation of a systemd based service with fallback to a normal linux process if systemd is not available
- POC communication with xDiskSvc to take a snapshot via the Crash Consistent Path available on Azure portal today
- Logical code path separation so that dev can progress seamlessly without affecting the other path

## Things not covered in this PR:

- logging when safefreeze is active
- Billing info logic


VMBackup/main/handle.py
```diff
@@ -668,6 +668,10 @@ def enable():
     try:
         hutil.do_parse_context('Enable', configSeqNo)
         backup_logger.log('starting enable', True)
+        try:
+            create_host_based_service()
```
iamashwani-who, line 672: can we add a check that long running extension setting should be enabled before we invoke the host based service. Otherwise we can skip the steps.

VMBackup/main/handle_host_daemon.py
```diff
@@ -0,0 +1,861 @@
...
+if IS_PYTHON3:
+    import urllib.parse as urllib
+else:
+    import urllib
+
+
+SCRIPT_DIR=os.path.dirname(os.path.realpath(__file__))
+BASE_URI="http://168.63.129.16"
+STORAGE_DEVICE_PATH = '/sys/bus/vmbus/devices/'
+GEN2_DEVICE_ID = 'f8b3781a-1e82-4818-a1c3-63d806ec15bb'
```
iamashwani-who, line 42: any reason for a hard coded device id?

VMBackup/main/handle.py
rakrashba: This would be a CRP passed setting I am assuming?

VMBackup/main/handle_host_daemon.py
rakrashba: Yes. The resource disk seems to have a fixed GUID when attached to the VM. The old implementation uses this as well [here](https://github.com/Azure/azure-linux-extensions/blob/e6dc7a79bf29f317beeafec29206666f0cec7f0e/VMBackup/main/Utils/ResourceDiskUtil.py#L13C1-L13C56).

ankitha-p (approved):

PR #1872, opened 2024-01-29
Add cpu quotas for AMA

hestolz:
set based on benchmarks and real-world guest agent telemetry

### DIFF
--- a/AzureMonitorAgent/HandlerManifest.json
+++ b/AzureMonitorAgent/HandlerManifest.json
@@ -24,7 +24,7 @@
         },
         {
           "name": "azuremonitor-agentlauncher"
-          "cpuQuotaPercentage": 2
+          "cpuQuotaPercentage": 4
         },
         {
           "name": "azuremonitor-coreagent"
--- a/AzureMonitorAgent/HandlerManifest.json
+++ b/AzureMonitorAgent/HandlerManifest.json
@@ -17,21 +17,26 @@
       "services": [
         {
           "name": "azuremonitoragent"
+          "cpuQuotaPercentage": 250
         },
         {
           "name": "azuremonitoragentmgr"
         },
         {
           "name": "azuremonitor-agentlauncher"
+          "cpuQuotaPercentage": 2
         },
         {
           "name": "azuremonitor-coreagent"
+          "cpuQuotaPercentage": 200
         },
         {
           "name": "metrics-extension"
+          "cpuQuotaPercentage": 5
         },
         {
           "name": "metrics-sourcer"
+          "cpuQuotaPercentage": 10
         }
       ]
     }


AzureMonitorAgent/HandlerManifest.json
```diff
@@ -17,21 +17,26 @@
       "services": [
         {
           "name": "azuremonitoragent"
+          "cpuQuotaPercentage": 250
         },
         {
           "name": "azuremonitoragentmgr"
```
hestolz, line 23: azuremonitoragentmgr cgroup may have any number of mdsd tenants running under it in 1p case. as such, we cannot reasonably set a limit

simathih (approved):

hestolz:
> to clarify, 250 means upto 250% on a 4 core machine? curious because, 250% on 3 core vs 4 core would leave different amounts of room behind for other applications

250 means 2.5 cores. This setting is directly translated by guest agent into systemd `CPUQuota` attribute for the service cgroup https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#CPUQuota=

simathih:
> > to clarify, 250 means upto 250% on a 4 core machine? curious because, 250% on 3 core vs 4 core would leave different amounts of room behind for other applications
> 
> 250 means 2.5 cores. This setting is directly translated by guest agent into systemd `CPUQuota` attribute for the service cgroup https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#CPUQuota=

is the higher value safe in that case customer has fewer cores? I am assuming there is a way for customer to override this and we have documents on how they can do it

hestolz:
> > > to clarify, 250 means upto 250% on a 4 core machine? curious because, 250% on 3 core vs 4 core would leave different amounts of room behind for other applications
> > 
> > 
> > 250 means 2.5 cores. This setting is directly translated by guest agent into systemd `CPUQuota` attribute for the service cgroup https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#CPUQuota=
> 
> is the higher value safe in that case customer has fewer cores? I am assuming there is a way for customer to override this and we have documents on how they can do it

Currently we have no limits. So this is better than what we have currently. We can't make this variable on our side depending on how many cores the machine has etc; in order to not impact high EPS customers, we need to avoid setting a cap that will interfere with their workload. In case override is desired, it can be set as documented [here](https://github.com/Azure/azure-vmextension-publishing/wiki/2.0-Partner-Guide-Handler-Design-Details#2711-override-extension-cpu-memory-limits-in-linux)

simathih:
> > > > to clarify, 250 means upto 250% on a 4 core machine? curious because, 250% on 3 core vs 4 core would leave different amounts of room behind for other applications
> > > 
> > > 
> > > 250 means 2.5 cores. This setting is directly translated by guest agent into systemd `CPUQuota` attribute for the service cgroup https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#CPUQuota=
> > 
> > 
> > is the higher value safe in that case customer has fewer cores? I am assuming there is a way for customer to override this and we have documents on how they can do it
> 
> Currently we have no limits. So this is better than what we have currently. We can't make this variable on our side depending on how many cores the machine has etc; in order to not impact high EPS customers, we need to avoid setting a cap that will interfere with their workload. In case override is desired, it can be set as documented [here](https://github.com/Azure/azure-vmextension-publishing/wiki/2.0-Partner-Guide-Handler-Design-Details#2711-override-extension-cpu-memory-limits-in-linux)

thanks! makes sense

simathih (approved):

KidiIT (commented):
All reviewed and approved❤️🫶

KidiIT:
All checked and approved❤️🫶

PR #1882, opened 2024-02-26
Change from UDP to Unix Socket for telegraf => ME communication

simathih:
to handle occasional bursts in traffic that ME isn't able to handle
https://portal.microsofticm.com/imp/v3/incidents/incident/463024897/summary

### DIFF
--- a/AzureMonitorAgent/packaging.sh
+++ b/AzureMonitorAgent/packaging.sh
@@ -31,6 +31,7 @@ cp ../Common/WALinuxAgent-2.0.16/waagent .
 cp -r  ../LAD-AMA-Common/metrics_ext_utils .
 cp -r  ../LAD-AMA-Common/telegraf_utils .
 cp -r  ../Diagnostic/services .
+cp -f ../AzureMonitorAgent/services/metrics-extension.service services/metrics-extension.service
 
 # cleanup packages, ext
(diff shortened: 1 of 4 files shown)


AzureMonitorAgent/services/metrics-extension.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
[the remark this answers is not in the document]
simathih: it was using this service file which was shared with LAD, I separated it now https://github.com/Azure/azure-linux-extensions/blob/simathih/MESocket/Diagnostic/services/metrics-extension.service

AzureMonitorAgent/packaging.sh
```diff
@@ -31,6 +31,7 @@ cp ../Common/WALinuxAgent-2.0.16/waagent .
 cp -r  ../LAD-AMA-Common/metrics_ext_utils .
 cp -r  ../LAD-AMA-Common/telegraf_utils .
 cp -r  ../Diagnostic/services .
```
thperapp, line 33: Would it be necessary to remove the old ME service file from being copied?

thperapp: cp -f below should replace it, resolving

thperapp (approved):

PR #1883, opened 2024-02-26
Prevent AgentLauncher & Fluent-Bit restarts frequently during normal …

simathih:
…Linux AMA operation

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -542,7 +542,8 @@ def enable():
         start_syslogconfig_process()
     elif ensure.get("azuremonitoragentmgr") or is_gcs_single_tenant:
         # In GCS scenarios, ensure that AMACoreAgent is running
-        ensure["azuremonitor-coreagent"] = True
+        if platform.machine() != 'aarch64':
+            ensure["azuremonitor-coreagent"] = True
 
     hutil_log_info('Handler initiating onboarding.')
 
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -532,8 +532,11 @@ def enable():
         log_and_exit("Enable", GenericErrorCode, "Failed to add environment variables to {0}: {1}".format(config_file, e))
 
     if "ENABLE_MCS" in default_configs and default_configs["ENABLE_MCS"] == "true":
-        start_amacoreagent()
-        restart_launcher()
+        if platform.machine() != 'aarch64':
+            # enable processes for Custom Logs
+            ensure["azuremonitor-agentlauncher"] = True
+            ensure["azuremonitor-coreagent"] = True
+
```
hestolz, line 539: there's also line 545 (github not letting me comment there ...) another instance of `start_amacoreagent`, that should probably be changed as well to ensure

AzureMonitorAgent/agent.py
```diff
@@ -532,14 +532,17 @@ def enable():
...
+            ensure["azuremonitor-agentlauncher"] = True
+            ensure["azuremonitor-coreagent"] = True
+            
         # start the metrics and syslog watcher only in 3P mode
         start_metrics_process()
         start_syslogconfig_process()
     elif ensure.get("azuremonitoragentmgr") or is_gcs_single_tenant:
         # In GCS scenarios, ensure that AMACoreAgent is running
-        start_amacoreagent()
+        ensure["azuremonitor-coreagent"] = True
```
AaronRM, line 545: Need to add a check for aarch64 as in the ENABLE_MCS case.

AaronRM (changes requested):

AzureMonitorAgent/agent.py
hestolz: linter we have in packaging script would have found it 👍

AzureMonitorAgent/agent.py
simathih: added a check

hestolz (approved):

AaronRM (approved):

PR #1884, opened 2024-02-26
Copy only the telegraf service file from common directory

simathih:


AzureMonitorAgent/packaging.sh
```diff
@@ -30,8 +30,7 @@ cp ../Common/WALinuxAgent-2.0.16/waagent .
 
 cp -r  ../LAD-AMA-Common/metrics_ext_utils .
 cp -r  ../LAD-AMA-Common/telegraf_utils .
-cp -r  ../Diagnostic/services .
-cp -f ../AzureMonitorAgent/services/metrics-extension.service services/metrics-extension.service
```
thperapp, line 34: why are we removing this?

thperapp (approved):

AzureMonitorAgent/packaging.sh
simathih: src and dst are the same, this is redundant

PR #1892, opened 2024-03-08
Fix Ubuntu 22 xfs and partuuid errors.

vimish:
This PR fixes below issues:
1. OS encryption failure on Ubuntu 22
    For some reason, blkid or lsblk is not returning root partition's PARTUUID at the time we request it. The cause is still unclear.
    Updated the logic to not rely on the command ans use partuuid symlinks to get root partition's PARTUUID.

2. XFS module corruption after OS disk encryption on Ubuntu 22
    conv=sparse and bs=64K seems to be skipping some required blocks encryption causing the module corruption.
    Update bs to 16M so that blocks are not skipped.


bamuir (approved):
Looks good to me

PR #1895, opened 2024-03-13
add debian12 support

peizhen-w:
Add Debian 12 support
Debian 12 does not have rsyslog by default, install rsyslog if not present

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -326,10 +326,11 @@ def install():
 
     # Check if Debian 12 VMs have rsyslog package (required for AMA 1.31+)
     if (vm_dist.startswith('debian')) and vm_ver.startswith('12'):
-        check_rsyslog, _ = run_command_and_log("dpkg -l | grep rsyslog")
+        check_rsyslog, _ = run_command_and_log("dpkg -s rsyslog")
         if check_rsyslog != 0:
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -324,6 +324,15 @@ def install():
             if insserv_exit_code != 0:
                 return insserv_exit_code, insserv_output
 
+    # Check if Debian 12 VMs have rsyslog package (required for AMA 1.31+)
+    if (vm_dist.startswith('debian')) and vm_ver.startswith('12'):
+        check_rsyslog, _ = run_command_and_log("dpkg -l | grep rsyslog")
+        if check_rsyslog != 0:
+            hutil_log_info("'rsyslog' package missing from Debian 12 machine, installing to allow AMA to run.")
+            rsyslog_exit_code, rsyslog_output = run_command_and_log("apt update && apt install -y rsyslog")
```
hestolz, line 332: in addition to `-y` you may need to set [non-interactive debian frontend](https://askubuntu.com/questions/556385/how-can-i-install-apt-packages-non-interactively)

hestolz:
let's make sure that when the [distro compat table](https://learn.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview#linux) is updated for 1.31 or w/e release supports debian 12, there is a footnote for deb 12 explaining that rsyslog is required for ama functionality and will be installed and configured if not present.

AaronRM (approved):

PR #1904, opened 2024-03-21
fix imp python module for versions higher than 3.4

NarineM:
Based on https://docs.python.org/3.6/library/imp.html imp has been deprecated starting python version 3.4

### DIFF
--- a/Utils/WAAgentUtil.py
+++ b/Utils/WAAgentUtil.py
@@ -17,9 +17,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import imp
 import os
 import os.path
+import sys
+
+# imp was deprecated in python 3.6
+if sys.version_info >= (3, 6):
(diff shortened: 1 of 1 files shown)


NarineM:
Not able to repro on python 3.6, Can run import imp. Putting this change on hold.

matthew-rollitt:
This isn't enough https://docs.python.org/dev/whatsnew/3.12.html#imp

on line 54 you also need to do this 

import importlib.util
import importlib.machinery

def load_source(modname, filename):
    loader = importlib.machinery.SourceFileLoader(modname, filename)
    spec = importlib.util.spec_from_file_location(modname, filename, loader=loader)
    module = importlib.util.module_from_spec(spec)
    # The module is always executed and not cached in sys.modules.
    # Uncomment the following line to cache the module.
    # sys.modules[module.__name__] = module
    loader.exec_module(module)
    return module

PR #1914, opened 2024-04-23
Onboard Ubuntu 24.04 to online encryption and add support for RHEL 9.3

vimish:
Ubuntu now has sperate boot partition that enables ADE to onboard it to online encryption path. 

Currently tested with Ubuntu 23.10.


bamuir (approved):
Looks good to me.

PR #1926, opened 2024-06-13
Copy libs to test directory before running ldd to check dependencies

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -301,6 +302,9 @@ def copy_mdsd_binaries():
     mdsd_bin = "/opt/microsoft/azuremonitoragent/bin/mdsd"
     mdsdmgr_bin = "/opt/microsoft/azuremonitoragent/bin/mdsdmgr"
 
+    # copy the required libs to our test directory first
+    copytree("/opt/microsoft/azuremonitoragent/lib", os.getcwd() + "/lib")
```
AaronRM, line 306: We should delete after the tests are run.

AaronRM (changes requested):

AaronRM (approved):

PR #1929, opened 2024-06-24
If VM does associated with single MSI, then don't force to provide in public setting

pankajosh:
In last commit, it is forced to user for providing MSI id even VM associated with only one MSI. 
MSI must be provided if VM is associated with multiple identities. If VM associated with single identity providing MSI information in public setting is optional.


canfikret (approved):

pankajosh:
reverting changes for enforcing identity in ADE public setting. 
There is no need to provide identity for SKR if VM is associated with only one identity.

PR #1952, opened 2024-08-10
Python 2.7 compat and fluent-bit ldd check

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -312,7 +313,9 @@ def copy_mdsd_fluentbit_binaries():
         compare_and_copy_bin(mdsd_bin_local_path, mdsd_bin)
         compare_and_copy_bin(mdsdmgr_bin_local_path, mdsdmgr_bin)
 
-    compare_and_copy_bin(fluentbit_bin_local_path, fluentbit_bin)
+    canUseSharedfluentbit, _ = run_command_and_log('ldd ' + fluentbit_bin_local_path + ' | grep "not found"')
+    if canUseSharedfluentbit != 0:
```
NarineM, line 317: should not there be some logging for else condition and what is used if ldd fails

simathih: Default is to use statically linked fluent bit

thperapp (approved):

PR #1955, opened 2024-08-15
Fix unexpected keyword argument 'crypt_id' on FreeBSD 14.1

kro-cat:
fix(distroutils): encode cmd_input argument for ext_utils.run_send_stdin()
    
    The method ext_utils.run_send_stdin() (in extensionutils.py) requires
    the second argument, cmd_input, to be a bytes object, not a string
    object. This is due to the usage of: subprocess.Popen().communicate().
    Encode the password as utf-8 when passing to ext_utils.run_send_stdin().
----

    fix(distroutils): add **kwargs to FreeBSDDistro.chpasswd()
    
    When operating FreeBSD, the argument crypt_id is passed to an
    overridden method chpasswd() in the FreeBSDDistro class. The definition
    of this method does not contain an argument for 'crypt_id' or
    'salt_len'. To address this issue, add **kwargs to the definition of
    chpasswd() in the FreeBSDDistro class.

----

fixes: #1839

### DIFF
--- a/Utils/distroutils.py
+++ b/Utils/distroutils.py
@@ -281,7 +281,7 @@ class FreeBSDDistro(GenericDistro):
 
     # noinspection PyMethodOverriding
     def chpasswd(self, user, password, **kwargs):
-        return ext_utils.run_send_stdin(['pw', 'usermod', 'user', '-h', '0'], password, log_cmd=False)
+        return ext_utils.run_send_stdin(['pw', 'usermod', 'user', '-h', '0'], password.encode('utf-8'), log_cmd=False)
 
     def create_account(self, user, password, expiration, thumbprint, enable_nopasswd):
         """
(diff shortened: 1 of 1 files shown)


kro-cat:
@microsoft-github-policy-service agree

Utils/distroutils.py
```diff
@@ -280,8 +280,8 @@ def __init__(self, config):
 
 
     # noinspection PyMethodOverriding
-    def chpasswd(self, user, password):
-        return ext_utils.run_send_stdin(['pw', 'usermod', 'user', '-h', '0'], password, log_cmd=False)
+    def chpasswd(self, user, password, **kwargs):
```
D1v38om83r, line 283: Explicitly match the number of arguments and types with the base generic distro definition. Even if they might not be used?

kro-cat: Yeah, that would be fine.  Both solutions would be just as good.

kro-cat: Yeah, that would be fine.  Both solutions would be just as good.

D1v38om83r:
Please also get approval from @narrieta

PR #1962, opened 2024-08-22
Remove renaming of older status files with _ suffix

ankitha-p:
Older WALA code had the logic to read the status file with highest sequence number. So VMSnapshotLinux extension would rename all other seq# files with _status suffix. But current WALA code reads the status of the currently executing sequence number. Renaming the status files are not needed. Guest agent has logic only to clear .status files not the ones with _status suffix. This means, older files can pile up in the VM as they don't get cleaned up.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -61,6 +61,12 @@ elif sys.version_info[0] == 2:
     from urlparse import urlparse
     import urllib2 as urlerror
 
+# python shim can only make IMDS calls which shouldn't go through proxy
+try:
+    urllib.getproxies = lambda x = None: {}
+except Exception as e:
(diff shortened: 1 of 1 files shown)


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -116,7 +116,11 @@ def __init__(self, log, error, short_name):
 
     def _get_log_prefix(self):
         return '[%s-%s]' % (self._context._name, self._context._version)
-
+    
+    # It is possible in certain restore scenarios, where old setting files are present in the config folder
+    # When new VM is created with same extension, the seq# starts with 0 again.
```
iamashwani-who, line 121: probably better to explain how older setting files are retained in a new VM? I guess you are referring to the case when OS disk is created from a restore point or a snaphost.

PR #1964, opened 2024-08-23
support ubuntu24.04 with python3.12

marcurdy:
Supporting Python3.12 requires adding "r" for raw in regexes and replacing platform for module distro. This change needs to be tested on FreeBSD. Additionally, to allow ubuntu 24.04 to work, the supported distros list was expanded.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -39,6 +39,7 @@ import re
 import hashlib
 import fileinput
 import contextlib
+import distro
 from collections import OrderedDict
 from hashlib import sha256
 from shutil import copyfile
@@ -812,7 +813,7 @@ def handle_mcs_config(public_settings, protected_settings, default_configs):
                 bypass = json_data["proxy.bypass"]
                 if bypass == "AMA":
                     BypassProxy = True
-                    
+
             if not BypassProxy and json_data is not None and "proxy.url" in json_data:
                 url = json_data["proxy.url"]
                 # only non-authenticated proxy config is supported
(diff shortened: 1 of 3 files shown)


marcurdy:
@microsoft-github-policy-service agree
@microsoft-github-policy-service agree company="Accenture"

gilbahat:
@nkuchta @D1v38om83r Hi, this is really important for us as ubuntu 24.04 blocked from using azure monitor features.

AzureMonitorAgent/agent.py
```diff
@@ -39,6 +39,7 @@
 import hashlib
 import fileinput
 import contextlib
+import distro
```
dennisjlee, line 42: Sorry for the drive-by comments @marcurdy  - I don't work on this project but I would like to get this issue resolved so my team can use Ubuntu 24.04 hosts on Azure without issue!

distro is a [third-party package](https://pypi.org/project/distro/), and from what I can see in the way that this AzureMonitorAgent project is [set up](https://github.com/Azure/azure-linux-extensions/tree/master/AzureMonitorAgent), the project has no requirements.txt / setup.py or any other mechanism to declare third-party dependencies. So I think this is intended to be deployed as just a single script, and adding a third-party dep will not work?

AzureMonitorAgent/agent.py
```diff
@@ -1651,12 +1652,11 @@ def find_vm_distro(operation):
     # platform commands used below aren't available after Python 3.6
     if sys.version_info < (3,7):
         try:
-            vm_dist, vm_ver, vm_id = platform.linux_distribution()
+            vm_dist = distro.id()
```
dennisjlee, line 1655: since you're adding fixes for Python 3.12 and higher, I'd leave this branch alone that was meant for Python <3.7

AzureMonitorAgent/agent.py
```diff
@@ -1651,12 +1652,11 @@ def find_vm_distro(operation):
     # platform commands used below aren't available after Python 3.6
```
dennisjlee, line 1652: This comment and check seem wrong actually - `platform.dist` and `platform.linux_distribution` were documented in Python 3.7 and removed in Python 3.8.

Common/WALinuxAgent-2.0.16/waagent
```diff
@@ -47,6 +47,7 @@ import zipfile
 import json
 import datetime
 import xml.sax.saxutils
+import distro
```
dennisjlee, line 50: Same comment as above - waagent seems to be set up to be a single file script, so I don't think third-party dependencies would be allowed

AzureMonitorAgent/agent.py
gilbahat: on my ubuntu 24.04, distro comes ootb even in minimized installations. IMHO dropping it would be wrong, but it would be prudent to put it behind a try clause.

PR #1965, opened 2024-08-29
Handle non-ascii output from running commands in python wrapper

simathih:
fix for https://portal.microsofticm.com/imp/v5/incidents/details/536301475/summary

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -2229,11 +2229,11 @@ def run_get_output(cmd, chk_err = False, log_cmd = True):
             exit_code = e.returncode
             output = e.output
 
-    if type(output) == unicode:
+    if type(output) == 'unicode':
         output = output.encode('utf-8')
 
     # On python 3, encode returns a byte object, so we must decode back to a string
-    if sys.version_info >= (3,) and type(output) == bytes:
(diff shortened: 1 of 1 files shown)


hestolz:
is this verified on python2 and python3 both?

simathih:
> is this verified on python2 and python3 both?

yeah verified in python 2 and 3:
https://onecompiler.com/python2/42qmmxpjm
https://onecompiler.com/python/42qmmw9eu

AzureMonitorAgent/agent.py
```diff
@@ -2229,7 +2229,7 @@ def run_get_output(cmd, chk_err = False, log_cmd = True):
             exit_code = e.returncode
             output = e.output
 
-    if type(output) == str:
+    if all(ord(c) < 128 for c in output):
```
hestolz, line 2232: what about doing `errors='replace'` instead of checking the input?

simathih: this works in python3 but not python2:
 txt = txt.encode("utf-8", errors='replace')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 13: ordinal not in range(128)

https://onecompiler.com/python2/42qmtwtm5

hestolz (approved):

PR #1966, opened 2024-09-04
Update bls cmdline if grub2 version is greater than 2.06.69

anshuljain26:
Boot filesystem mount failure on RHEL 9 ( >= 2.06-69) after enabling ADE feature on a VM. 
The VM will not boot and the serial console log shows that ADE tried to update BLS cmdline with grub2-mkconfig to add rd.luks.ade.partuuid and rd.luks.ade.bootuuid parameters.


vimish (approved):

anshuljain26:
@microsoft-github-policy-service agree company="Microsoft"

PR #1969, opened 2024-09-12
Changes to Support CVMV2 / Vmmd Blob Snapshot

arisettisanjana:
To support CVMv2 feature that includes an additional VmmdBlob.
Solution:
When a Vmmd or Vmgs Blob is present, the snapshot mode will be configured to take the snapshot through the host. Passing an additional setting to BHS, which is configured based on the input received by extensions (i.e., isVmmdBlobIncluded) from the CRP in order to control the snapshotting process at BHS.

Files Modified:
VMBackup/main/parameterparser.py-> Modified to adjust the setting value for scenarios where the VMMDBlob is included.


iamashwani-who (approved):
Looks good

ankitha-p (approved):

PR #1979, opened 2024-10-14
Fixed Compatibility issue with timestamp.total_seconds() method for python version <=2.6

vipunj:
This PR is regarding fix of Compatibility Issue with total_seconds() for python version <=2.6
There was a compatibility issue of total_seconds() whith python version < = 2.6, so updated it to use timedelta which is compatible for all versions of Python, including Python 2.6, 2.7, and 3.x. New implementation utilizes total_seconds() method if available, otherwise falls back to custom implementation that converts it using custom logic to calculate elapsed seconds (delta.days * 86400 + delta.seconds + delta.microseconds / 1e6)

### DIFF
--- a/VMBackup/main/hostsnapshotter.py
+++ b/VMBackup/main/hostsnapshotter.py
@@ -180,7 +180,8 @@ class HostSnapshotter(object):
                         except:
                             creationTimeObj = datetime.datetime.strptime(creationTimeString, "%Y-%m-%dT%H:%M:%SZ")
                         self.logger.log("Converting the creationTime string received in UTC format to UTC Ticks")
-                        timestamp = (creationTimeObj - epochTime).total_seconds()
+                        delta = creationTimeObj - epochTime
+                        timestamp = delta.days * 86400 + delta.seconds + delta.microseconds / 1e6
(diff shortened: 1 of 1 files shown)


VMBackup/main/hostsnapshotter.py
```diff
@@ -180,7 +180,8 @@ def get_snapshot_info(self, responseBody):
                         except:
                             creationTimeObj = datetime.datetime.strptime(creationTimeString, "%Y-%m-%dT%H:%M:%SZ")
                         self.logger.log("Converting the creationTime string received in UTC format to UTC Ticks")
-                        timestamp = (creationTimeObj - epochTime).total_seconds()
+                        delta = creationTimeObj - epochTime
+                        timestamp = delta.days * 86400 + delta.seconds + delta.microseconds / 1e6
```
iamashwani-who, line 184: can we create a helper method to getTotalSeconds and inside that method, add a check if total_seconds is unsupported then only use the backup method?

iamashwani-who: also add a log statement for using the fallback logic

iamashwani-who (approved):

ankitha-p (approved):

PR #1982, opened 2024-10-16
ADE Linux Extension Release - 1.4.0.8

anshuljain26:


bamuir (approved):
Looks good to me; I can help with the release, assuming we have done a full automation run to make sure the changes are good. I noticed also the change is only for redhat; would this be applicable to other distros?

PR #1984, opened 2024-10-18
Remove proxy files if proxy config is removed

simathih:


peizhen-w (approved):

AzureMonitorAgent/agent.py
```diff
@@ -949,8 +956,32 @@ def set_proxy(address, username, password):
...
+            os.remove("/etc/systemd/system/azuremonitor-coreagent.service.d/proxy.conf")
+            hasSettings=True
+            
+        # Update ME
+        if os.path.exists("/etc/systemd/system/metrics-extension.service.d/proxy.conf"):
+            os.remove("/etc/systemd/system/metrics-extension.service.d/proxy.conf")
+            hasSettings=True
+            
+        if hasSettings:
+            run_command_and_log("systemctl daemon-reload")
```
hestolz, line 980: this requires a service restart as well after daemon-reload -- are we doing that later after this gets called? (I assume so, just double checking)

simathih: we were previously not doing it but it worked with just daemon-reload but I have explicitly added it as well now

hestolz (approved):

PR #1991, opened 2024-10-30
Fix null characters in guest agent stdout.

vimish:
sys.stderr.write was adding null characters to guest agent stdout.
These null characters were causing failure in guest agent telemetry.

Also, removed a log line that was masking extension exit message to go guest agent telemetry.


bamuir (approved):
Do we know if/how this will impact the QOS telemetry data in kusto?

PR #1996, opened 2024-11-13
Updated python installation path search in VmSnapshotLinux for FreeBSD

vipunj:
Application consistent Restore Point for Free BSD Linux distro was not working because of the way VmSnapshotLinux extension looks for python installation. On FreeBSD python is installed into `/usr/local/bin` whereas the current implementation was looking for python installation only in /usr/bin/ path.
So modified the script to check into /usr/local/bin if it is not there at /usr/bin.

### DIFF
--- a/VMBackup/main/handle.sh
+++ b/VMBackup/main/handle.sh
@@ -58,9 +58,9 @@ do
 	if [ ! -f "${cmnd}" ]
 	then
 		cmnd="/usr/local/bin/${pythonVersion}"
-    fi
+	fi
 	if [ -f "${cmnd}" ]
-    then
+	then
 		echo "`date -u`- ${pythonVersion} path exists" >> $logfile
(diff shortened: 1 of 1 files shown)


VMBackup/main/handle.sh
```diff
@@ -55,6 +55,10 @@ pythonVersionList="python3.8 python3.7 python3.6 python3.5 python3.4 python3.3 p
 for pythonVersion in ${pythonVersionList};
 do
 	cmnd="/usr/bin/${pythonVersion}"
+	if [ ! -f "${cmnd}" ]
+	then
```
iamashwani-who, line 59: formatting.

iamashwani-who (approved):

arisettisanjana:
Hey looks good to me, did we test to confirm if it's working now with this change.

PR #2002, opened 2024-12-12
Remove insserv-compat check as this is no longer required in AMA

thperapp:


simathih (approved):

simathih:
why is this no longer needed dependency?

PR #2008, opened 2025-01-10
log imds url to improve diagnosability

NarineM:
we have seen a few incidents (Arc and Cloud) where due to hardening or firewall rules customer used to see failure from python code but not able to determine the actual url that ama is querying  post ama auto upgrade/during enable calls. With this change ama will  log imds url info for easy troubleshooting.

### DIFF
--- a/LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
+++ b/LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
@@ -580,7 +580,7 @@ def get_handler_vars():
     return logFolder, configFolder
 
 
-def get_imds_values(is_lad):
+def get_imds_values(is_lad, HUtilObj=None):
     """
     Query imds to get required values for MetricsExtension config for this VM
     """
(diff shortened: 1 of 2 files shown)


simathih (approved):

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -600,6 +600,12 @@ def get_imds_values(is_lad):
         else:
             imds_url = "http://169.254.169.254/metadata/instance?api-version=2019-03-11"
 
+    message = "IMDS url to query: " + imds_url
+    if HUtilObj is not None:
+        HUtilObj.log(message)
+    else:
+        print('Info: {0}'.format(message))
```
hestolz, line 607: how frequently is this going to be logged?

NarineM: during enable call. this is needed for troubleshooting connectivity issuesthat can happen post ama install.

hestolz: need to ensure this does not cause repeated logs. see multiple incidents regarding extension.log size explosion

PR #2011, opened 2025-01-30
Fix for persistent fluent-bit config update without any port change

simathih:
Fix for https://portal.microsofticm.com/imp/v5/incidents/details/578913013/summary

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -1185,7 +1185,7 @@ def metrics_watcher(hutil_error, hutil_log):
                 defaultPortSetting = 'Port'
                 portUpdated = False
                 with open(FluentCfgPath) as f:
-                    found = re.search(".*Port.*" + fluent_port + "$", f.read())
+                    found = re.search(".*Port.*" + fluent_port, f.read())
                     if not found:
                         portUpdated = True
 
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -1181,8 +1181,8 @@ def metrics_watcher(hutil_error, hutil_log):
                 f.close()
             
             if fluent_port != '' and os.path.isfile(FluentCfgPath):
-                portSetting = "    Port    "  + fluent_port
-                defaultPortSetting = 'Port    '
+                portSetting = "    Port                       "  + fluent_port
```
AaronRM, line 1184: Is there a reason we can't do a regex match here? Needing to match on the precise amount of space feels fragile.

AzureMonitorAgent/agent.py
```diff
@@ -1181,11 +1181,12 @@ def metrics_watcher(hutil_error, hutil_log):
...
             
             if fluent_port != '' and os.path.isfile(FluentCfgPath):
-                portSetting = "    Port    "  + fluent_port
-                defaultPortSetting = 'Port    '
+                portSetting = "    Port                       "  + fluent_port + "\n"
+                defaultPortSetting = 'Port'
                 portUpdated = False
                 with open(FluentCfgPath) as f:
-                    if portSetting not in f.read():
+                    found = re.search(".*Port.*" + fluent_port, f.read())
```
AaronRM, line 1188: Can we tighten up the regex to use a start of line anchor and only match "Port" surrounded by whitespace? Just thinking about cases where "Port" is elsewhere in the config file (maybe as part of a custom log file path).

AaronRM (changes requested):

AaronRM (approved):

PR #2012, opened 2025-02-03
use appropriate fx and eat exceptions

hestolz:


simathih:
can we please move the urllib.Request here into the try as well? https://github.com/Azure/azure-linux-extensions/blob/10fc3c5a6efbb972301f28ebbbe7a6ef209f030e/AzureMonitorAgent/agent.py#L1938

simathih (approved):

PR #2026, opened 2025-03-27
Add AMA extension.log logrotate

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -1718,6 +1718,13 @@ def parse_context(operation):
             logFileName = 'extension.log'
             hutil = HUtil.HandlerUtility(waagent.Log, waagent.Error, logFileName=logFileName)
             hutil.do_parse_context(operation)
+
+            # As per VM extension team, we have to manage rotation for our extension.log
+            # for now, this is our extension code, but to be moved to HUtil library.
+            if not os.path.exists('/etc/logrotate.d/azuremonitoragentextension'):
```
AaronRM, line 1724: We need to remove this file on uninstall. Otherwise, an update running this path will not patch it.

AaronRM (changes requested):

AzureMonitorAgent/agent.py
simathih: addressed it

AzureMonitorAgent/agent.py
```diff
@@ -450,7 +450,11 @@ def uninstall():
         log_and_exit("Uninstall", UnsupportedOperatingSystem, "The OS has neither rpm nor dpkg" )
     hutil_log_info('Running command "{0}"'.format(AMAUninstallCommand))
 
-    remove_localsyslog_configs()
+    remove_localsyslog_configs()    
+
+    # remove the logrotate config
+    if os.path.exists('/etc/logrotate.d/azuremonitoragentextension'):      
+        os.remove('/etc/logrotate.d/azuremonitoragentextension')
```
AaronRM, line 457: We should include error handling if the file is locked, etc...

AzureMonitorAgent/agent.py
```diff
@@ -450,7 +450,11 @@ def uninstall():
         log_and_exit("Uninstall", UnsupportedOperatingSystem, "The OS has neither rpm nor dpkg" )
     hutil_log_info('Running command "{0}"'.format(AMAUninstallCommand))
 
-    remove_localsyslog_configs()
+    remove_localsyslog_configs()    
+
+    # remove the logrotate config
+    if os.path.exists('/etc/logrotate.d/azuremonitoragentextension'):
```
AaronRM, line 456: We should be using a constant value for this path/filename.

AzureMonitorAgent/agent.py
simathih: added

AzureMonitorAgent/agent.py
simathih: added

AaronRM (approved):

Klaas-:
hi @simathih 
is there a  reason you're only doing `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/extension.log`
and not the other log files?
* `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/CommandExecution.log`
* `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/telegraf.log`

Klaas-:
also there is a subdir `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/events/` but it's empty on the servers I looked at -- does that serve a purpose and needs logrotate as well? Or should the dir be removed?

simathih:
> hi @simathih is there a reason you're only doing `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/extension.log` and not the other log files?
> 
> * `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/CommandExecution.log`
> * `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/telegraf.log`

you are right, we will add log rotate for these too

simathih:
> also there is a subdir `/var/log/azure/Microsoft.Azure.Monitor.AzureMonitorLinuxAgent/events/` but it's empty on the servers I looked at -- does that serve a purpose and needs logrotate as well? Or should the dir be removed?

this folder is already garbage collected by our service so it is handled

PR #2027, opened 2025-03-27
Fix an issue where AMA install can fail if a previous attempt was interrupted

AaronRM:
Fix an issue where AMA install can fail if a previous attempt was interrupted

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -320,7 +320,15 @@ def copy_mdsd_fluentbit_binaries():
     fluentbit_bin = "/opt/microsoft/azuremonitoragent/bin/fluent-bit"
 
     # copy the required libs to our test directory first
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -320,7 +320,15 @@ def copy_mdsd_fluentbit_binaries():
     fluentbit_bin = "/opt/microsoft/azuremonitoragent/bin/fluent-bit"
 
     # copy the required libs to our test directory first
-    copytree("/opt/microsoft/azuremonitoragent/lib", os.getcwd() + "/lib")
+    lib_dir = os.path.join(os.getcwd(), "lib")
+    if os.path.exists(lib_dir):
+        rmtree(lib_dir)
```
hestolz, line 325: is there anything we should do to check `lib_dir` is the path we intend? this may be overthinking, but i'm imagining accidentally removing `/lib`

dylanbun (approved):

PR #2039, opened 2025-04-16
Disable Console Logging

arisettisanjana:
Disabling the console logging by default, can be enabled by setting the ConsoleLoggingOff flag in vmbackup.conf file to 1.


VMBackup/main/handle.py
```diff
@@ -72,7 +72,7 @@ def main():
         total_used_size = 0
         size_calculation_failed = False
         eventlogger = None
-        HandlerUtil.waagent.LoggerInit('/dev/console','/dev/stdout')
+        HandlerUtil.waagent.LoggerInit(None,'/dev/stdout')
```
iamashwani-who, line 75: it might be better to add list of loggers to initialize in a config and pass those instead of checking specific config in backuplogger?

PR #2045, opened 2025-05-01
Refactor code to have one list of supported distros for install and ama troubleshooter

thperapp:


dylanbun (dismissed):

AzureMonitorAgent/ama_tst/modules/install/check_os.py
NarineM, line 1: You ll need to update the tool version and publish the new tool as an archive for these changes to be taken by  ama build pipeline. From what I remember we have not moved packaging of TST tool into build pipeline. Can you please check that and add the new tg as part of the PR? A

NarineM (changes requested):

AzureMonitorAgent/ama_tst/modules/install/check_os.py
thperapp: done

NarineM (approved):

PR #2059, opened 2025-05-28
Sajithmuh patch 1.35

sajithmuh:
* Enable metrics-extension service always
    * Fix missing agent version in metrics-extension-cmv2.service

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -494,6 +494,9 @@ def install():
     # Needs to be revisited for aarch64
     copy_kqlextension_binaries()
 
+    # Install azureotelcollector
+    install_azureotelcollector()
+
     # Copy mdsd and fluent-bit with OpenSSL dynamically linked
     if is_feature_enabled('useDynamicSSL'):
         # Check if they have libssl.so.1.1 since AMA is built against this version
@@ -553,8 +556,7 @@ def uninstall():
 
     remove_localsyslog_configs()
 
-    # remove azureotelcollector service
-    remove_azureotelcollector()
+    uninstall_azureotelcollector()
 
     # remove the logrotate config
     if os.path.exists(AMAExtensionLogRotateFilePath):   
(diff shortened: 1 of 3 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -1481,9 +1480,43 @@ def metrics_watcher(hutil_error, hutil_log):
     last_crc = None
     last_crc_fluent = None
     me_msi_token_expiry_epoch = None
+    enabled_me_CMv2_mode = False
+    log_messages = ""
 
     while True:
         try:
+            if not me_handler.is_running(is_lad=False):
+                me_service_template_path = os.getcwd() + "/services/metrics-extension.service"
```
simathih, line 1489: maybe delete the service template file only inside the if or elif below? since this is an aggressive while loop

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
```diff
@@ -394,6 +394,31 @@ def setup_me_service(is_lad, configFolder, monitoringAccount, metrics_ext_bin, m
     return True
 
 
+def start_metrics_cmv2():
```
simathih, line 397: why do we need a separate method for cmv2?

AzureMonitorAgent/agent.py
sajithmuh: > maybe delete the service template file only inside the if or elif below? since this is an aggressive while loop

Thanks, moved it within if and elif statements below. However, if the ME process is not running yet, it will still perform the operation as CMv2 feature flag is enabled everywhere. Moving it inside, just in case we want to move back to CMv1 by disabling the CMv2 feature flag.

We can move it back to as it was when we no longer need the feature flag and remove all CMv1 code.

LAD-AMA-Common/metrics_ext_utils/metrics_ext_handler.py
sajithmuh: > why do we need a separate method for cmv2?
Moving it to a separate method, as `start_metrics` supports other scenarios (CMv1/LAD), `start_metrics_cmv2` only works with CMv2 - meant to work only where systemd is available. Plan is to remove `start_metrics` altogether if there is no objections - I believe there is no need to support LAD specific code here?

simathih: ok we can remove it in future iterations

simathih (approved):

PR #2062, opened 2025-06-03
T feadeaga user error

FeyiAdeaga:
Adding error clarification for vmaccess extension.

### DIFF
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -68,6 +68,7 @@ from os.path import join
 DateTimeFormat = "%Y-%m-%dT%H:%M:%SZ"
 
 MANIFEST_XML = "manifest.xml"
+_error_clarification_name = "ErrorClarification"
 
 
 class HandlerContext:
@@ -83,7 +84,6 @@ class HandlerContext:
         self._status_file = None
         self._settings_file = None
         self._config = None
-        self._error_clarification_name = "ErrorClarification"
         return
 
 
@@ -325,14 +325,14 @@ class HandlerUtility:
             "version": self._context._version,
             "timestampUTC": tstamp,
             "status": {
-            "name": self._context._name,
-            "operation": operation,
-            "status": status,
-            "code": status_code,
-            "formattedMessage": {
-                "lang": "en-US",
-                "message": message
-            }
+                "name": self._context._name,
+                "operation": operation,
+                "status": status,
+                "code": status_code,
+                "formattedMessage": {
+                    "lang": "en-US",
+                    "message": message
+                }
             }
         }]
         if substatus is not None:
@@ -355,7 +355,7 @@ class HandlerUtility:
             substatus = None
             if error_code is not None:
                 substatus = [{
-                    "Name": self._context._error_clarification_name,
+                    "Name": _error_clarification_name,
                     "Code": error_code
                 }]
             self.do_status_report(operation, status, code, message,error_code, substatus)
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -83,6 +83,7 @@ class HandlerContext:
         self._status_file = None
         self._settings_file = None
         self._config = None
+        self._error_clarification_name = "ErrorClarification"
         return
 
 
@@ -333,7 +334,7 @@ class HandlerUtility:
                     "message": message
                 }
             } , 
-            "SubStatus": substatus
+            "substatus": substatus
         }]
         stat_rept = json.dumps(stat)
         if self._context._status_file:
@@ -348,12 +349,13 @@ class HandlerUtility:
         if ext_utils.set_file_contents(heartbeat_file, health_report) is None:
             self.error('Unable to wite heartbeat info to ' + heartbeat_file)
 
-    def do_exit(self, exit_code, operation, status, code, message,error_code=None):
+    def do_exit(self, exit_code, operation, status, code, message, error_code=None):
         try:
+            substatus = None
             if error_code is not None:
                 substatus = [{
-                    "Name": "ErrorClarification",
-                    "Code": error_code
+                    "Name": self._context._error_clarification_name,
+                    "Code": error_codefh
                 }]
             self.do_status_report(operation, status, code, message,error_code, substatus)
         except Exception as e:
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -351,10 +351,10 @@ class HandlerUtility:
     def do_exit(self, exit_code, operation, status, code, message,error_code=None):
         try:
             if error_code is not None:
-                substatus = {
+                substatus = [{
                     "Name": "ErrorClarification",
                     "Code": error_code
-                }
+                }]
             self.do_status_report(operation, status, code, message,error_code=None, substatus=None)
         except Exception as e:
             self.log("Can't update status: " + str(e))
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -355,7 +355,7 @@ class HandlerUtility:
             if error_code is not None:
                 substatus = [{
                     "Name": self._context._error_clarification_name,
-                    "Code": error_codefh
+                    "Code": error_code
                 }]
             self.do_status_report(operation, status, code, message,error_code, substatus)
         except Exception as e:
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -68,7 +68,7 @@ from os.path import join
 DateTimeFormat = "%Y-%m-%dT%H:%M:%SZ"
 
 MANIFEST_XML = "manifest.xml"
-_error_clarification_name = "ErrorClarification"
+from VMAccess.vmaccess import ErrorClarificationName
 
 
 class HandlerContext:
@@ -355,7 +355,7 @@ class HandlerUtility:
             substatus = None
             if error_code is not None:
                 substatus = [{
-                    "Name": _error_clarification_name,
+                    "Name": ErrorClarificationName,
                     "Code": error_code
                 }]
             self.do_status_report(operation, status, code, message,error_code, substatus)
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -355,7 +355,7 @@ class HandlerUtility:
                     "Name": "ErrorClarification",
                     "Code": error_code
                 }]
-            self.do_status_report(operation, status, code, message,error_code=None, substatus=None)
+            self.do_status_report(operation, status, code, message,error_code, substatus)
         except Exception as e:
             self.log("Can't update status: " + str(e))
         sys.exit(exit_code)
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -315,7 +315,9 @@ class HandlerUtility:
     def _set_most_recent_seq(self, seq):
         ext_utils.set_file_contents('mrseq', str(seq))
 
-    def do_status_report(self, operation, status, status_code, message):
+    def do_status_report(self, operation, status, status_code, message, error_code=None):
+        if error_code is not None:
+            message = "{0} (error code: {1})".format(message, error_code)
         self.log("{0},{1},{2},{3}".format(operation, status, status_code, message))
         tstamp = time.strftime(DateTimeFormat, time.gmtime())
         stat = [{
@@ -345,9 +347,9 @@ class HandlerUtility:
         if ext_utils.set_file_contents(heartbeat_file, health_report) is None:
             self.error('Unable to wite heartbeat info to ' + heartbeat_file)
 
-    def do_exit(self, exit_code, operation, status, code, message):
+    def do_exit(self, exit_code, operation, status, code, message,error_code=None):
         try:
-            self.do_status_report(operation, status, code, message)
+            self.do_status_report(operation, status, code, message,error_code=None)
         except Exception as e:
             self.log("Can't update status: " + str(e))
         sys.exit(exit_code)
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -315,7 +315,7 @@ class HandlerUtility:
     def _set_most_recent_seq(self, seq):
         ext_utils.set_file_contents('mrseq', str(seq))
 
-    def do_status_report(self, operation, status, status_code, message, error_code=None):
+    def do_status_report(self, operation, status, status_code, message, error_code=None, substatus=None):
         if error_code is not None:
             message = "{0} (error code: {1})".format(message, error_code)
         self.log("{0},{1},{2},{3}".format(operation, status, status_code, message))
@@ -332,7 +332,8 @@ class HandlerUtility:
                     "lang": "en-US",
                     "message": message
                 }
-            }
+            } , 
+            "SubStatus": substatus
         }]
         stat_rept = json.dumps(stat)
         if self._context._status_file:
@@ -349,7 +350,12 @@ class HandlerUtility:
 
     def do_exit(self, exit_code, operation, status, code, message,error_code=None):
         try:
-            self.do_status_report(operation, status, code, message,error_code=None)
+            if error_code is not None:
+                substatus = {
+                    "Name": "ErrorClarification",
+                    "Code": error_code
+                }
+            self.do_status_report(operation, status, code, message,error_code=None, substatus=None)
         except Exception as e:
             self.log("Can't update status: " + str(e))
         sys.exit(exit_code)
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -325,17 +325,18 @@ class HandlerUtility:
             "version": self._context._version,
             "timestampUTC": tstamp,
             "status": {
-                "name": self._context._name,
-                "operation": operation,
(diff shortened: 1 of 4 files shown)


Utils/handlerutil2.py
```diff
@@ -345,9 +348,14 @@ def do_heartbeat_report(self, heartbeat_file, status, code, message):
...
-    def do_exit(self, exit_code, operation, status, code, message):
+    def do_exit(self, exit_code, operation, status, code, message,error_code=None):
         try:
-            self.do_status_report(operation, status, code, message)
+            if error_code is not None:
+                substatus = [{
+                    "Name": "ErrorClarification",
+                    "Code": error_code
+                }]
+            self.do_status_report(operation, status, code, message,error_code, substatus)
```
mayankdaruka-msft, line 361: You are only assigning to substatus within the `if` block. If that block is not executed, you will run into an error where substatus is being referenced before assignment.

mayankdaruka-msft: This will likely be the case with install, uninstall, disable, update. Those scenarios would need to be tested even though they contain minimal logic.

Utils/handlerutil2.py
```diff
@@ -330,7 +332,8 @@ def do_status_report(self, operation, status, status_code, message):
                     "lang": "en-US",
                     "message": message
                 }
-            }
+            } , 
+            "SubStatus": substatus
```
mayankdaruka-msft: Nit: "substatus" instead of "SubStatus".

mayankdaruka-msft: When CRP deserializes the aggregate status, it expects the substatus list to be provided via the "substatus" property:

image

VMAccess/vmaccess.py
```diff
@@ -244,7 +266,7 @@ def _set_user_account_pub_key(protect_settings, hutil):
 
     # user name must be provided if set ssh key or password
     if not protect_settings or 'username' not in protect_settings:
-        return
+        return ("None", None)
```
norakoiralamsft: Can we add to the list of error codes? If we can, maybe we should add an error code for username error, or bundle this under USER_ERROR.

Utils/handlerutil2.py
```diff
@@ -345,9 +348,14 @@ def do_heartbeat_report(self, heartbeat_file, status, code, message):
         if ext_utils.set_file_contents(heartbeat_file, health_report) is None:
             self.error('Unable to wite heartbeat info to ' + heartbeat_file)
 
-    def do_exit(self, exit_code, operation, status, code, message):
+    def do_exit(self, exit_code, operation, status, code, message,error_code=None):
```
norakoiralamsft: nit: space between , and error_code (similarly for line 201)

VMAccess/vmaccess.py
```diff
@@ -165,19 +168,37 @@ def enable():
 
         if remove_user:
             ext_utils.add_extension_event(name=hutil.get_name(), op="scenario", is_success=True, message="remove-user")
-            _remove_user_account(remove_user, hutil)
+            errorstatus,errormsg = _remove_user_account(remove_user, hutil)
+            if errorstatus is not None:
+                raise Exception(errormsg)
+
+
```
norakoiralamsft: nit: extra line break (similarly for lines 195-196)

VMAccess/vmaccess.py
```diff
@@ -165,19 +168,37 @@ def enable():
 
         if remove_user:
             ext_utils.add_extension_event(name=hutil.get_name(), op="scenario", is_success=True, message="remove-user")
-            _remove_user_account(remove_user, hutil)
+            errorstatus,errormsg = _remove_user_account(remove_user, hutil)
+            if errorstatus is not None:
+                raise Exception(errormsg)
+
+        
+        output = _set_user_account_pub_key(protect_settings, hutil)
```
norakoiralamsft, line 176: Consider saving the output similarly to line 171.

VMAccess/vmaccess.py
```diff
@@ -559,17 +581,22 @@ def check_and_repair_disk(hutil):
             err_msg = ("check_disk and repair_disk was both specified."
                        "Only one of them can be specified")
             hutil.error(err_msg)
-            hutil.do_exit(1, 'Enable', 'error', '0', 'Enable failed.')
+            return error_code.ErrorCode.CHECK_REPAIR_DISK_ERROR
+            # hutil.do_exit(1, 'Enable', 'error', '0', 'Enable failed.', errorstatus.value)
```
norakoiralamsft: Remove

Utils/handlerutil2.py
```diff
@@ -315,7 +315,9 @@ def set_inused_config_seq(self, seq):
     def _set_most_recent_seq(self, seq):
         ext_utils.set_file_contents('mrseq', str(seq))
 
-    def do_status_report(self, operation, status, status_code, message):
+    def do_status_report(self, operation, status, status_code, message, error_code=None, substatus=None):
```
norakoiralamsft, line 319: Not super familiar with the testing for vmaccess linux, is there a way we can unit test or create LISA tests for these changes?

Utils/handlerutil2.py
```diff
@@ -345,9 +348,14 @@ def do_heartbeat_report(self, heartbeat_file, status, code, message):
         if ext_utils.set_file_contents(heartbeat_file, health_report) is None:
             self.error('Unable to wite heartbeat info to ' + heartbeat_file)
 
-    def do_exit(self, exit_code, operation, status, code, message):
+    def do_exit(self, exit_code, operation, status, code, message,error_code=None):
         try:
-            self.do_status_report(operation, status, code, message)
+            if error_code is not None:
+                substatus = [{
+                    "Name": "ErrorClarification",
```
malhotrarohit: NIT - The word "ErrorClassification" should be put in a place where constants in the repo are kept. Generally speaking it is a good idea to have all magic strings, magic numbers etc. put in dedicated files and reference the constants through that instead of putting the literals in everywhere in code.

VMAccess/vmaccess.py
```diff
@@ -578,15 +605,14 @@ def _fsck_check(hutil):
     try:
         retcode = ext_utils.run(['fsck', '-As', '-y'])
         if retcode > 0:
-            hutil.log(retcode)
+            hutil.log(str(retcode))
```
mayankdaruka-msft: Instead of raising an exception and catching it both in `_fsck_check`, why not just return the error code here?

VMAccess/vmaccess.py
```diff
@@ -559,17 +581,22 @@ def check_and_repair_disk(hutil):
...
             hutil.error(err_msg)
-            hutil.do_exit(1, 'Enable', 'error', '0', 'Enable failed.')
+            return error_code.ErrorCode.CHECK_REPAIR_DISK_ERROR
+            # hutil.do_exit(1, 'Enable', 'error', '0', 'Enable failed.', errorstatus.value)
 
         if check_disk:
             ext_utils.add_extension_event(name=hutil.get_name(), op="scenario", is_success=True, message="check_disk")
             outretcode = _fsck_check(hutil)
+            if outretcode is not None:
+                return outretcode
```
mayankdaruka-msft: We already return `outretcode` at the end of this function, so this feels slightly redundant. If this is about the "Successfully checked disk" log, you could log that in `_fsck_check` if retcode indicates success.

Same comment applies for repair disk.

Utils/handlerutil2.py
mayankdaruka-msft: Thanks Nora for bringing this up. @FeyiAdeaga I forgot to mention we have LISA tests for VMAccess Linux. Let's go over that sometime next week.

VMAccess/errrorclarification.py
```diff
@@ -0,0 +1,9 @@
+import enum
+
+class ErrorCode(enum.Enum):
+    SYSTEM_ERROR = 0
+    USER_ERROR = 1
```
jscalev: What does USER_ERROR mean? We should strive for more specific error codes such as those below.

VMAccess/vmaccess.py
```diff
@@ -165,19 +168,37 @@ def enable():
...
+            if errorstatus is not None:
+                raise Exception(errormsg)
+
+        
+        output = _set_user_account_pub_key(protect_settings, hutil)
 
-        _set_user_account_pub_key(protect_settings, hutil)
+        if output is not None:
+                errorstatus,errormsg = output[0], output[1]
+                raise Exception(errormsg)
```
jscalev: Not sure we want to throw an exception here since then the status wouldn't be written.

Utils/handlerutil2.py
FeyiAdeaga: No problem

VMAccess/vmaccess.py
FeyiAdeaga: From the previous implementation of _set_user_account_pub_key, anywhere they put a message and raised an exception I replaced with return the message and error status so I could raise the exception in the main function instead.

jscalev (approved):

Utils/handlerutil2.py
```diff
@@ -330,7 +333,8 @@ def do_status_report(self, operation, status, status_code, message):
                     "lang": "en-US",
                     "message": message
                 }
-            }
+            } , 
+            "substatus": substatus
```
D1v38om83r: NIT: don't set substatus if its none?

VMAccess/errorclarification.py
```diff
@@ -0,0 +1,8 @@
+import enum
+
+class ErrorCode(enum.Enum):
+    SYSTEM_ERROR = 0
+    USER_PASSWORD_ERROR = 1
+    REPAIR_DISK_ERROR = 2
+    CHECK_DISK_ERROR = 3
+    CHECK_REPAIR_DISK_ERROR = 4
```
D1v38om83r: NIT: no newline at the end of file.

VMAccess/vmaccess.py
```diff
@@ -221,12 +238,13 @@ def _remove_user_account(user_name, hutil):
...
+        return error_code.ErrorCode.SYSTEM_ERROR, "Failed to remove user {0}".format(e)
 
     ext_utils.add_extension_event(name=hutil.get_name(),
                                   op=constants.WALAEventOperation.Enable,
                                   is_success=True,
                                   message="Successfully removed user")
+    return None, "Successfully removed user {0}".format(user_name)
 
 
 def _set_user_account_pub_key(protect_settings, hutil):
```
D1v38om83r, line 251: raise exception in line 310, 354 may need to be replaced with returning error code and message. Return None in line 265?

VMAccess/vmaccess.py
```diff
@@ -165,19 +168,33 @@ def enable():
...
+        if errorstatus is not None:
+            if errorstatus == error_code.ErrorCode.CHECK_REPAIR_DISK_ERROR:
+                hutil.do_exit(1, 'Enable', 'error', '0', 'Enable failed.', errorstatus.value)
+            if errorstatus == error_code.ErrorCode.CHECK_DISK_ERROR:
+                hutil.do_exit(1, 'Enable', 'error', '0', 'Check disk failed.', errorstatus.value)
+            if errorstatus == error_code.ErrorCode.REPAIR_DISK_ERROR:
+                hutil.do_exit(1, 'Enable', 'error', '0', 'Repair disk failed.', errorstatus.value)
+            
         hutil.do_exit(0, 'Enable', 'success', '0', 'Enable succeeded.')
     except Exception as e:
```
D1v38om83r, line 195: I would recommend calling hutil.do_exit explicitly where possible and then use a blanket catch at the end to handle cases that weren't accounted for. For example line 179, it can be called explicitly.

D1v38om83r: Just like line 184 to 191

VMAccess/vmaccess.py
FeyiAdeaga: I added a return None in line 265. In line 310 I raise an exception so it the message is managed in except block on the enable function. The exception on line 354 is managed in the except block of the encompassing try block. I just modified it so whatever the message is sent back to the enable function.

VMAccess/vmaccess.py
FeyiAdeaga: Will fix that

D1v38om83r (dismissed):

Utils/handlerutil2.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, name):
         self._status_file = None
         self._settings_file = None
         self._config = None
+        self._error_clarification_name = "ErrorClarification"
```
mayankdaruka-msft: You could define this as a constant in vmaccess.py (at the top, along with the other global variables)

Utils/handlerutil2.py
```diff
@@ -315,23 +316,27 @@ def set_inused_config_seq(self, seq):
...
             "status": {
-                "name": self._context._name,
-                "operation": operation,
-                "status": status,
-                "code": status_code,
-                "formattedMessage": {
-                    "lang": "en-US",
-                    "message": message
-                }
+            "name": self._context._name,
```
mayankdaruka-msft: Indent this and the lines under it?

VMAccess/vmaccess.py
```diff
@@ -165,19 +168,33 @@ def enable():
 
         if remove_user:
             ext_utils.add_extension_event(name=hutil.get_name(), op="scenario", is_success=True, message="remove-user")
-            _remove_user_account(remove_user, hutil)
+            errorstatus,errormsg = _remove_user_account(remove_user, hutil)
```
mayankdaruka-msft: Nit: Spacing between errorstatus and errormsg

VMAccess/vmaccess.py
```diff
@@ -165,19 +168,33 @@ def enable():
...
+            if errorstatus is not None:
+                raise Exception(errormsg)
 
-        _set_user_account_pub_key(protect_settings, hutil)
+        output = _set_user_account_pub_key(protect_settings, hutil)
 
+        if output is not None:
+                errorstatus,errormsg = output[0], output[1]
```
mayankdaruka-msft: Nit: Spacing here and the line underneath as well.

VMAccess/vmaccess.py
```diff
@@ -559,33 +577,30 @@ def check_and_repair_disk(hutil):
...
-        else:
-            return retcode
-    except Exception as e:
+    retcode = ext_utils.run(['fsck', '-As', '-y'])
+    if retcode > 0:
+        hutil.log(str(retcode))
         hutil.error("Failed to run disk check with error: {0}, {1}".format(
-            str(e), traceback.format_exc()))
-        hutil.do_exit(1, 'Check', 'error', '0', 'Check failed.')
+        str("Disk check was not successful"), traceback.format_exc()))
```
mayankdaruka-msft: "Disk check ..." does not need to be in str()

Utils/handlerutil2.py
```diff
@@ -68,6 +68,7 @@
 DateTimeFormat = "%Y-%m-%dT%H:%M:%SZ"
 
 MANIFEST_XML = "manifest.xml"
+_error_clarification_name = "ErrorClarification"
```
mayankdaruka-msft: I meant this can go along with the rest of the global variables in *vmaccess.py*.

VMAccess/vmaccess.py
```diff
@@ -559,33 +577,30 @@ def check_and_repair_disk(hutil):
...
-        else:
-            return retcode
-    except Exception as e:
+    retcode = ext_utils.run(['fsck', '-As', '-y'])
+    if retcode > 0:
+        hutil.log(str(retcode))
         hutil.error("Failed to run disk check with error: {0}, {1}".format(
-            str(e), traceback.format_exc()))
-        hutil.do_exit(1, 'Check', 'error', '0', 'Check failed.')
+        "Disk check was not successful", traceback.format_exc()))
```
mayankdaruka-msft: Nit: Extra tab here?

VMAccess/vmaccess.py
```diff
@@ -559,33 +578,29 @@ def check_and_repair_disk(hutil):
             err_msg = ("check_disk and repair_disk was both specified."
                        "Only one of them can be specified")
             hutil.error(err_msg)
-            hutil.do_exit(1, 'Enable', 'error', '0', 'Enable failed.')
+            return error_code.ErrorCode.CHECK_REPAIR_DISK_ERROR
 
         if check_disk:
             ext_utils.add_extension_event(name=hutil.get_name(), op="scenario", is_success=True, message="check_disk")
             outretcode = _fsck_check(hutil)
-            hutil.log("Successfully checked disk")
```
jscalev, line 567: Why were the trace messages removed?

VMAccess/errorclarification.py
```diff
@@ -0,0 +1,10 @@
+import enum
+
+class ErrorCode(enum.Enum):
+    REPAIR_DISK_ERROR = -2
+    CHECK_DISK_ERROR = -1
+
+    SYSTEM_ERROR = 0
+    
+    USER_PASSWORD_ERROR = 1
+    CHECK_REPAIR_DISK_ERROR = 4
```
jscalev: The naming here is confusing given that we have the system errors REPAIR_DISK_ERROR and CHECK_DISK_ERROR above.

PR #2063, opened 2025-06-11
Add Support for Sending BlobMetadata Information for ADE VM's

arisettisanjana:
Problem:
Currently, restore point metadata for ADE VMs does not include encryption details for associated blobs, this causes problem while restore.

Solution:
(Host based snapshotting)
This change introduces the capability to send BlobMetadata information(encryption details) for Azure Disk Encryption (ADE) virtual machines to Xstore while taking a snapshot of the disks.

To support this, we pass the additional flag isVMADEEnabled as part of the PreSnapshot request to BHS. BHS then computes the blob encryption details and determines the boolean flags isOsDiskADEEncrypted and isDataDiskADEEncrypted, returning them in the response to the extension.

These encryption-related details are included again in the subsequent DoSnapshot call to BHS, which uses them to invoke Xstore with the necessary information for snapshot creation.

This round trip is required because BHS is stateless and does not retain any data in memory between the PreSnapshot and DoSnapshot phases.

(Guest Based Snapshotting)
This change introduces the capability to send BlobMetadata information(encryption details) for Azure Disk Encryption (ADE) virtual machines to Xstore while taking a snapshot of the disks.

To support this, we pass the additional metadata details as part of headers when we make a request to Xstore to take a snapshot


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -590,7 +590,6 @@ def timedelta_total_seconds(self, delta):
     def add_to_telemetery_data(key,value):
         HandlerUtility.telemetry_data[key]=value
 
-    @staticmethod
```
iamashwani-who: This change **removes the `@staticmethod` decorator** from the `get_telemetry_data` method in the `HandlerUtil.py` file.

### What the Change Does
- **Before:**  
  ```python
  @staticmethod
  def get_telemetry_data(key):
      return HandlerUtility.telemetry_data[key]
  ```
- **After:**  
  ```python
  def get_telemetry_data(key):
      return HandlerUtility.telemetry_data[key]
  ```

### Is This Safe?

**Potential Issues:**
- Removing `@staticmethod` means `get_telemetry_data` will now expect a `self` (or `cls`) parameter if called as a method of a class, even though the definition still only specifies `key`.
- **If this method is called as `HandlerUtility.get_telemetry_data(key)`, it will now throw a `TypeError`:**
  ```
  TypeError: get_telemetry_data() takes 1 positional argument but 2 were given
  ```
  This is because Python will pass the class as the first argument, expecting a class method or static method.
- If the intent was to make this an instance method, it should have `self` as the first argument and be called on an instance.
- If it should remain a static method, the decorator **must stay**.

### Conclusion

**This is NOT a safe change** **unless**:
- All calls to `get_telemetry_data` have been updated to call it as a plain function, or
- The method is intended to be an instance method (and all calls are updated accordingly).

**Action:**  
Check all usages of `get_telemetry_data`.  
- If it's always called as `HandlerUtility.get_telemetry_data(key)`, keep `@staticmethod`.
- If it's called on instances, add `self` as the first argument.

**Summary:**  
**Removing `@staticmethod` without other changes will likely break code that calls this as a static method. This is generally NOT a safe change.**

VMBackup/main/freezesnapshotter.py
```diff
@@ -512,3 +524,28 @@ def timedelta_total_seconds(self, delta):
...
 
+    def fetchDiskBlobMetadata(self):
+        headers = self.generate_headers()
+        http_util = HttpUtil(self.logger)
+        for blob in self.para_parser.blobs:
+            sasuri_obj = urlparser.urlparse(blob + '&comp=metadata')
+            result, httpResp, errMsg = http_util.HttpCallGetResponse('GET', sasuri_obj, None, headers = headers)
+            if(result == CommonVariables.success and httpResp != None):
+                resp_headers = httpResp.getheaders()
+                key = "x-ms-meta-DiskEncryptionSettings"
```
iamashwani-who: nit:move this to constants?

VMBackup/main/guestsnapshotter.py
```diff
@@ -94,6 +94,8 @@ def snapshot(self, sasuri, sasuri_index, settings, meta_data, snapshot_result_er
                         key = meta['Key']
                         value = meta['Value']
                         headers["x-ms-meta-" + key] = value
+                if(disk_encryption_details != None):
```
iamashwani-who: let us add validation that value and key are also not Null/Empty

VMBackup/main/guestsnapshotter.py
```diff
@@ -150,9 +152,11 @@ def snapshot_seq(self, sasuri, sasuri_index, settings, meta_data):
                         key = meta['Key']
                         value = meta['Value']
                         headers["x-ms-meta-" + key] = value
+                if(disk_encryption_metadata != None):
```
iamashwani-who: add validation that key and value are not null/empty before setting them

VMBackup/main/guestsnapshotter.py
```diff
@@ -150,9 +152,11 @@ def snapshot_seq(self, sasuri, sasuri_index, settings, meta_data):
...
+                    headers[disk_encryption_metadata[0]] = disk_encryption_metadata[1]
                 if(CommonVariables.isSnapshotTtlEnabled in settings and settings[CommonVariables.isSnapshotTtlEnabled]):
                     self.logger.log("Not passing the TTL header via Guest path though it is enabled")
-                self.logger.log(str(headers))
+                #self.logger.log(str(headers))
```
iamashwani-who: nit:remove commented code.

VMBackup/main/hostsnapshotter.py
```diff
@@ -76,6 +79,14 @@ def snapshotall(self, paras, freezer, g_fsfreeze_on, taskId):
                         temp_dict[CommonVariables.key] = flag
                         temp_dict[CommonVariables.value] = paras.wellKnownSettingFlags[flag]
                         settings.append(temp_dict)
+                '''
+                HostBased changes, hence commenting out
+                if(paras.isVMADEEnabled == True):
+                    settings.append({CommonVariables.key:CommonVariables.isOsDiskADEEncrypted, CommonVariables.value:self.isOsDiskADEEncrypted})
+                    settings.append({CommonVariables.key:CommonVariables.areDataDisksADEEncrypted, CommonVariables.value:self.areDataDisksADEEncrypted})
+                    meta_data.append({CommonVariables.key:CommonVariables.encryptionDetails, CommonVariables.value:self.encryptionDetails})
+                '''
```
iamashwani-who: we can skip adding this for now and include it in next PR to keep the clean set of changes.

VMBackup/main/hostsnapshotter.py
```diff
@@ -149,6 +163,20 @@ def pre_snapshot(self, paras, taskId):
...
+                    elif(responseBody != None):
+                        if (paras.isVMADEEnabled == True):
+                            response = json.loads(responseBody)
+                            self.isOsDiskADEEncrypted = response.get(CommonVariables.isOsDiskADEEncrypted)
+                            self.areDataDisksADEEncrypted = response.get(CommonVariables.areDataDisksADEEncrypted)
+                            self.encryptionDetails = response.get(CommonVariables.encryptionDetails)
+                            self.logger.log("PreSnapshotResponse: isOsDiskADEEncrypted: "+ str(self.isOsDiskADEEncrypted))
+                            self.logger.log("PreSnapshotResponse: areDataDisksADEEncrypted: "+ str(self.areDataDisksADEEncrypted))
+                            if self.encryptionDetails is not None:
+                                self.logger.log("PreSnapshotResponse: encryptionDetails: "+ str(len(self.encryptionDetails)))
```
iamashwani-who, line 181: is it safe to log the whole presnapshot response?

VMBackup/main/hostsnapshotter.py
arisettisanjana: done

VMBackup/main/hostsnapshotter.py
```diff
@@ -156,7 +149,20 @@ def pre_snapshot(self, paras, taskId, fetch_disk_details = False):
...
+                    try:
+                        response_json = json.loads(responseBody)
+                        if "bhsVersion" in response_json:
+                            self.logger.log("PreSnapshotResponse: bhsVersion: " + str(response_json["bhsVersion"]))
+                        if "nodeId" in response_json:
+                            self.logger.log("PreSnapshotResponse: nodeId: " + str(response_json["nodeId"]))
+                        if "responseTime" in response_json:
+                            self.logger.log("PreSnapshotResponse: responseTime: " + str(response_json["responseTime"]))
+                        if "result" in response_json:
+                            self.logger.log("PreSnapshotResponse: result: " + str(response_json["result"]))
```
iamashwani-who, line 162: will result carry encryption details? If so, let us not log result, rather we can log if result was not null

arisettisanjana: The result just would have success or failure, they would not be having the encryption details

iamashwani-who (approved):

prchingh (approved):

iamashwani-who (approved):

PR #2080, opened 2025-07-11
Provide control options for OpenTelemetry metrics input hosts and ports for Metrics Agent and disabling Prometheus input

sajithmuh:


AzureMonitorAgent/services/metrics-extension-cmv2.service
```diff
@@ -3,7 +3,20 @@ Description=Metrics Extension service for Linux Agent metrics sourcing
 After=network.target
 
 [Service]
-ExecStart=%ME_BIN% -TokenSource AMCS -ManagedIdentity %ME_MANAGED_IDENTITY% -DataDirectory %ME_DATA_DIRECTORY% -Input influxdb_local,otlp_grpc,otlp_grpc_prom -InfluxDbSocketPath /var/run/azuremonitoragent/mdm_influxdb.socket -LogLevel Info -Logger Console -OperationEnvironment AMA-Linux/AMA_VERSION
+Environment="OTLP_GRPC_HOST=127.0.0.1"
+Environment="OTLP_GRPC_PORT=4317"
```
GreyCat, line 7: I believe this is the default, so we can skip it?

sajithmuh: Left it there so that cx knows what to change in case they hit the issue. Or is there a risk with it?

simathih (approved):

PR #2082, opened 2025-07-14
Install AMA extension log rotate only if Guest agent log rotate doesn…

simathih:
…'t exist


AzureMonitorAgent/agent.py
```diff
@@ -2044,9 +2044,13 @@ def parse_context(operation):
 
             # As per VM extension team, we have to manage rotation for our extension.log
             # for now, this is our extension code, but to be moved to HUtil library.
-            if not os.path.exists(AMAExtensionLogRotateFilePath):      
-                logrotateFilePath = os.path.join(os.getcwd(), 'azuremonitoragentextension.logrotate')
-                copyfile(logrotateFilePath,AMAExtensionLogRotateFilePath)
+            if os.path.exists(WAGuestAgentLogRotateFilePath):      
+                if os.path.exists(AMAExtensionLogRotateFilePath):
+                    os.remove(AMAExtensionLogRotateFilePath)
```
maddieford: Some customers have already created a directory in their images: '/etc/logrotate.d/azuremonitoragentextension' to avoid this logrotate file from being created. 
os.remove will throw exception on those machines, so this should probably just be a best attempt at removing the file wrapped in a try/catch block

Same for anywhere else you remove this file

PR #2084, opened 2025-07-17
Enable AzureOTelCollector in all regions

sajithmuh:


simathih (approved):

sajithmuh:
@Klaas- This is part of a preview feature and we are still working on the licensing to be part of the package. We have internal license/NOTICE file for this, which we could possibly share with you if you are interested.

PR #2087, opened 2025-07-24
[AMA] Improve the uninstall of AMA by removing files for clean uninstall situation

dylanbun:
This PR removes the files and directories related to Azure Monitor Agent if it is a clean uninstall with the command being `./shim.sh --uninstall --clean`. This does not affect the current uninstall logic in the event of an update.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -574,6 +574,8 @@ def uninstall():
 
     uninstall_azureotelcollector()
 
+    remove_host_configs()
+
     # remove the logrotate config
     if os.path.exists(AMAExtensionLogRotateFilePath):   
         try:
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -42,9 +42,7 @@ import contextlib
 import ama_tst.modules.install.supported_distros as supported_distros
 from collections import OrderedDict
 from hashlib import sha256
-from shutil import copyfile
-from shutil import copytree
-from shutil import rmtree
+from shutil import copyfile, rmtree
 
 from threading import Thread
 import telegraf_utils.telegraf_config_handler as telhandler
@@ -117,8 +115,8 @@ if sys.version_info < (2,7):
 # Global Variables
 PackagesDirectory = 'packages'
 # The BundleFileName values will be replaced by actual values in the release pipeline. See apply_version.sh.
-BundleFileNameDeb = 'azuremonitoragent.deb'
-BundleFileNameRpm = 'azuremonitoragent.rpm'
+BundleFileNameDeb = 'azuremonitoragent_1.35.9-1053_x86_64.deb'
+BundleFileNameRpm = 'azuremonitoragent-1.35.8-1020.x86_64.rpm'
 BundleFileName = ''
 TelegrafBinName = 'telegraf'
 InitialRetrySleepSeconds = 30
@@ -133,7 +131,7 @@ PreviewFeaturesDirectory = '/etc/opt/microsoft/azuremonitoragent/config-cache/pr
 ArcSettingsFile = '/var/opt/azcmagent/localconfig.json'
 AMAAstTransformConfigMarkerPath = '/etc/opt/microsoft/azuremonitoragent/config-cache/agenttransform.marker'
 AMAExtensionLogRotateFilePath = '/etc/logrotate.d/azuremonitoragentextension'
-WAGuestAgentLogRotateFilePath = '/etc/logrotate.d/waagent-extn.logrotate'
+
 SupportedArch = set(['x86_64', 'aarch64'])
 
 # Error codes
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -114,11 +112,14 @@ def __str__(self):
...
 
+AMA_STATE_DIR = '/var/lib/azuremonitoragent'
+AMA_UNINSTALL_CONTEXT_FILE = os.path.join(AMA_STATE_DIR, 'uninstall-context')
+
 # Global Variables
 PackagesDirectory = 'packages'
 # The BundleFileName values will be replaced by actual values in the release pipeline. See apply_version.sh.
-BundleFileNameDeb = 'azuremonitoragent.deb'
-BundleFileNameRpm = 'azuremonitoragent.rpm'
+BundleFileNameDeb = 'azuremonitoragent_1.35.9-1053_x86_64.deb'
```
simathih: this was unintentional rt?

AzureMonitorAgent/agent.py
```diff
@@ -114,6 +112,9 @@ def __str__(self):
     subprocess.check_output = check_output
     subprocess.CalledProcessError = CalledProcessError
 
+AMA_STATE_DIR = '/var/lib/azuremonitoragent'
```
simathih: should it be inside /var/opt/microsoft though?

AzureMonitorAgent/agent.py
```diff
@@ -686,6 +704,126 @@ def force_uninstall_azure_monitor_agent():
...
+        if exit_code != 0 or not output:
+            hutil_log_info("Could not get package file list for cleanup")
+            return []
+
+        # Parse the file list
+        files = [line.strip() for line in output.strip().split('\n') if line.strip()]
+        
+        # Collect all azuremonitor-related paths
+        azuremonitoragent_files = []
+        directory_paths_found = set()
```
simathih: are we using this?

simathih: directory_paths_found i.e.

AzureMonitorAgent/agent.py
```diff
@@ -553,10 +554,20 @@ def uninstall():
     if not is_installed:
         hutil_log_info("Azure Monitor Agent is not installed, nothing to uninstall.")
         return 0, "Azure Monitor Agent is not installed, nothing to uninstall."
-
     if PackageManager != "dpkg" and PackageManager != "rpm":
         log_and_exit("Uninstall", UnsupportedOperatingSystem, "The OS has neither rpm nor dpkg." )
 
+    # Determine uninstall context
+    uninstall_context = _get_uninstall_context()
```
simathih: how are we using the uninstall_context in determining the operation below? it seems like we always clean everything up regardless of the context ...i.e. whether it is update or purge

simathih: do we want to do below only if its "complete"?

dylanbun: uninstall_context is used in `_remove_package_files_from_list()` since we need to know if we want to remove `/var/opt/microsoft/azuremonitoragent` as well or not. For now I do think preserving the folders under this path is desirable in the event of an update while a purge/complete uninstall preserving it makes sense.

AzureMonitorAgent/agent.py
dylanbun: It could work as well, moving it then.

AzureMonitorAgent/agent.py
dylanbun: Good catch, removing.

AzureMonitorAgent/agent.py
simathih: ah got it! please verify these two scenarios:
1. clean install on a VM and delete using portal -> extension tab : ensure /var/opt/microsoft/azuremonitoragent directory is cleared
2. install older version of AMA and kick off upgrade: ensure /var/opt/microsoft logs for mdsd before update are preserved.

simathih: also make sure all the file IO commands are good with both python 2 and 3

simathih (dismissed):

simathih (dismissed):

AzureMonitorAgent/agent.py
```diff
@@ -686,6 +700,110 @@ def force_uninstall_azure_monitor_agent():
...
+        # Determine uninstall context based on if the context file exists
+        uninstall_context = _get_uninstall_context()
+        hutil_log_info("Uninstall context: {0}".format(uninstall_context))
+        
+        if uninstall_context == 'complete':
+            hutil_log_info("Complete uninstall context - removing everything")
+            cleanup_paths.add(AmaDataPath)
+
+        # Sort paths by depth (deepest first) to avoid removing parent before children
+        sorted_paths = sorted(cleanup_paths, key=lambda x: x.count('/'), reverse=True)
```
simathih, line 784: ensure lambda works in both py 2&3

simathih (approved):

NarineM (approved):

PR #2094, opened 2025-08-13
Host based Changes for ADE

arisettisanjana:
Problem:
Currently, restore point metadata for ADE VMs does not include encryption details for associated blobs, this causes problem while restore.

Solution:
(Host based snapshotting)
This change introduces the capability to send BlobMetadata information(encryption details) for Azure Disk Encryption (ADE) virtual machines to Xstore while taking a snapshot of the disks.

To support this, we pass the additional flag isVMADEEnabled as part of the PreSnapshot request to BHS. BHS then computes the blob encryption details and determines the boolean flags isOsDiskADEEncrypted and isDataDiskADEEncrypted, returning them in the response to the extension.

These encryption-related details are included again in the subsequent DoSnapshot call to BHS, which uses them to invoke Xstore with the necessary information for snapshot creation.

This round trip is required because BHS is stateless and does not retain any data in memory between the PreSnapshot and DoSnapshot phases.


VMBackup/main/freezesnapshotter.py
```diff
@@ -432,7 +432,7 @@ def takeSnapshotFromFirstHostThenGuest(self):
         unable_to_sleep = False
         blob_snapshot_info_array = None
         snap_shotter = HostSnapshotter(self.logger, self.hostIp)
-        pre_snapshot_statuscode, responseBody = snap_shotter.pre_snapshot(self.para_parser, self.taskId, False)
+        pre_snapshot_statuscode, responseBody = snap_shotter.pre_snapshot(self.para_parser, self.taskId, True)
```
prchingh, line 435: Are we sending this as true always??
 I dont see a conditional check to see when do we send this to true?

VMBackup/main/common.py
```diff
@@ -68,7 +68,7 @@ class CommonVariables:
     isVMADEEnabled = 'isVMADEEnabled'
     isOsDiskADEEncrypted = 'isOsDiskADEEncrypted'
     areDataDisksADEEncrypted = 'areDataDisksADEEncrypted'
-    encryptionDetails = 'encryptionDetails'
+    encryptionDetails = 'EncryptionDetails'
```
prchingh, line 71: nit: this isn't following the same convention as other variables in this file. Why?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -861,6 +861,18 @@ def convert_to_string(txt):
         else:
             txt = str(txt)
```
prchingh, line 862: pls get some SME review as well for this PR.

VMBackup/main/common.py
arisettisanjana: "EncryptionDetails" string this is the keyValue expected

VMBackup/main/freezesnapshotter.py
arisettisanjana: yes, we make few other pre_snapshot calls as well for those it would be defaulted to false, only for this call this is always true to ensure that we are fetching the blob details once.

prchingh: this presnapshot is only when ADE enabled? where do we have this check?

prchingh (approved):
approving with suggestions. Pls, get a SME review.

VMBackup/main/freezesnapshotter.py
deveshjagwani, line 435: Do we also need to mark this as True in other places where hostsnapshotter.pre_snapshot is called ?

Like in is_command_timedout() which is called from takeSnapshotFromOnlyHost() and is_command_timedout() ?

deveshjagwani: Code was updated in latest commit.

mearvind (approved):

PR #2095, opened 2025-08-13
Retrieves the sequence number for VMAccess from the environment variable

jscalev:
Fixes consistent customer issue where previous handler files are deleted. 

We now default to reading the seqNo from the environment variable, falling back to the previous file method if it's not there.
Also fixes two bugs in the previous file method:
   == None error
   We would fail to retrieve the seqNo if one file did not contain a .

### DIFF
--- a/Utils/handlerutil2.py
+++ b/Utils/handlerutil2.py
@@ -69,6 +69,7 @@ DateTimeFormat = "%Y-%m-%dT%H:%M:%SZ"
 
 MANIFEST_XML = "manifest.xml"
 
+ENV_CONFIG_SEQUENCE_NUMBER = "ConfigSequenceNumber"
 
 class HandlerContext:
     def __init__(self, name):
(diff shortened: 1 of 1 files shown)


viveklingaiah (approved):

Utils/handlerutil2.py
```diff
@@ -131,20 +132,30 @@ def _get_current_seq_no(self, config_folder):
...
-                        if (current_file_m_time > freshest_time):
-                            freshest_time = current_file_m_time
-                            seq_no = cur_seq_no
-                except ValueError:
-                    continue
+
+        # First read the sequence number from the environment variable
+        seq_no_from_env = os.getenv(ENV_CONFIG_SEQUENCE_NUMBER)
+        if (seq_no_from_env is not None):
+            seq_no = int(seq_no_from_env)
```
D1v38om83r: What if the env variable existed, but isn't in the format that can be parsed into int?

Catch ValueError and/or TypeError?

PR #2110, opened 2025-09-23
Removing azureotelcollector installation from  phase

sajithmuh:


AzureMonitorAgent/agent.py
```diff
@@ -593,8 +590,6 @@ def uninstall():
 
     remove_localsyslog_configs()
 
-    uninstall_azureotelcollector()
```
simathih, line 596: why do we need to remove it from here?

sajithmuh: Intent was to move uninstall_azureotelcollector to -disable phase from -uninstall

I believe it is okay to leave it like that because its a package itself and is not activated until MetricsExtension runs - reverting the change

simathih (approved):

PR #2114, opened 2025-10-09
[AMA TST] Improve DCR Parsing and Troubleshooter -A Mode

dylanbun:
Improve the -A mode for the troubleshooter so it does not stop upon first failure. 

Updated the troubleshooter to allow for DCR configs without the channel key (`AgentSettings`) to prevent a DCR being marked as an "error."

Patched how td-agent.conf is parsed and how the endpoint for a connection is validated.

### DIFF
--- a/AzureMonitorAgent/ama_tst/modules/main.py
+++ b/AzureMonitorAgent/ama_tst/modules/main.py
@@ -24,63 +24,65 @@ def check_sudo():
         return True
 
 def check_all(interactive):
-    all_success = NO_ERROR
-    # 1: Install
-    checked_install = check_installation(interactive)
-    if (is_error(checked_install)):
-        return checked_install
-    else:
-        all_success = checked_install
-
+    """
+    Run all troubleshooter checks, continuing even if errors occur.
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/ama_tst/modules/helpers.py
```diff
@@ -220,19 +227,45 @@ def run_cmd_output(cmd):
...
+                            else:
+                                settings_list = settings_str
+                            
+                            # Process each setting
+                            for setting in settings_list:
+                                name = setting['name']
+                                value = setting['value']
+                                if name:
+                                    agent_settings[name] = value
+                        except (json.JSONDecodeError, TypeError) as e:
```
simathih, line 265: did you create a AgentSettings DCR to test this?

dylanbun: Yes I did. The DCR contents just has two keys, kind which correlates to a string to identity DCR type (AgentSettings) and a list of settings referring to MaxDiskQuotaInMB and UseTimeReceivedForForwardedEvents.

PR #2121, opened 2025-10-28
[Instant Access DD Snapshot] Pre Snapshot and Do Snapshot API Changes

deveshjagwani:
Task 32983493, 35720465

### DIFF
--- a/VMBackup/main/hostsnapshotter.py
+++ b/VMBackup/main/hostsnapshotter.py
@@ -236,7 +236,9 @@ class HostSnapshotter(object):
                         if 'instantAccessDurationMinutes' in snapshot_info['ddSnapshotIdentifier']:
                             instantAccessDurationMinutes = snapshot_info['ddSnapshotIdentifier']['instantAccessDurationMinutes']
                         ddSnapshotIdentifierInfo = HostSnapshotObjects.DDSnapshotIdentifier(creationTimeUTCTicks , snapshot_info['ddSnapshotIdentifier']['id'], snapshot_info['ddSnapshotIdentifier']['token'], instantAccessDurationMinutes)
-                        self.logger.log("ddSnapshotIdentifier Information from Host- creationTime : {0}, id : {1}".format(ddSnapshotIdentifierInfo.creationTime, ddSnapshotIdentifierInfo.id))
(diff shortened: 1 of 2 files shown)


VMBackup/main/hostsnapshotter.py
prchingh, line 1: Do we need any changes in processing the response from BHS, generally in Linux extension code what do we follow??

prchingh (approved):
@arisettisanjana   Pls, review these PRs as well for IA related feature code in extensions

VMBackup/main/hostsnapshotter.py
```diff
@@ -231,8 +231,11 @@ def get_snapshot_info(self, responseBody):
                         self.logger.log("Converting the creationTime string received in UTC format to UTC Ticks")
                         delta = creationTimeObj - epochTime
                         timestamp = self.get_total_seconds(delta)
-                        creationTimeUTCTicks = str(int(timestamp * 1000)) 
-                        ddSnapshotIdentifierInfo = HostSnapshotObjects.DDSnapshotIdentifier(creationTimeUTCTicks , snapshot_info['ddSnapshotIdentifier']['id'], snapshot_info['ddSnapshotIdentifier']['token'])
+                        creationTimeUTCTicks = str(int(timestamp * 1000))
+                        instantAccessDurationMinutes = None
+                        if 'instantAccessDurationMinutes' in snapshot_info['ddSnapshotIdentifier']:
```
arisettisanjana, line 236: can we use the constant instantAccessDurationMinutes defined in common.py in lines 236 and 237 also while accessing the property

arisettisanjana (approved):

VMBackup/main/hostsnapshotter.py
deveshjagwani: no changes needed here. the response received is basically extracted based on the fields that are expected and snapshot info is prepared using get_snapshot_info() and final extension run status is populated using update_snapshotinfoarray().

VMBackup/main/hostsnapshotter.py
deveshjagwani: correct. I will update later because this IA string is related to response actually and there are many other strings like ddsnapshotidentifier, creationtime, etc. which needs similar changes. will take this up in future while writing UTs and Refactoring.

PR #2122, opened 2025-11-03
Decrypt Protected Settings in VMSnapshot Extension for FIPS 140-3 AES256 support and backward compatibility for DES_EDE3_CBC encryption

deveshjagwani:
FedRamp requirement security task: 34614091

### DIFF
--- a/VMBackup/main/Utils/HandlerUtil.py
+++ b/VMBackup/main/Utils/HandlerUtil.py
@@ -243,7 +243,7 @@ class HandlerUtility:
         try:
             cms_cmd = base64_cmd + " | " + self.patching.openssl_path + " cms -inform DER -decrypt -recip " + cert_path + " -inkey " + pkey_path
             self.log("Attempting decryption using OpenSSL CMS command (supports AES256 and DES_EDE3_CBC)")
-            result = waagent.RunGetOutput(cms_cmd)
+            result = waagent.RunGetOutput(cms_cmd, chk_err=False, log_cmd=False)
             if result[0] == 0 and result[1]:  # Success (return code 0) and non-empty output
                 cleartxt = result[1]
                 self.log("Successfully decrypted protected settings using CMS command")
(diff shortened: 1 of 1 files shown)


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -205,6 +205,67 @@ def error(self, message):
...
+                self.log("Successfully decrypted protected settings using SMIME command (fallback)")
+                return cleartxt
+            else:
+                self.error("SMIME decryption also failed with return code: " + str(result[0]))
+        except Exception as e:
+            self.error("SMIME decryption failed with exception: " + str(e))
+        
+        # If both methods fail, raise an error
+        if not cleartxt:
+            self.error("Failed to decrypt protected settings using both CMS and SMIME commands")
```
ppusuluri-ms, line 272: Earlier we were not raising errors, would this cause regressions?

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -205,6 +205,67 @@ def error(self, message):
...
+                cleartxt = result[1]
+                self.log("Successfully decrypted protected settings using CMS command")
+                return cleartxt
+            else:
+                self.log("CMS decryption failed with return code: " + str(result[0]) + ", attempting fallback to SMIME")
+        except Exception as e:
+            self.log("CMS decryption failed with exception: " + str(e) + ", attempting fallback to SMIME")
+        
+        # Fallback to OpenSSL SMIME command (supports DES_EDE3_CBC only)
+        try:
```
ppusuluri-ms, line 257: what is the testing strategy for this usecase?

deveshjagwani: Deliberately fail the CMS command (just by throwing exception) for both VM cases: opted in and opted out and then see the expected results.

Failure can be added locally and then tested in Linux VM.

VMBackup/main/Utils/HandlerUtil.py
deveshjagwani: earlier there was if-else case based on base 64 decode option only. If the smime command would have failed - then error would have been thrown anyhow I think. We can remove this if we want either way. Just because this is added, doesn't mean it would regress, I think logically we're also reporting error in try-except block, so will see if this is needed or not.

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -205,6 +205,67 @@ def error(self, message):
...
+            cms_cmd = base64_cmd + " | " + self.patching.openssl_path + " cms -inform DER -decrypt -recip " + cert_path + " -inkey " + pkey_path
+            self.log("Attempting decryption using OpenSSL CMS command (supports AES256 and DES_EDE3_CBC)")
+            result = waagent.RunGetOutput(cms_cmd)
+            if result[0] == 0 and result[1]:  # Success (return code 0) and non-empty output
+                cleartxt = result[1]
+                self.log("Successfully decrypted protected settings using CMS command")
+                return cleartxt
+            else:
+                self.log("CMS decryption failed with return code: " + str(result[0]) + ", attempting fallback to SMIME")
+        except Exception as e:
```
arisettisanjana, line 253: is there any failure scenario logs like the decryption via the cms fails in that case we are logging the error and falling back to the other approach, but does the error have any details related to key or cert as this data would be pushed to kusto and it should not be having any sensitive info

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -205,6 +205,67 @@ def error(self, message):
     def fetch_log_message(self):
         return self.log_message
```
arisettisanjana, line 214: Since protected settings are used in guest path, please do test the guest path

deveshjagwani: done. results are in the corresponding work item: 35782489

VMBackup/main/Utils/HandlerUtil.py
deveshjagwani: sensitive info check also done. nothing logged. recent testing results in work item: 35782489

PR #2124, opened 2025-11-06
Updates for deprecated modules in Python 3.12 and Invalid escape sequence issues

deveshjagwani:
****ADO Task:** 34492797**

Fix Python 3.12 compatibility issues

- Replace distutils.version with packaging/distutils/custom fallback in WaagentLib.py
- Replace imp module with importlib.util fallback in WAAgentUtil.py
- Fix invalid escape sequences in HandlerUtil.py JSON processing
- Add setuptools fallback for distutils removal in setup.py

Resolves VMBackup extension failure on Debian 12 with Python 3.12. All changes maintain backward compatibility with older Python versions.

Fix platform.dist() deprecation error for Python 3.8+

- Added conditional import for distro package
- Implemented multi-tier fallback strategy in get_dist_info()
- Enhanced error handling with proper logging
- Maintains backward compatibility across all Python versions
- Resolves AttributeError: module 'platform' has no attribute 'dist'

Fix Python 3.13+ compatibility: Handle deprecated crypt module import

- Add try/except block for crypt import to handle Python 3.13+ where crypt module was removed
- Update gen_password_hash() to gracefully handle missing crypt module 
- Add TODO comment for future cleanup of unused code in VMBackup extension
- Maintains backward compatibility with older Python versions
- Resolves import error while preserving all existing functionality

### DIFF
--- a/VMBackup/main/WaagentLib.py
+++ b/VMBackup/main/WaagentLib.py
@@ -23,7 +23,16 @@
 # http://msdn.microsoft.com/en-us/library/cc227259%28PROT.13%29.aspx
 #
 
-import crypt
+# TODO: Many classes, methods, and imports in this file might not be needed by VM Backup Extension
+# and should be removed to reduce file size and eliminate unnecessary dependencies.
+# Future cleanup should analyze actual VMBackup usage and remove unused code.
+
+# Note: crypt module deprecated in Python 3.13+, but gen_password_hash() is not used by VMBackup
+try:
+    import crypt
+except ImportError:
+    # Python 3.13+ removed crypt module, but VMBackup doesn't use password functions
+    crypt = None
 import random
 import base64
 
@@ -463,6 +472,8 @@ class AbstractDistro(object):
             return "Failed to set password for {0}: {1}".format(username, output)
 
     def gen_password_hash(self, password, crypt_id, salt_len):
+        if crypt is None:
+            raise ImportError("crypt module not available (Python 3.13+). This function is not used by VMBackup.")
         collection = string.ascii_letters + string.digits
         salt = ''.join(random.choice(collection) for _ in range(salt_len))
         salt = "${0}${1}".format(crypt_id, salt)


VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -552,28 +559,74 @@ def get_dist_info(self):
...
+            if hasattr(platform, 'linux_distribution'):
+                try:
+                    distinfo = list(platform.linux_distribution(full_distribution_name=0))
+                    # remove trailing whitespace in distro name
+                    if(distinfo[0] == ''):
+                        osfile= open("/etc/os-release", "r")
+                        for line in osfile:
+                            lists=str(line).split("=")
+                            if(lists[0]== "NAME"):
+                                distroname = lists[1].split("\"")
```
arisettisanjana, line 583: lines 579 to 587 is same as 605 to 614, can we not have this only 1time, why do we need the last try block

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -655,9 +708,18 @@ def do_status_report(self, operation, status, status_code, message, taskId = Non
...
         stat_rept = "[" + json.dumps(stat_rept, cls = ComplexEncoder) + "]"
-        stat_rept = stat_rept.replace('\\\/', '\/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
+        # Replace placeholder GUID with actual Microsoft JSON date format first
+        # Before: "timestampUTC":"e2794170-c93d-4178-a8da-9bc7fd91ecc0"
+        # After: "timestampUTC":"\/Date(time_span)\/"
         stat_rept = stat_rept.replace(date_place_holder,date_string)
+        # Now remove JSON-escaped forward slashes to get clean date format for C# DateTimeOffset
+        # Before: "timestampUTC":"\/Date(time_span)\/"
+        # After: "timestampUTC":"/Date(time_span)/"
+        stat_rept = stat_rept.replace(r'\/', '/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
```
arisettisanjana, line 722: Is this been tested with older python versions as well.

VMBackup/main/Utils/HandlerUtil.py
deveshjagwani: yes key value parsing seems to be the same in both cases but for the second case its a bit different after that. I'll try to refactor later during another PR when introducing Unit Tests, and for now keeping it this way as it works during manual testing.

Try catch is always good to have for log any exception in any line.

VMBackup/main/Utils/HandlerUtil.py
deveshjagwani: older python versions basically gave the warning for this case saying that it would fail in future but didn't actually fail. In new python version it did fail as per the previous warning. Currently I checked for newer python versions only for in ubuntu 25 and debian 13 vms, but it will work for older versions as well since I tried to write a sample UT case locally (not here in code) and checked that the replace worked.

We'll also test this in canary first during deployment so any other anomalies can be caught during that time.

deveshjagwani:
Extension success status result logs are present in the work item in PR description (ADO: 34492797). Also the rpc names are there @arisettisanjana

arisettisanjana:
> Extension success status result logs are present in the work item in PR description (ADO: 34492797). Also the rpc names are there @arisettisanjana

The tests were run to verify on py3.13 versions so, just wanted to be sure that we have tested these changes for backward compatibility i.e with lesser python versions

PR #2125, opened 2025-11-07
Decode output string only if not a string already

simathih:


AzureMonitorAgent/agent.py
```diff
@@ -3057,17 +3057,9 @@ def run_get_output(cmd, chk_err = False, log_cmd = True):
...
-        unicode_type = str # Python 3
-
-    if all(ord(c) < 128 for c in output) or isinstance(output, unicode_type):
-        output = output.encode('utf-8')
-
-    # On python 3, encode returns a byte object, so we must decode back to a string
-    if sys.version_info >= (3,) and type(output) == bytes:
-        output = output.decode('utf-8', 'ignore')
+    # On python 3, check_output returns a binary object
+    if not isinstance(output, str):
```
AaronRM: In Python2, this could hit an error if output is type unicode. In that case, "not isinstance(output, str)" returns True, but since there is no 'decode' method on a unicode object, we would get "AttributeError: 'unicode' object has no attribute 'decode'"

Copilot is suggesting this as a safer alternative change that will retain python2 support:

``` python
    # Python 2: encode unicode -> UTF-8 bytes (str). Python 3: decode bytes -> str.
    try:  # Python 2
        if isinstance(output, unicode):  # type: ignore  # noqa: F821
            output = output.encode('utf-8', 'ignore')
    except NameError:  # Python 3
        if isinstance(output, (bytes, bytearray)):
            output = bytes(output).decode('utf-8', 'ignore')
```

simathih: fixed it

PR #2128, opened 2025-11-13
avoid throwing exception on decrypt failure

thperapp:


Utils/HandlerUtil.py
```diff
@@ -201,24 +201,19 @@ def _parse_config(self, ctxt):
...
                     try:
-                        session = subprocess.Popen([decrypt_cmd], shell=True,
-                                                stdin=subprocess.PIPE,
-                                                stderr=subprocess.STDOUT,
-                                                stdout=subprocess.PIPE)
-                        output = session.communicate(unencodedSettings)
-                        # success only if return code is 0 and we have output
-                        if session.returncode == 0 and output[0]:
-                            protected_settings_str = output[0]
+                        output = waagent.RunSendStdin(decrypt_cmd, unencodedSettings)
```
simathih, line 208: does this reference this file: https://github.com/Azure/azure-linux-extensions/blob/1169f53640b1ae539f7875bf5ebaaf57255e6c78/Utils/WAAgentUtil.py

thperapp: https://github.com/Azure/azure-linux-extensions/blob/1169f53640b1ae539f7875bf5ebaaf57255e6c78/Common/WALinuxAgent-2.0.16/waagent#L2615C1-L2615C74

thperapp: it should be this one

thperapp: It's the function call we were using before my FIPS changes

simathih (approved):

Utils/HandlerUtil.py
```diff
@@ -201,24 +201,19 @@ def _parse_config(self, ctxt):
...
+                        if output[0] == 0 and output[1]:
+                            protected_settings_str = output[1]
                             if decrypt_cmd == cms_cmd:
                                 self.log('Decrypted protectedSettings using openssl cms.')
                             else:
                                 self.log('Decrypted protectedSettings using openssl smime fallback.')
                             break
                         else:
-                            self.log('Attempt to decrypt protectedSettings with "{0}" failed (rc={1}).'.format(decrypt_cmd, session.returncode))
+                            self.log('Attempt to decrypt protectedSettings with "{0}" failed (rc={1}).'.format(decrypt_cmd, output[0]))
```
NarineM: this will be hit even when output[0] is null,and when it is 0 since you are checking on output[1] against null , should not it also be checked? what happens in those cases? i think it would still be exception that you want to catch/avoid

thperapp: @NarineM, linking the func we're calling - https://github.com/Azure/azure-linux-extensions/blob/1169f53640b1ae539f7875bf5ebaaf57255e6c78/Common/WALinuxAgent-2.0.16/waagent#L2615C1-L2615C74

Both of the scenarios wouldn't end up making it past the function call since it is being handled in RunSendStdin()

Utils/HandlerUtil.py
```diff
@@ -201,24 +201,19 @@ def _parse_config(self, ctxt):
...
-                        session = subprocess.Popen([decrypt_cmd], shell=True,
-                                                stdin=subprocess.PIPE,
-                                                stderr=subprocess.STDOUT,
-                                                stdout=subprocess.PIPE)
-                        output = session.communicate(unencodedSettings)
-                        # success only if return code is 0 and we have output
-                        if session.returncode == 0 and output[0]:
-                            protected_settings_str = output[0]
+                        output = waagent.RunSendStdin(decrypt_cmd, unencodedSettings)
+                        if output[0] == 0 and output[1]:
```
NarineM: can you please add a comment why output[0] and output[1] are special combinations and what their existance means?

PR #2156, opened 2026-03-30
Add fallbacks for import crypt

norakoiralamsft:
This PR is aimed to add fallbacks for importing the cypt library with was deprecated in python 3.11, takes a similar approach as to Linux Agent, first trying to import crypt (which includes crypt  + crypt_r), legacycrypt, and finally passlib. 

This was chosen over the crypt_fallback that was introduced in this [PR ](https://github.com/Azure/azure-linux-extensions/pull/2152) because we want to continue supporting older distros that may not have cyptes available, and Canocial raised concerns about the crypt_r lib as well, and have ensured that passlib will be available on Ubuntu.

### DIFF
--- a/Utils/distroutils.py
+++ b/Utils/distroutils.py
@@ -4,33 +4,39 @@ import random
 import string
 import hashlib
 import sys
-
-# crypt module was removed in Python 3.13
-# For Python < 3.11: use builtin crypt
-# For Python >= 3.11: try crypt_r package, then ctypes fallback
-if sys.version_info >= (3, 11):
-    try:
-        import crypt_r as crypt
-    except ImportError:
-        try:
-            from Utils import crypt_fallback as crypt
-        except ImportError:
-            crypt = None
-else:
-    try:
-        import crypt
-    except ImportError:
-        try:
-            from Utils import crypt_fallback as crypt
-        except ImportError:
-            crypt = None
-
 import platform
 import re
 import Utils.logger as logger
 import Utils.extensionutils as ext_utils
 import Utils.constants as constants
 
+# crypt module was removed in Python 3.13
+# first we try to import the crypt module which covers crypt and crypt_r
+# then we try legacyrypt and finally fallback to passlib 
+
+cryptImported = False
+passLibImported = False
+
+try:
+    from crypt import crypt as crypt
+    cryptImported = True
+except ImportError:
+    pass 
+
+if cryptImported == False: 
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 13 or sys.version_info[0] > 3:
+        try: 
+            from legacycrypt import crypt
+            cryptImported = True
+        except ImportError:
+            pass 
+
+if cryptImported == False: 
+    try: 
+        from passlib.hash import sha512_crypt
+        passLibImported = True
+    except ImportError:
+        pass 
 
 def get_my_distro(config, os_name=None):
     if 'FreeBSD' in platform.system():
@@ -174,7 +180,13 @@ class GenericDistro(object):
         collection = string.ascii_letters + string.digits
         salt = ''.join(random.choice(collection) for _ in range(salt_len))
         salt = "${0}${1}".format(crypt_id, salt)
-        return crypt.crypt(password, salt)
+
+        if cryptImported:
+            return crypt.crypt(password, salt)
+        elif passLibImported:
+            return sha512_crypt.hash(password)
+        else:
+            logger.error("This feature requires one of the 'crypt', 'legacycrypt', 'crypt-r', or passlib Python packages to be installed.")
 
     def create_account(self, user, password, expiration, thumbprint, enable_nopasswd):
         """
--- a/Utils/distroutils.py
+++ b/Utils/distroutils.py
@@ -171,6 +171,9 @@ class GenericDistro(object):
 
     def chpasswd(self, username, password, crypt_id=6, salt_len=10):
         passwd_hash = self.gen_password_hash(password, crypt_id, salt_len)
+        if passwd_hash is None:
+            return "This feature requires one of the 'crypt', 'legacycrypt', 'crypt-r', or 'passlib' Python packages to be installed."
+
         cmd = ['usermod', '-p', passwd_hash, username]
         ret, output = ext_utils.run_command_get_output(cmd, log_cmd=False)
         if ret != 0:
@@ -182,11 +185,11 @@ class GenericDistro(object):
         salt = "${0}${1}".format(crypt_id, salt)
 
         if cryptImported:
-            return crypt.crypt(password, salt)
+            return crypt(password, salt)
         elif passLibImported:
             return sha512_crypt.hash(password)
         else:
-            logger.error("This feature requires one of the 'crypt', 'legacycrypt', 'crypt-r', or passlib Python packages to be installed.")
+            return None
 
     def create_account(self, user, password, expiration, thumbprint, enable_nopasswd):
         """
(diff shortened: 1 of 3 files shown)


jscalev (dismissed):

Utils/distroutils.py
```diff
@@ -4,33 +4,39 @@
...
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 13 or sys.version_info[0] > 3:
+        try: 
+            from legacycrypt import crypt
+            cryptImported = True
+        except ImportError:
+            pass 
+
+if cryptImported == False: 
+    try: 
+        from passlib.hash import sha512_crypt
```
alsanmsft, line 36: question for my understanding: why do we not try to import passlib right away after crypt fails? why is it our second fallback, not our first? what benefits does legacycrypt have?

Utils/distroutils.py
```diff
@@ -174,7 +183,13 @@ def gen_password_hash(self, password, crypt_id, salt_len):
         collection = string.ascii_letters + string.digits
         salt = ''.join(random.choice(collection) for _ in range(salt_len))
         salt = "${0}${1}".format(crypt_id, salt)
-        return crypt.crypt(password, salt)
+
+        if cryptImported:
+            return crypt(password, salt)
```
alsanmsft, line 188: maybe a dumb question, but is crypt() a different or the same algorithm as sha512_crypt()? why do we prefer to use one over the other?

Utils/distroutils.py
narrieta: The order doesn't really matter.

In terms of functionality, they are equivalent. Currently they are ordered in terms of which option may be more frequent in the fleet, but that will change over time and the perf difference is negligible, so order does not matter.

Utils/distroutils.py
narrieta: crypt with crypt_id==6 is equivalent to sha512_crypt

which one is used depends on which module was successfully imported

Utils/distroutils.py
narrieta: The changes in this PR are modelled on https://github.com/Azure/WALinuxAgent/pull/3542, btw

malhotrarohit (commented):
test comment

Utils/distroutils.py
```diff
@@ -4,33 +4,39 @@
...
+passLibImported = False
+
+try:
+    from crypt import crypt as crypt
+    cryptImported = True
+except ImportError:
+    pass 
+
+if cryptImported == False: 
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 13 or sys.version_info[0] > 3:
```
malhotrarohit, line 27: would be good to explain what these numbers mean or rather what this expression means

malhotrarohit (dismissed):

Utils/distroutils.py
```diff
@@ -174,7 +183,13 @@ def gen_password_hash(self, password, crypt_id, salt_len):
         collection = string.ascii_letters + string.digits
         salt = ''.join(random.choice(collection) for _ in range(salt_len))
         salt = "${0}${1}".format(crypt_id, salt)
-        return crypt.crypt(password, salt)
+
+        if cryptImported:
+            return crypt(password, salt)
+        elif passLibImported:
```
birukyitbarek, line 189: This shouldn't ignore crypt_id and salt. This could result in a bug depending on n distro configuration?

Utils/distroutils.py
viveklingaiah, line 27: add braces for sys.version_info[0] == 3 and sys.version_info[1] >= 13 ? (for easier readability)

Utils/distroutils.py
```diff
@@ -4,33 +4,39 @@
...
+            cryptImported = True
+        except ImportError:
+            pass 
+
+if cryptImported == False: 
+    try: 
+        from passlib.hash import sha512_crypt
+        passLibImported = True
+    except ImportError:
+        pass
```
viveklingaiah, line 39: If all 3 imports fail, do we need to throw an error here + log ?

Utils/distroutils.py
dylanbun: +1, I think it would be a good idea to get the hashes for all scenarios/libraries first and confirming they all match prior to merging. (i.e. assuming they are using `sha512_crypt` and passing in salt).

Utils/distroutils.py
narrieta: This condition (all 3 imports failing) needs to be handled, but raising an exception here would require all modules depending on this module, directly or indirectly, to handle that exception, or the extension would crash.

@norakoiralamsft - in the original PR, an exception is raised on usage of these functions if no module was imported. I see that you are handling this at line 192 by returning None. Do verify that this error handling is appropriate for your usage.

Utils/distroutils.py
narrieta: The crypt_id defines the hashing algorithm to use and is included as part of the salt parameter in the call to crypt.crypt  (notice that it is prepended to the salt). For some reason, the WALinuxAgent allowed users to set the crypt_id in waagent.conf; I am assuming that VmAccess is using legacy code that reads this value from waagent.conf.

@norakoiralamsft - For WALinuxAgent, we decided to make this change (see the README in the develop branch, marked with ******* below):

```
Provisioning.PasswordCryptId
Type: String
Default: 6

Algorithm used by crypt when generating password hash. 

******This parameter, as well as PasswordCryptSaltLength, is used only on Python versions previous to 3.13, and SHA-512 is the default for most distros. From Python 3.13 onwards, SHA-512 is always used, regardless of the value of this parameter.*******

1 - MD5
2a - Blowfish
5 - SHA-256
6 - SHA-512

```
I think that would be appropriate for VmAccess as well, but pointing it out in case you want to follow a different approach.

narrieta: The call to passlib.hash() is not given a salt. One will be generated automatically and that is the recommended usage.

Utils/test/test_distroutils_password_hash.py
```diff
@@ -0,0 +1,173 @@
...
+        password = "TestP@ssw0rd!"
+        hash_val = distro.gen_password_hash(password, crypt_id=6, salt_len=10)
+
+        # A valid crypt hash starts with the salt prefix
+        self.assertTrue(hash_val.startswith("$6$"), "Expected SHA-512 hash prefix '$6$'")
+        # Verify: re-hashing with the produced hash as the salt must equal the hash
+        self.assertEqual(self._crypt(password, hash_val), hash_val,
+                         "Hash does not verify against the original password")
+
+    def test_different_passwords_produce_different_hashes(self):
```
narrieta, line 72: A more important test is that the same password produces different hashes (i.e. it is being salted). 

Similar for the passlib test.

Utils/distroutils.py
norakoiralamsft: Decided to throw exception in the gen_passwsord_hash method instead of returning None and floating up the error message.

Utils/distroutils.py
norakoiralamsft: +1 to everything Norberto mentioned, I've added comments to clarify the salt + crypt_id usage for the different hashing methods. Opting to not pass in salt for sha512_crypt(), as a random salt will be generated on our behalf and for crypt() we generate random salt 

I've also added unit tests to verifying hashing is working as expected.

Utils/test/test_distroutils_password_hash.py
norakoiralamsft: Added

narrieta: thanks

Utils/distroutils.py
narrieta: @norakoiralamsft - thanks; lgtm

viveklingaiah (approved):

PR #2158, opened 2026-03-31
Fix incorrect configurationId when multiple DCRs share the same counters for ME metrics

simathih:
When multiple DCRs configured the same performance counters, all metrics were sent to ME tagged with only the **first** DCR's `configurationId` (due to an early `break`). ME uses this tag to route data per-DCR, so all subsequent DCRs received incorrectly attributed metrics.

## Changes

- **`parse_config` in `telegraf_config_handler.py`**
  - Collect all unique `configurationId`s across all fields in a plugin before emitting any config
  - For each `configId`, emit a **separate telegraf config file** named `<omiclass>-<configId>.conf` (previously one file per plugin, e.g. `filesystem.conf`)
  - Each generated file carries only its own `configurationId` in the `[inputs.<plugin>.tags]` block

**Before** (one file per plugin, wrong configId for DCR 2+):
```
filesystem.conf          # configurationId="dcr-fa9e2b03..." for all DCRs
```

**After** (one file per plugin per DCR, correct configId in each):
```
filesystem-dcr-c93b5596....conf   # configurationId="dcr-c93b5596..."
filesystem-dcr-fa9e2b03....conf   # configurationId="dcr-fa9e2b03..."
```

after
before

### DIFF
--- a/LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
+++ b/LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
@@ -358,12 +358,23 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
                 telegraf_plugin = splitResult[0]
 
             if not is_lad:
-                configIds = counterConfigIdMap[telegraf_json[omiclass][plugin][field]["displayName"]]
-                for configId in configIds:
-                    input_str += "\n"
-                    input_str += " "*2 + "[inputs." + telegraf_plugin + ".tags]\n"
-                    input_str += " "*4 + "configurationId=\"" + configId + "\"\n\n"
-                    break
+                # Collect all unique configurationIds from all fields in this plugin
+                all_config_ids = []
+                for f in telegraf_json[omiclass][plugin]:
+                    display_name = telegraf_json[omiclass][plugin][f]["displayName"]
+                    if display_name in counterConfigIdMap:
+                        for cid in counterConfigIdMap[display_name]:
+                            if cid not in all_config_ids:
+                                all_config_ids.append(cid)
+                if all_config_ids:
+                    base_str = input_str
+                    curr_input_str = ""
+                    for cid in all_config_ids:
+                        curr_input_str = base_str
+                        curr_input_str += "\n"
+                        curr_input_str += " "*2 + "[inputs." + telegraf_plugin + ".tags]\n"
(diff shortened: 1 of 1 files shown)


LAD-AMA-Common/telegraf_utils/telegraf_config_handler.py
```diff
@@ -358,12 +358,23 @@ def parse_config(data, me_url, mdsd_url, is_lad, az_resource_id, subscription_id
                 telegraf_plugin = splitResult[0]
 
             if not is_lad:
-                configIds = counterConfigIdMap[telegraf_json[omiclass][plugin][field]["displayName"]]
-                for configId in configIds:
-                    input_str += "\n"
-                    input_str += " "*2 + "[inputs." + telegraf_plugin + ".tags]\n"
-                    input_str += " "*4 + "configurationId=\"" + configId + "\"\n\n"
-                    break
+                # Collect all unique configurationIds from all fields in this plugin
```
hestolz: Can you show the before-after of the result conf? That will help illustrate the fix.

simathih: when we have multiple DCRs with the same counters, we need to send the config id for ME to figure what to do with the data for each DCR. so config like this:

[[inputs.disk]]
  fieldpass = ["inodes_free_percent", "free_percent", "inodes_used_percent", "used_percent", "free"]
  interval = "30s"


  [inputs.disk.tags]
    configurationId="dcr-fa9e2b0327fa46d78bd343a9224cf6e2"


becomes

file 1:

[[inputs.disk]]
  fieldpass = ["inodes_free_percent", "free_percent", "inodes_used_percent", "used_percent", "free"]
  interval = "30s"


  [inputs.disk.tags]
    configurationId="dcr-fa9e2b0327fa46d78bd343a9224cf6e2"

file 2:
[[inputs.disk]]
  fieldpass = ["inodes_free_percent", "free_percent", "inodes_used_percent", "used_percent", "free"]
  interval = "30s"


  [inputs.disk.tags]
    configurationId="dcr-c93b559600e14b169592d3cf34b689ca"

simathih:
it will look like this now: 
image

from:
image

simathih:
@copilot please review this

AaronRM (approved):
This change will generate an `[[inputs.plugin]]` config per DCR (potentially resulting in telegraf collecting duplicate metrics per collection interval if they are duplicated across DCRs). This appears to be unavoidable with Telegraf. Just calling it out for awareness.

PR #2160, opened 2026-04-20
Remove getcwd to load WAAgent properly and not fall to backup

gopaljain00:
This PR is a fix for this issue: https://msazure.visualstudio.com/One/_workitems/edit/36801896
(This WI has response from GuestAgent team which provides insights on the possible errors.)

VMBackup uses a library called **WAAgent** (https://github.com/Azure/WALinuxAgent)
This library comes bundled with the VMBackup extension (present in **VMBackup\main\WaagentLib.py**), and is loaded by a utility file (**VMBackup/main/Utils/WAAgentUtil.py**)

WAAgentUtil.py loads WAAgent in 2 steps.
Step-1: Try to load the bundled file by constructing its path -
```agentPath = os.path.join(os.getcwd(), "main/WaagentLib.py")```

Step-2: If above path is not resolved correctly, check if this library is present in the user's VM in the directory:
```agentPath = '/usr/sbin/waagent'```

Ideally, Step-1 should work, and we should not need to reach fallback (i.e Step-2).

The issue here is the path resolution in Step-1 is fragile. It will only work when executed from extension root (as it uses getcwd()). Which might result in loading the fallback.
And it is not guaranteed that the fallback works as expected with our extension as intended as it is coming from User's VM so there can be version mismatch, etc (resulting in missing dependencies such as ShellUtil not defined).

**In this PR, I've added changes to**:
1. Solidify the path resolution from the Util file to the bundled WAAgent library - This should fix the shellUtil errors

Some additional changes done:
1. Similar issue was present for loading handle.sh from handle.py - This should improve the errors related to the extension not being initialized successfully.
2. We are loading the environment variable PYTHONPATH. If the User's VM does not have this variable set, it will throw an error, therefore added a fallback value of empty string.

### DIFF
--- a/VMBackup/main/Utils/WAAgentUtil.py
+++ b/VMBackup/main/Utils/WAAgentUtil.py
@@ -41,10 +41,10 @@ import os.path
 # it as a submodule of current module
 #
 def searchWAAgent():
-    agentPath = os.path.join(os.getcwd(), "main/WaagentLib.py")
+    agentPath = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir, "WaagentLib.py")
     if(os.path.isfile(agentPath)):
         return agentPath
-    user_paths = os.environ['PYTHONPATH'].split(os.pathsep)
+    user_paths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
     for user_path in user_paths:
         agentPath = os.path.join(user_path, 'waagent')
         if(os.path.isfile(agentPath)):
@@ -55,7 +55,7 @@ def searchWAAgentOld():
     agentPath = '/usr/sbin/waagent'
     if(os.path.isfile(agentPath)):
         return agentPath
-    user_paths = os.environ['PYTHONPATH'].split(os.pathsep)
+    user_paths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
     for user_path in user_paths:
         agentPath = os.path.join(user_path, 'waagent')
         if(os.path.isfile(agentPath)):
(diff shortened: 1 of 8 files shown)


gopaljain00:
> @gopaljain00 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="{Microsoft}"]

VMBackup/main/Utils/WAAgentUtil.py
OMJayswal, line 76: This line broke in the last release which was reverted as a part of this rollback.
I suggest you to either checkout from main and then merge your changes(might need conflict resolution) or removed what's undone in this [PR  ](https://github.com/Azure/azure-linux-extensions/pull/2163/) 
Also, remove files with whitespace changes only from the PR.

gopaljain00: Got it. I will see if we can reproduce the same (+ fix) in UTs.

PR #2162, opened 2026-04-20
Fix proxy.mode=none not honored when https_proxy is set

thperapp:
When proxy.mode is set to 'none' in AMA public settings, the extension should set MDSD_PROXY_MODE=none so that mdsd knows to skip all proxy configuration, including system environment variables (https_proxy, etc).

Previously, handle_mcs_config() only checked for mode=='application'. When mode=='none', it fell through without setting MDSD_PROXY_MODE, causing mdsd to pick up the system https_proxy environment variable.

Add an explicit check for proxy.mode=='none' before the 'application' check that sets MDSD_PROXY_MODE=none, calls unset_proxy() to remove systemd proxy overrides, and marks proxySet=True to prevent Arc proxy fallback.

Fixes ICM 21000000985043

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -1097,7 +1097,11 @@ def handle_mcs_config(public_settings, protected_settings, default_configs):
     proxySet = False
 
     # fetch proxy settings
-    if public_settings is not None and "proxy" in public_settings and "mode" in public_settings.get("proxy") and public_settings.get("proxy").get("mode") == "application":
+    if public_settings is not None and "proxy" in public_settings and "mode" in public_settings.get("proxy") and public_settings.get("proxy").get("mode") == "none":
+        default_configs["MDSD_PROXY_MODE"] = "none"
+        unset_proxy()
+        proxySet = True
+    elif public_settings is not None and "proxy" in public_settings and "mode" in public_settings.get("proxy") and public_settings.get("proxy").get("mode") == "application":
         default_configs["MDSD_PROXY_MODE"] = "application"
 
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -1097,7 +1097,11 @@ def handle_mcs_config(public_settings, protected_settings, default_configs):
     proxySet = False
 
     # fetch proxy settings
-    if public_settings is not None and "proxy" in public_settings and "mode" in public_settings.get("proxy") and public_settings.get("proxy").get("mode") == "application":
+    if public_settings is not None and "proxy" in public_settings and "mode" in public_settings.get("proxy") and public_settings.get("proxy").get("mode") == "none":
+        default_configs["MDSD_PROXY_MODE"] = "none"
+        unset_proxy()
```
simathih, line 1102: unset_proxy() is called in ln 1147 as well, not necessary here

simathih (dismissed):

AzureMonitorAgent/agent.py
hestolz: only if `!proxySet`; so, this is a bit confusing — I understand the intention that `none` in AMA public settings should cause us to call `unset_proxy()`, but now 
1. we are calling `unset_proxy()` twice
2. the meaning of `proxySet = True` is ambiguous

thperapp: refactored to clean up messy logic and align if/else blocks with actual expected order of precedence.

hestolz (approved):

PR #2166, opened 2026-04-29
Add alias for `which` via `command -v` for SUSE 16

dylanbun:
Since `which` does not come by default with SUSE 16, we will create an alias via `command -v` to substitute for `which` on SUSE/SLES 16.


dylanbun:
@copilot review this PR, point out anything that could impact AMA's ability to be installed on a SUSE 16 VM.

simathih (approved):

PR #2174, opened 2026-05-19
VMAccess: Add Azure Linux / Mariner support to sshd_config reset

jonathanbrenes:
## Problem

As preparation for the release of Azure Linux 4.0, this patch adds support for Azure Linux and CBL-Mariner to the VMAccess extension's sshd_config reset functionality.

When `az vm user reset-ssh` is triggered on Azure Linux (or CBL-Mariner), `_get_default_ssh_config_filename()` falls through to the generic `default` resource file because `NAME="Azure Linux"` does not match any of the existing distro regexes (centos, debian, fedora, red hat, suse, ubuntu).

The `default` resource file is a Debian/Ubuntu-era sshd_config that causes the following problems on Azure Linux:

- **SFTP broken**: wrong Subsystem path (`/usr/lib/openssh/sftp-server` instead of `/usr/libexec/openssh/sftp-server`)
- **Multiple deprecated directive warnings** from sshd -t
- **Unable to load host key**: `/etc/ssh/ssh_host_dsa_key` (DSA removed)
- **Include directive dropped**: `Include /etc/ssh/sshd_config.d/*.conf` removed, causing all drop-in config snippets (cloud-init, Azure) to be silently lost
- **PasswordAuthentication explicitly set to yes**
- **PermitRootLogin** uses old `without-password` syntax

## Fix

1. Added `azure\s?linux|mariner` regex to `_get_default_ssh_config_filename()` returning `azurelinux_default`. This matches:
   - Azure Linux: `NAME="Azure Linux"`
   - CBL-Mariner: `NAME="Common Base Linux Mariner"`

2. Added `VMAccess/resources/azurelinux_default`: stock Azure Linux sshd_config with no deprecated directives, correct SFTP path, and Include directive preserved.

## Testing

Tested on an Azure Linux VM:

| Check | Before (default) | After (azurelinux_default) |
|---|---|---|
| sshd -t errors | 6 warnings/errors | 0 |
| SFTP Subsystem path | `/usr/lib/openssh/sftp-server` (wrong) | `/usr/libexec/openssh/sftp-server` (correct) |
| Include directive | Missing | Present |
| DSA host key reference | Present (file doesn't exist) | Absent |
| PasswordAuthentication | yes | Commented out (default: no) |
| sshd restart | Clean (after warnings) | Clean |

## Files Changed

- `VMAccess/vmaccess.py` — added Azure Linux / Mariner match in `_get_default_ssh_config_filename()`
- `VMAccess/resources/azurelinux_default` — new sshd_config template for Azure Linux

### DIFF
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -398,6 +398,8 @@ def _get_default_ssh_config_filename():
             return "debian_default"
         if re.search("fedora", OSName, re.IGNORECASE):
             return "fedora_default"
+        if re.search(r"azure\s?linux|mariner", OSName, re.IGNORECASE):
+            return "fedora_default"
         if re.search(r"red\s?hat", OSName, re.IGNORECASE):
             return "redhat_default"
         if re.search("suse", OSName, re.IGNORECASE):
--- a/VMAccess/vmaccess.py
+++ b/VMAccess/vmaccess.py
@@ -399,7 +399,7 @@ def _get_default_ssh_config_filename():
         if re.search("fedora", OSName, re.IGNORECASE):
             return "fedora_default"
         if re.search(r"azure\s?linux|mariner", OSName, re.IGNORECASE):
-            return "fedora_default"
+            return "azurelinux_default"
         if re.search(r"red\s?hat", OSName, re.IGNORECASE):
             return "redhat_default"
         if re.search("suse", OSName, re.IGNORECASE):
--- a/VMAccess/resources/azurelinux_default
+++ b/VMAccess/resources/azurelinux_default
@@ -1,0 +1,132 @@
+#	$OpenBSD: sshd_config,v 1.105 2024/12/03 14:12:47 dtucker Exp $
+
+# This is the sshd server system-wide configuration file.  See
+# sshd_config(5) for more information.
+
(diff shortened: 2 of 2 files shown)


birukyitbarek:
See merge conflict. Branch is behind master. After rebase the raw-string convention will be obvious.

VMAccess/vmaccess.py
```diff
@@ -398,6 +398,8 @@ def _get_default_ssh_config_filename():
             return "debian_default"
         if re.search("fedora", OSName, re.IGNORECASE):
             return "fedora_default"
+        if re.search(r"azure\s?linux|mariner", OSName, re.IGNORECASE):
+            return "fedora_default"
```
norakoiralamsft: This should reurn azurelinux_default

norakoiralamsft (changes requested):

VMAccess/vmaccess.py
jonathanbrenes: Good catch, you are right.

I verified the OS detection path and on Azure Linux the detected OS name is "Azure Linux", so this branch is hit. Returning fedora_default there is not what we want for the Azure Linux-specific template path.

I updated azurelinux_default so the dedicated Azure Linux sshd baseline is used.

norakoiralamsft (approved):

PR #2177, opened 2026-05-23
VMBackup: Fix Python 3.12+ compatibility for Azure Linux 4.0

jonathanbrenes:
## Problem

The VMSnapshot (VMBackup) extension crashes on Azure Linux 4.0 (Python 3.14) due to four Python APIs that were removed in recent Python versions. The extension fails at import time, causing Azure Backup jobs to fail with error **400039** after hours of retries.

### Removed APIs used by VMBackup

| API | Removed In | Usage |
|-----|-----------|-------|
| `distutils.version.LooseVersion` | Python 3.12 | Version string comparisons |
| `imp` module | Python 3.12 | Fallback module loader for waagent |
| `platform.dist()` | Python 3.8 | Distro detection in telemetry |
| `platform.linux_distribution()` | Python 3.8 | Distro detection for patching |

### `WaagentLib.py`
- Add `try/except` for `LooseVersion` import with a minimal shim class providing comparison operators (`__lt__`, `__gt__`, `__eq__`, `__le__`, `__ge__`) via regex-based version parsing.

> **Why a shim instead of `packaging.version.Version`?** PEP 632 recommends the third-party `packaging` library as the replacement, but it is not suitable here: (1) it is not in the stdlib and may be absent on minimal VM installs, (2) it enforces strict PEP 440 parsing and rejects the loose version strings this codebase passes (e.g. kernel versions, agent versions), and (3) adding a new dependency to a VM extension that must run on every Linux distro from RHEL 7 (Python 2.7) through AZL4 (Python 3.14) is too risky. The inline shim is ~20 lines, zero-dependency, and drop-in compatible.

### `WAAgentUtil.py`
- Add `try/except` around the `import imp` fallback so it doesn't crash on Python 3.12+ when the primary `importlib.util` path fails.

### `HandlerUtil.py`
- Replace `platform.dist()` in `get_dist_info()` with `/etc/os-release` parsing (`NAME` + `VERSION` fields).
- Fix invalid escape sequence: `replace('\/', '/')` was a no-op (invalid escape treated as literal); changed to `replace('\\/', '/')`.

### `patch/__init__.py`
- Add `/etc/os-release` `ID` parsing to `DistInfo()` fallback for distros where `platform.linux_distribution()` is gone.
- Add `'azurelinux'` mapping in `GetMyPatching()` → `AzureLinuxPatching`.

### `patch/AzureLinuxPatching.py` *(new)*
- Dedicated patching class for Azure Linux 4.0.
- All binary paths set to `/usr/bin/` (AZL4 uses merged `/usr`).
- Uses `dnf` for package installation.

### End-to-end Azure Backup on AZL4
Full backup job completed successfully (26 min, all phases: pre-snapshot, snapshot, post-snapshot).

### Backward compatibility
All patches are no-ops on older Python versions (guarded by `try/except` with fallback to original imports). Verified across 15 distros (Python 2.7.5 – 3.14.3), including end-to-end Azure Backup on all Python 2 VMs:

| Distro | Python | Result |
|--------|--------|--------|
| RHEL 8.10 | 3.6.8 | PASS |
| RHEL 9.7 | 3.9.25 | PASS |
| Oracle Linux 9.7 | 3.9.25 | PASS |
| SLES 15 SP6 | 3.6.15 | PASS |
| SLES 16.0 | 3.13.13 | PASS |
| Debian 12 | 3.11.2 | PASS |
| Ubuntu 20.04 | 3.8.10 | PASS |
| Ubuntu 22.04 | 3.10.12 | PASS |
| Ubuntu 24.04 | 3.12.3 | PASS |
| Azure Linux 4.0 | 3.14.3 | PASS |
| CentOS 7.9 | 2.7.5 | PASS |
| RHEL 7.9 | 2.7.5 | PASS |
| Ubuntu 16.04 | 2.7.12 | PASS |
| Debian 10 | 2.7.16 | PASS |
| SLES 12 SP5 | 2.7.18 | PASS |

### DIFF
--- a/VMBackup/main/Utils/HandlerUtil.py
+++ b/VMBackup/main/Utils/HandlerUtil.py
@@ -627,8 +627,20 @@ class HandlerUtility:
                 distinfo[0] = distinfo[0].strip()
                 return  distinfo[0]+"-"+distinfo[1],platform.release()
             else:
-                distinfo = platform.dist()
-                return  distinfo[0]+"-"+distinfo[1],platform.release()
+                # platform.dist() removed in Python 3.8; parse /etc/os-release instead
+                try:
+                    with open("/etc/os-release", "r") as f:
+                        os_name, os_version = "Unknown", "Unknown"
+                        for line in f:
+                            k, _, v = line.strip().partition("=")
+                            v = v.strip('"')
+                            if k == "NAME":
+                                os_name = v
+                            elif k == "VERSION":
+                                os_version = v
+                        return os_name + "-" + os_version, platform.release()
+                except Exception:
+                    return "Unknown", "Unknown"
         except Exception as e:
             errMsg = 'Failed to retrieve the distinfo with error: %s, stack trace: %s' % (str(e), traceback.format_exc())
             self.log(errMsg)
@@ -715,7 +727,7 @@ class HandlerUtility:
         stat_rept.timestampUTC = date_place_holder
         date_string = r'\/Date(' + str((int)(time_span)) + r')\/'
         stat_rept = "[" + json.dumps(stat_rept, cls = ComplexEncoder) + "]"
-        stat_rept = stat_rept.replace('\\\/', '\/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
+        stat_rept = stat_rept.replace('\\/', '/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
         stat_rept = stat_rept.replace(date_place_holder,date_string)
 
         # Add Status as sub-status for Status to be written on Status-File
--- a/VMBackup/main/Utils/HandlerUtil.py
+++ b/VMBackup/main/Utils/HandlerUtil.py
@@ -81,6 +81,7 @@ import datetime
 import Utils.Status
 from Utils.EventLoggerUtil import EventLogger
 from Utils.LogHelper import LoggingLevel, LoggingConstants, FileHelpers
+from Utils.DistroUtil import read_os_release
 from MachineIdentity import MachineIdentity
 import ExtensionErrorCodeHelper
 import traceback
@@ -628,19 +629,10 @@ class HandlerUtility:
                 return  distinfo[0]+"-"+distinfo[1],platform.release()
             else:
                 # platform.dist() removed in Python 3.8; parse /etc/os-release instead
-                try:
-                    with open("/etc/os-release", "r") as f:
-                        os_name, os_version = "Unknown", "Unknown"
-                        for line in f:
-                            k, _, v = line.strip().partition("=")
-                            v = v.strip('"')
-                            if k == "NAME":
-                                os_name = v
-                            elif k == "VERSION":
-                                os_version = v
-                        return os_name + "-" + os_version, platform.release()
-                except Exception:
-                    return "Unknown", "Unknown"
+                osr = read_os_release()
+                if osr:
+                    return osr.get("NAME", "Unknown") + "-" + osr.get("VERSION", "Unknown"), platform.release()
+                return "Unknown", "Unknown"
         except Exception as e:
             errMsg = 'Failed to retrieve the distinfo with error: %s, stack trace: %s' % (str(e), traceback.format_exc())
             self.log(errMsg)
@@ -727,7 +719,7 @@ class HandlerUtility:
         stat_rept.timestampUTC = date_place_holder
         date_string = r'\/Date(' + str((int)(time_span)) + r')\/'
         stat_rept = "[" + json.dumps(stat_rept, cls = ComplexEncoder) + "]"
-        stat_rept = stat_rept.replace('\\/', '/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
+        stat_rept = stat_rept.replace('\\\/', '\/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
         stat_rept = stat_rept.replace(date_place_holder,date_string)
 
         # Add Status as sub-status for Status to be written on Status-File
--- a/VMBackup/main/Utils/WAAgentUtil.py
+++ b/VMBackup/main/Utils/WAAgentUtil.py
@@ -66,10 +66,11 @@ try:
             spec.loader.exec_module(waagent)       
         except ImportError:
             # For Python 3.4 and earlier, use imp module
-            import imp
-            waagent = imp.load_source('waagent', agentPath)
-        except Exception:
-            raise Exception("Can't load waagent.")
+            try:
+                import imp
+                waagent = imp.load_source('waagent', agentPath)
+            except ImportError:
+                raise Exception("Can't load waagent: importlib.util and imp both unavailable.")
     else:
         raise Exception("Can't load new or old waagent. Agent path not found.")
 except Exception as e:
--- a/VMBackup/main/Utils/WAAgentUtil.py
+++ b/VMBackup/main/Utils/WAAgentUtil.py
@@ -58,19 +58,29 @@ try:
        # Search for the old agent path if the new one is not found
        agentPath = searchWAAgentOld()
     if agentPath:
+        # Choose the loader by capability, not by exception type. Previously this
+        # relied on `except ImportError`, but on Python 3.3-3.4 the missing
+        # `importlib.util.module_from_spec` raises AttributeError (not ImportError),
+        # which escaped the handler and crashed the extension (broke PR #2124).
+        # `imp` was also removed in Python 3.12, so a blind fallback is unsafe.
         try:
-            # For Python 3.5 and later, use importlib
             import importlib.util
+            useImportlib = hasattr(importlib.util, 'module_from_spec')
+        except ImportError:
+            # Python 2.6 / 2.7 have no importlib.util
+            useImportlib = False
+
+        if useImportlib:
+            # Python 3.5+
             spec = importlib.util.spec_from_file_location('waagent', agentPath)
             waagent = importlib.util.module_from_spec(spec)
-            spec.loader.exec_module(waagent)       
-        except ImportError:
-            # For Python 3.4 and earlier, use imp module
-            try:
-                import imp
-                waagent = imp.load_source('waagent', agentPath)
-            except ImportError:
-                raise Exception("Can't load waagent: importlib.util and imp both unavailable.")
+            spec.loader.exec_module(waagent)
+        else:
+            # Python 2.6 - 3.4. 'imp' was removed in 3.12, but every version that
+            # reaches this branch still has 'imp', because 3.5+ always provides
+            # importlib.util.module_from_spec and takes the branch above.
+            import imp
+            waagent = imp.load_source('waagent', agentPath)
     else:
         raise Exception("Can't load new or old waagent. Agent path not found.")
 except Exception as e:
--- a/VMBackup/main/WaagentLib.py
+++ b/VMBackup/main/WaagentLib.py
@@ -56,7 +56,48 @@ import zipfile
 import json
 import datetime
 import xml.sax.saxutils
-from distutils.version import LooseVersion
+try:
+    from distutils.version import LooseVersion
+except ImportError:
+    # distutils removed in Python 3.12; minimal shim for version comparisons
+    import re as _re
+    class LooseVersion(object):
+        def __init__(self, vstring):
+            self.vstring = str(vstring)
+            self._cmp_key = self._parse(self.vstring)
+        @staticmethod
+        def _parse(s):
+            parts = []
+            for tok in _re.split(r'(\d+)', s):
+                if tok.isdigit():
+                    parts.append(int(tok))
+                elif tok:
+                    parts.append(tok)
+            return parts
+        def __str__(self):
+            return self.vstring
+        def __repr__(self):
+            return 'LooseVersion(%r)' % self.vstring
+        def _cmp(self, other):
+            if not isinstance(other, LooseVersion):
+                other = LooseVersion(other)
+            a, b = self._cmp_key, other._cmp_key
+            for i in range(max(len(a), len(b))):
+                ai = a[i] if i < len(a) else 0
+                bi = b[i] if i < len(b) else 0
+                if type(ai) != type(bi):
+                    ai, bi = str(ai), str(bi)
+                if ai < bi:
+                    return -1
+                if ai > bi:
+                    return 1
+            return 0
+        def __lt__(self, other): return self._cmp(other) < 0
+        def __le__(self, other): return self._cmp(other) <= 0
+        def __eq__(self, other): return self._cmp(other) == 0
+        def __ge__(self, other): return self._cmp(other) >= 0
+        def __gt__(self, other): return self._cmp(other) > 0
+        def __ne__(self, other): return self._cmp(other) != 0
 
 if not hasattr(subprocess, 'check_output'):
     def check_output(*popenargs, **kwargs):
@@ -4514,6 +4555,8 @@ def GetMyDistro(dist_class_name=''):
                 Distro = 'oracle'
             elif ('redhat'.lower() in Distro.lower()):
                 Distro = 'redhat'
+            elif ('azurelinux' in Distro.lower()):
+                Distro = 'fedora'
             elif ('Kali'.lower() in Distro.lower()):
                 Distro = 'Kali'
             elif ('FreeBSD'.lower() in  Distro.lower() or 'gaia'.lower() in Distro.lower() or 'panos'.lower() in Distro.lower()):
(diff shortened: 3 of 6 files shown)


jonathanbrenes:
## Evidence: Extension is broken on Azure Linux 4.0

### Production failure

On an unpatched Azure Linux 4.0 VM (Python 3.14.3), the VMSnapshot extension **cannot start at all**. Every enable attempt exits with code 1, causing Azure Backup jobs to fail with error `400039` after ~9 hours of retries.

**Extension log (`shell.log`) — 7 consecutive failures:**
```
Thu May 21 05:04:31 PM UTC 2026- 1 returned from handle.py
Thu May 21 05:06:01 PM UTC 2026- 1 returned from handle.py
Thu May 21 08:04:37 PM UTC 2026- 1 returned from handle.py
Thu May 21 11:24:42 PM UTC 2026- 1 returned from handle.py
Fri May 22 12:49:51 AM UTC 2026- 1 returned from handle.py
Fri May 22 02:36:42 AM UTC 2026- 1 returned from handle.py
Fri May 22 03:08:32 AM UTC 2026- 1 returned from handle.py
```

**Crash traceback — cascading `ModuleNotFoundError`:**
```
File "WAAgentUtil.py", line 66
    spec.loader.exec_module(waagent)
  File "WaagentLib.py", line 59
    from distutils.version import LooseVersion
ModuleNotFoundError: No module named 'distutils'

During handling of the above exception:
  File "WAAgentUtil.py", line 69
    import imp
ModuleNotFoundError: No module named 'imp'
```

The import chain `handle.py` → `mounts.py` → `DiskUtil.py` → `HandlerUtil.py` → `WAAgentUtil.py` → `WaagentLib.py` fails at the very first import, so **no code in the extension ever executes**.

### All 4 removed APIs confirmed broken

```
Python version: 3.14.3
distutils.version.LooseVersion: FAIL - No module named 'distutils'
import imp:                     FAIL - No module named 'imp'
platform.linux_distribution():  FAIL - module 'platform' has no attribute 'linux_distribution'
platform.dist():                FAIL - module 'platform' has no attribute 'dist'
```

### After patching — extension works

**Unit tests on AZL4 VM (8/8 pass):**
```
=== Test 1: Import chain ===
  Utils.HandlerUtil: OK
=== Test 2: DistInfo ===
  DistInfo() = ['azurelinux', '4.0']
=== Test 3: GetMyPatching ===
  Patching class = AzureLinuxPatching
=== Test 4: Binary paths ===
  cryptsetup_path = /usr/bin/cryptsetup -> EXISTS
  getenforce_path = /usr/bin/getenforce -> EXISTS
  setenforce_path = /usr/bin/setenforce -> EXISTS
=== Test 5: get_dist_info ===
  get_dist_info() = ('Azure Linux-4.0 (Cloud Variant Alpha2)', '6.18.5-...')
=== Test 7: LooseVersion shim ===
  LooseVersion: OK (comparisons work)
=== Test 8: Full handle.py import chain ===
  mounts.Mounts: OK
=== ALL TESTS PASSED ===
```

**End-to-end backup — PASS:**

| Metric | Unpatched | Patched |
|--------|-----------|---------|
| Duration | ~9 hours → FAILED | 26 minutes |
| Backup size | 0 MB | 93 MB |
| Exit code | 1 (all attempts) | 0 |
| Error | 400039 GuestAgentSnapshotTaskStatusError | None |

### Backward compatibility — 15 distros, all PASS

| Distro | Python | Result |
|--------|--------|--------|
| RHEL 8.10 | 3.6.8 | PASS |
| RHEL 9.7 | 3.9.25 | PASS |
| Oracle Linux 9.7 | 3.9.25 | PASS |
| SLES 15 SP6 | 3.6.15 | PASS |
| SLES 16.0 | 3.13.13 | PASS |
| Debian 12 | 3.11.2 | PASS |
| Ubuntu 20.04 | 3.8.10 | PASS |
| Ubuntu 22.04 | 3.10.12 | PASS |
| Ubuntu 24.04 | 3.12.3 | PASS |
| Azure Linux 4.0 | 3.14.3 | PASS |
| CentOS 7.9 | 2.7.5 | PASS |
| RHEL 7.9 | 2.7.5 | PASS |
| Ubuntu 16.04 | 2.7.12 | PASS |
| Debian 10 | 2.7.16 | PASS |
| SLES 12 SP5 | 2.7.18 | PASS |

All patches are guarded by `try/except` — on older Python versions where the deprecated APIs still exist, the original code path runs unchanged (zero behavior change). Python 2 VMs were validated with full end-to-end Azure Backup (baseline + patched).

VMBackup/main/WaagentLib.py
arisettisanjana: ModuleNotFoundError: No module named 'crypt'
when running with python3.14

jonathanbrenes:
### Update: crypt removal handling (Python 3.13+ / Ubuntu 26.04) + hardened `LooseVersion` shim

Pushed one commit on top of the existing Python 3.12 compatibility work.

**1. `crypt` removal (Python 3.13+, e.g. Ubuntu 26.04)**
`crypt` was removed from the stdlib in Python 3.13, which crashed the extension at import. `WaagentLib.py` now guards the import with fallbacks:
- `from crypt import crypt` (Python < 3.13)
- `from legacycrypt import crypt` (Python 3.13+)
- `from passlib.hash import sha512_crypt`

`gen_password_hash()` uses whichever backend resolved, otherwise raises a clear `ImportError` instead of crashing at import.

**2. Hardened `LooseVersion` fallback shim**
Adopted the review-hardened shim (aligns with vupadhyay-ms `a8b9ff6`):
- Splits on `.`, `-`, `_` (handles kernel-style version strings)
- Numeric tokens compare as ints; pre-release tokens (`alpha < beta < rc < release`) use negative sentinels
- Mixed-type positions are coerced to `str` in `_cmp()` to avoid `TypeError` on Python 3
- `__ne__` (for Py 2.7) and `__repr__` retained; six comparators delegate to one `_cmp()` helper

**Validation**
- Shim unit-checked (pre-release ordering + mixed-type safety), all pass.
- Patched code executed under native Python on distro VMs from **2.7.5 through 3.12.3** (import guard, hashing, version compare, distro/patch-class selection all pass).
- Scope: VMBackup only.

jonathanbrenes:
**Tested distros / Python versions** (patched code executed on-VM under each native Python — import guard, password hashing, version compare, and distro/patch-class selection all PASS):

| Distro | Python | Result |
| --- | --- | --- |
| CentOS 7.9 | 2.7.5 | PASS |
| RHEL 7.9 | 2.7.5 | PASS |
| SLES 12 SP5 | 3.4.10 (+ 2.7.18) | PASS |
| Ubuntu 16.04 | 3.5.2 | PASS |
| RHEL 8.10 | 3.6.8 | PASS |
| SLES 15 SP6 | 3.6.15 | PASS |
| Ubuntu 20.04 | 3.8.10 | PASS |
| Oracle Linux 9.7 | 3.9.25 | PASS |
| RHEL 9.7 | 3.9.25 | PASS |
| Ubuntu 22.04 | 3.10.12 | PASS |
| Debian 12 | 3.11.2 | PASS |
| Ubuntu 24.04 | 3.12.3 | PASS |

Not yet re-validated on-VM (Azure run-command guest handler got stuck — infra, not code): **Ubuntu 26.04** (Python 3.13+, the `crypt`-removed case), **Azure Linux 4.0** (Python 3.14), **SLES 16**. The `crypt` fallback and shim were verified locally for these; the AZL4 end-to-end backup path was validated in the original change.

VMBackup/main/WaagentLib.py
```diff
@@ -56,23 +55,55 @@
...
+try:
+    from crypt import crypt as crypt
+    cryptImported = True
+except ImportError:
+    pass
+
+if cryptImported == False:
+    if (sys.version_info[0] == 3 and sys.version_info[1] >= 13) or (sys.version_info[0] > 3):
+        try:
+            from legacycrypt import crypt
```
arisettisanjana, line 78: i remember testing it out with the legacycrypt, i have seen an error with 3.14 python version, can you confirm that it is been tested in ubuntu26

VMBackup/main/WaagentLib.py
```diff
@@ -56,7 +55,80 @@
 import json
 import datetime
 import xml.sax.saxutils
-from distutils.version import LooseVersion
+
+cryptImported = False
+passLibImported = False
+
+try:
+    from crypt import crypt as crypt
```
vupadhyay-ms: **nit:** `from crypt import crypt as crypt` — the `as crypt` is a no-op. `from crypt import crypt` does the same thing.

VMBackup/main/WaagentLib.py
```diff
@@ -56,7 +55,80 @@
...
+    cryptImported = True
+except ImportError:
+    pass
+
+if cryptImported == False:
+    if (sys.version_info[0] == 3 and sys.version_info[1] >= 13) or (sys.version_info[0] > 3):
+        try:
+            from legacycrypt import crypt
+            cryptImported = True
+        except ImportError:
```
vupadhyay-ms: **non-blocking:** All three fallback branches (`crypt`, `legacycrypt`, `passlib`) use `except ImportError: pass`, which swallows the reason silently. If none of the three are available and `gen_password_hash` is later hit, debugging becomes archaeology. Could we log once at import time (or on first call) which sources were tried and failed?

Also worth a comment somewhere that `legacycrypt` and `passlib` are not stdlib — same reasoning we've discussed elsewhere for skipping `packaging.version`. Not a blocker since VMBackup doesn't call `gen_password_hash` today, but a note prevents someone from wiring this into a real code path assuming it works on every VM.

VMBackup/main/WaagentLib.py
```diff
@@ -377,7 +449,16 @@ def gen_password_hash(self, password, crypt_id, salt_len):
         collection = string.ascii_letters + string.digits
         salt = ''.join(random.choice(collection) for _ in range(salt_len))
         salt = "${0}${1}".format(crypt_id, salt)
-        return crypt.crypt(password, salt)
+
+        if cryptImported:
+            return crypt(password, salt)
+        elif passLibImported:
+            return sha512_crypt.hash(password)
```
vupadhyay-ms: **blocking (potential defect):** the caller constructed `salt = "${crypt_id}${salt}"` at line 451, but the `passlib` branch doesn't pass it in:

```python
elif passLibImported:
    return sha512_crypt.hash(password)   # <-- caller's salt is dropped
```

`passlib.hash.sha512_crypt.hash(password)` generates its own random salt with default rounds, so if this branch ever runs the customer gets a different hash format than the `crypt` and `legacycrypt` branches produce. That's a silent semantic drift under the customer.

Suggest either:
```python
sha512_crypt.using(salt=<parsed_salt_chars>, rounds=<rounds>).hash(password)
```
(parsing the salt back out of the `$id$salt` string), or just dropping the passlib branch entirely if we're not willing to preserve the exact hash format.

VMBackup/main/Utils/WAAgentUtil.py
```diff
@@ -66,10 +66,11 @@ def searchWAAgentOld():
             spec.loader.exec_module(waagent)       
         except ImportError:
```
vupadhyay-ms: this is exactly the shape that broke PR #2124 — `except ImportError` doesn't catch `AttributeError`, which is what Py versions have started raising when `importlib.util.module_from_spec` isn't available. A capability check is safer and immune to future exception-type churn:

```python
import importlib.util
if hasattr(importlib.util, 'module_from_spec'):
    spec = importlib.util.spec_from_file_location('waagent', agentPath)
    waagent = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(waagent)
else:
    import imp
    waagent = imp.load_source('waagent', agentPath)
```

T

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -715,7 +727,7 @@ def do_status_report(self, operation, status, status_code, message, taskId = Non
         stat_rept.timestampUTC = date_place_holder
         date_string = r'\/Date(' + str((int)(time_span)) + r')\/'
         stat_rept = "[" + json.dumps(stat_rept, cls = ComplexEncoder) + "]"
-        stat_rept = stat_rept.replace('\\\/', '\/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
+        stat_rept = stat_rept.replace('\\/', '/') # To fix the datetime format of CreationTime to be consumed by C# DateTimeOffset
```
vupadhyay-ms: **blocking (behavior change, not a pure fix):** the original `.replace('\\\/', '\/')` was a no-op — `\/` is an invalid escape sequence and Python treated it literally, so the replacement was `\/` → `\/` (identity). The new code `.replace('\\/', '/')` actually strips the backslashes, so the C# `DateTimeOffset` consumer downstream now sees `/Date(...)/` instead of `\/Date(...)\/`. That's a real payload change.

Could we get a sign-off from the C# side (the `CreationTime` consumer for this status file), or split this into its own PR so it can be audited independently? Right now it rides along inside a Python-3 compatibility PR, which makes it easy to miss during review.

VMBackup/main/Utils/HandlerUtil.py
```diff
@@ -627,8 +627,20 @@ def get_dist_info(self):
                 distinfo[0] = distinfo[0].strip()
                 return  distinfo[0]+"-"+distinfo[1],platform.release()
             else:
-                distinfo = platform.dist()
-                return  distinfo[0]+"-"+distinfo[1],platform.release()
+                # platform.dist() removed in Python 3.8; parse /etc/os-release instead
+                try:
+                    with open("/etc/os-release", "r") as f:
```
vupadhyay-ms: **nit:** the exact same `/etc/os-release` parsing block appears in three places:
- here (`get_dist_info`)
- `WaagentLib.py` `DistInfo()` around line 4641
- `patch/__init__.py` `DistInfo()` around line 64

All three do the same `open` + `partition('=')` + `strip('"')` loop. Would you consider extracting a single `read_os_release()` helper (e.g., in `Utils/`) so future distro additions only need one change? Copy-paste tends to drift over time.

VMBackup/main/patch/__init__.py
```diff
@@ -114,6 +130,8 @@ def GetMyPatching(logger):
             Distro = 'oracle'
         elif ('redhat'.lower() in Distro.lower()):
             Distro = 'redhat'
+        elif ('azurelinux' in Distro.lower()):
+            Distro = 'AzureLinux'
```
vupadhyay-ms, line 129: **nit:** the AzureLinux distro string is set to two different values in two code paths:
- `WaagentLib.GetMyDistro` (line 4598) maps `azurelinux` → `'fedora'`
- here in `GetMyPatching` maps `azurelinux` → `'AzureLinux'`

Is that intentional (patcher class name vs distro-string consumers), or should they match? Either way a one-line comment explaining the divergence would help future readers — right now it looks like a bug even if it isn't.

vupadhyay-ms (commented):
Rest all looks good to me.

VMBackup/main/WaagentLib.py
jonathanbrenes: **1. `nit` — redundant `as crypt` alias** 
Changed `from crypt import crypt as crypt` → `from crypt import crypt`.

VMBackup/main/WaagentLib.py
jonathanbrenes: Each backend's failure reason is now captured in a `cryptImportErrors` list and surfaced in `gen_password_hash()`'s error message (deterministic "on first call" reporting — there's no reliable logger at module import in the vendored waagent). Also added a comment noting `legacycrypt` is not stdlib and that hashing is only available where a backend is present.

VMBackup/main/WaagentLib.py
jonathanbrenes: Took the "drop it" option — the `passlib` branch and `passLibImported` flag are removed; hashing is now `crypt` → `legacycrypt`, else a clear `ImportError`. This eliminates the format drift and the non-stdlib dependency. Updated `test_waagentlib_password_hash.py` accordingly (removed the obsolete passlib test).

VMBackup/main/Utils/WAAgentUtil.py
jonathanbrenes: Replaced the exception-based branch with `hasattr(importlib.util, 'module_from_spec')`. One refinement over the snippet: I also guarded the `import importlib.util` itself, since Python 2.7 has no `importlib.util` (a bare import raises `ImportError` there). Covers all three cases — `importlib.util` missing (2.6/2.7), `module_from_spec` missing (3.3–3.4), present (3.5+).

VMBackup/main/Utils/HandlerUtil.py
jonathanbrenes: Good catch prompting a closer look. I tested a regression script that reproduces the exact `do_status_report` sequence and proved the two forms emit **byte-identical** output — the `replace()` runs *before* the `\/Date(...)\/` token is substituted, and `json.dumps` doesn't emit `\/` for normal content, so nothing changes on the wire. To keep this PR a pure compatibility change, I've **reverted the line to the `master` original** (verified: no diff vs master). The test stays in to guard the behavior.

VMBackup/main/Utils/HandlerUtil.py
jonathanbrenes: Extracted a single `Utils/DistroUtil.read_os_release()` helper and wired all three sites (`HandlerUtil.get_dist_info`, `patch.DistInfo`, `WaagentLib.DistInfo`). `WaagentLib` uses a guarded import so the vendored waagent still runs standalone.

VMBackup/main/patch/__init__.py
jonathanbrenes: It's intentional, and I added comments at both sites saying so. I also validated it on a **real Azure Linux 4.0 VM (Python 3.14.3)** with a small script — 8/8 checks pass: `patch.DistInfo()`/`WaagentLib.DistInfo()` = `['azurelinux','4.0']`, `GetMyPatching()` → `AzureLinuxPatching`, `GetMyDistro()` → `fedoraDistro`. So the patch layer uses the dedicated class while the waagent distro layer reuses the Fedora family, exactly as intended.

vupadhyay-ms:
LGTM

arisettisanjana (approved):
have tested the latest changes with py3.14 version and it is working fine, LGTM

PR #2178, opened 2026-06-02
[AMA] Improve TST CL path handling

hestolz:


AzureMonitorAgent/ama_tst/release.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
simathih, line 1: do we need to do a separate release? I thought its packaged in our VM extension and deployed automatically

simathih (approved):

AzureMonitorAgent/ama_tst/release.sh
hestolz: it's also a github release -- here in this repo. whether we should continue doing that can be up for discussion (scrum tomorrow, e.g.). it may be a holdover from when we wanted to make it available to customers who had yet to upgrade to the AMA version with TST packaged in the zip

PR #2184, opened 2026-06-09
Set default config file permissions after renaming

simathih:
Add permission change for the updated config file

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -968,7 +968,12 @@ def enable():
             new_config = "\n".join(["export {0}={1}".format(key, value) for key, value in default_configs.items()]) + "\n"
 
             tmp_fd = os.open(temp_config_file, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o400)
-            with os.fdopen(tmp_fd, "w") as f:
+            try:
+                f = os.fdopen(tmp_fd, "w")
+            except:
+                os.close(tmp_fd)
+                raise
+            with f:
                 f.write(new_config)
 
             if not os.path.isfile(temp_config_file):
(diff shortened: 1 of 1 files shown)


AzureMonitorAgent/agent.py
hestolz: let's ensure it's perm'd from the start, like the below or appropriate alternative, otherwise we're leaving it open temporarily and also using formatted `system` commands not python builtins
```
tmp_fd = os.open(temp_config_file, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o400)
```

hestolz (changes requested):

AzureMonitorAgent/agent.py
```diff
@@ -967,14 +967,15 @@ def enable():
         if os.path.isfile(config_file):
             new_config = "\n".join(["export {0}={1}".format(key, value) for key, value in default_configs.items()]) + "\n"
 
-            with open(temp_config_file, "w") as f:
+            tmp_fd = os.open(temp_config_file, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o400)
+            with os.fdopen(tmp_fd, "w") as f:
```
thperapp: is an os.close call required?

thperapp (changes requested):

AzureMonitorAgent/agent.py
simathih: fixed

simathih: 'with os.fdopen()' would have closed it automatically too but handling a case when fdopen fails

thperapp (approved):

hestolz (approved):

PR #2190, opened 2026-06-27
AMA: enable libcurl ODS upload (ENABLE_CURL_UPLOAD) in eastus2euap and centraluseuap

simathih:
## Summary

Gate the libcurl-based ODS upload path (`ENABLE_CURL_UPLOAD`) behind region detection so it is enabled **only in `eastus2euap`** for canary rollout.

## Changes

- Added `enableCurlUpload` to `feature_support_matrix` in `is_feature_enabled()`, scoped to `['eastus2euap']`.
- In `enable()`, when the feature is enabled for the region, write `export ENABLE_CURL_UPLOAD=true` into `/etc/default/azuremonitoragent` via `default_configs`.

## Why this approach

- Reuses the existing region-gating mechanism used by `useDynamicSSL` / `enableAzureOTelCollector` (region from IMDS via `get_azure_environment_and_region()`).
- Respects per-VM `previewFeatures` override flag files: `enableCurlUpload` force-enables and `enableCurlUploadDisabled` force-disables, useful for targeted testing or emergency opt-out.
- Only sets the variable when enabled; all other regions keep the default opt-in (off) behavior of the agent's `ENABLE_CURL_UPLOAD` env var.

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -909,6 +909,12 @@ def enable():
     ssl_cert_var_name, ssl_cert_var_value = get_ssl_cert_info('Enable')
     default_configs[ssl_cert_var_name] = ssl_cert_var_value
 
+    # Enable the libcurl-based ODS upload path (ENABLE_CURL_UPLOAD) only in regions
+    # where the feature has been gated on (see is_feature_enabled / feature_support_matrix).
+    # Currently limited to eastus2euap for canary rollout.
+    if is_feature_enabled('enableCurlUpload'):
+        default_configs["ENABLE_CURL_UPLOAD"] = "true"
+
     """
     Decide the mode and configuration. There are two supported configuration schema, mix-and-match between schemas is disallowed:
         Legacy:          allows one of [MCS, GCS single tenant, or GCS multi tenant ("Auto-Config")] modes
@@ -2744,7 +2750,8 @@ def is_feature_enabled(feature):
     feature_support_matrix = {
         'useDynamicSSL'             : ['all'],
         'enableCMV2'                : ['all'],
-        'enableAzureOTelCollector'  : ['all']
+        'enableAzureOTelCollector'  : ['all'],
+        'enableCurlUpload'          : ['eastus2euap']
     }
     
     featurePreviewFlagPath = PreviewFeaturesDirectory + feature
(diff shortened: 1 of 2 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -909,6 +909,12 @@ def enable():
     ssl_cert_var_name, ssl_cert_var_value = get_ssl_cert_info('Enable')
     default_configs[ssl_cert_var_name] = ssl_cert_var_value
 
+    # Enable the libcurl-based ODS upload path (ENABLE_CURL_UPLOAD) only in regions
+    # where the feature has been gated on (see is_feature_enabled / feature_support_matrix).
+    # Currently limited to eastus2euap for canary rollout.
```
hestolz: consider adding centraluseuap

simathih: addressed

hestolz (approved):

AzureMonitorAgent/tests/test_agent.py
```diff
@@ -319,5 +338,97 @@ def test_port_reset_allows_regeneration(self, mock_exists):
...
+
+        def exists_side_effect(path):
+            return path == disabled_path
+
+        with patch('os.path.exists', side_effect=exists_side_effect):
+            with patch('agent.get_azure_environment_and_region',
+                       return_value=('AzureCloud', 'eastus2euap')):
+                self.assertFalse(agent.is_feature_enabled('enableCurlUpload'))
+
+    def test_unknown_feature_disabled(self):
```
thperapp, line 399: nit: shouldn't this be in a general feature flag test class rather than a ENABLE_CURL specific one?

AzureMonitorAgent/tests/test_agent.py
```diff
@@ -319,5 +338,97 @@ def test_port_reset_allows_regeneration(self, mock_exists):
         self.assertEqual(agent.MDSDSyslogPort, 0)
 
 
+class TestIsFeatureEnabledCurlUpload(unittest.TestCase):
+    """Tests for is_feature_enabled('enableCurlUpload') region gating (PR #2190)."""
+
+    def test_curl_upload_in_feature_support_matrix(self):
```
thperapp, line 344: nit: Isn't this more of an "azure_envionment" test rather than "is feature in support matrix"? latter is covered by the test case below

PR #2192, opened 2026-07-01
Add symlink and chown checks when accessing key path

norakoiralamsft:
The PR aims to address improper handling of symbolic links in the process of resetting or setting SSH public keys for existing Linux users. 

This was the suggested mitigation : Open the target with O_NOFOLLOW (and O_EXCL when creating), or verify with os.lstat/os.path.islink that the final path component is not a symlink immediately before writing. Replace os.chown with os.lchown so ownership changes never follow a symlink. Before operating, validate that the parent .ssh directory and the authorized_keys file are owned by the target user and are not symlinks; refuse otherwise. Apply the same guards to the SSH2/PKCS8 conversion branch (ssh_deploy_public_key).


VMAccess/vmaccess.py
```diff
@@ -43,6 +44,48 @@
 # overwrite the default logger
 logger.global_shared_context_logger = logger.Logger('/var/log/waagent.log', '/dev/stdout')
 
+
+def validate_ssh_path(pub_path, user_name):
+    p = pwd.getpwnam(user_name)
```
jscalev: Can this return a nil or err?

VMAccess/vmaccess.py
```diff
@@ -43,6 +44,48 @@
...
+    uid = p.pw_uid
+
+    ssh_dir = os.path.dirname(pub_path)
+
```
jscalev: Similarly, can this fail?

VMAccess/vmaccess.py
```diff
@@ -43,6 +44,48 @@
...
+    # Validate .ssh directory is not a symlink and is owned by the target user
+    if os.path.islink(ssh_dir):
+        raise Exception("Refusing to write: {0} is a symlink".format(ssh_dir))
+    if os.path.exists(ssh_dir):
+        st = os.lstat(ssh_dir)
+        if st.st_uid != uid and st.st_uid != 0:
+            raise Exception("Refusing to write: {0} is not owned by {1} or root".format(ssh_dir, user_name))
+
+    # Validate authorized_keys is not a symlink and is owned by the target user
+    if os.path.islink(pub_path):
```
jscalev: Nit: could create a helper method since the code above is the same

VMAccess/vmaccess.py
```diff
@@ -43,6 +44,48 @@
 # overwrite the default logger
 logger.global_shared_context_logger = logger.Logger('/var/log/waagent.log', '/dev/stdout')
 
+
```
jscalev: Overall comment. unit tests?

VMAccess/vmaccess.py
norakoiralamsft: yes, can return error, I added a catch in the caller to handle the exception

VMAccess/vmaccess.py
norakoiralamsft: won't fail, only return an empty string, added a check before passing the path into helper function

VMAccess/vmaccess.py
norakoiralamsft: added

VMAccess/vmaccess.py
norakoiralamsft: added

VMAccess/vmaccess.py
```diff
@@ -302,17 +305,36 @@ def _set_user_account_pub_key(protect_settings, hutil):
...
                     [constants.Openssl, 'x509', '-in', 'temp.crt', '-noout', '-pubkey'], "temp.pub")
                 if retcode > 0:
                     raise Exception("Failed to generate public key file.")
 
+                # Remove pub_path before ssh_deploy_public_key so it cannot be a
+                # pre-existing symlink that set_file_contents would follow.
+                if os.path.lexists(pub_path):
+                    if os.path.islink(pub_path):
+                        raise Exception("Refusing to write: {0} is a symlink".format(pub_path))
+                    os.remove(pub_path)
```
D1v38om83r: is pub_path the path to .ssh/authorized_keys? do we want to remove it and not append to it?

D1v38om83r: os.path.lexists will not return true on non-symbolic links?

D1v38om83r: validate_ssh_path will throw exception before we get here, if we have symbolic links in pub_path?

norakoiralamsft: 1) Opted to honor remove_prior_keys input from the user, so will append / rewrite as needed (no removal)
2) Removed that section, but yes lexists will return true on non-symblic links 
3) Yes it will, so removed redundant check

jscalev (approved):

PR #2194, opened 2026-07-08
AzureMonitorAgent - ME: Remove env var for OTLP ports

ngowran:
Remove env var for OTLP ports in ME service file 

Tested changes to service config in ubuntu 24 & RHEL 9.8


AzureMonitorAgent/services/metrics-extension-cmv2.service
```diff
@@ -4,7 +4,6 @@ After=network.target
 
 [Service]
 Environment="OTLP_GRPC_HOST=127.0.0.1"
-Environment="OTLP_GRPC_PORT=4317"
 Environment="OTLP_GRPC_PROM_HOST=127.0.0.1"
 Environment="OTLP_GRPC_PROM_PORT=4316"
```
ngowran: @sajithmuh can we remove the prom env vars? this will be "localhost:4316" by default without the config override

simathih (dismissed):

AzureMonitorAgent/services/metrics-extension-cmv2.service
sajithmuh: I'd keep this untouched until agent settings are made available by AMCS. There might be customers who are using custom values for overriding the ports and it will break for them.

Since we are releasing this for PMs to test, can't they manually override this and test it out?

If agent settings are available and documented in public, we can remove this and we can share the agent setting public docs to customers who want to override port settings

ngowran: When you say keep this untouched , do you mean the prom port or also the grpc one? As in leave all changes to this file once available? We can definitely manually override for testing

sajithmuh: I meant keep otlp port env var for otel and prom intact.

There are 2 ways we can do this:
1. Either we go ahead with the changes, inform CSS in advance that we are including a breaking change for a very small subset of customers and ask them to pin to AMA 1.43.0 version until agent settings is available.
2. Only apply these changes when agent settings are available in all clouds.

The first one has it's risk that we basically block on any updates for AMA as a whole (for that small subset of customers).
If there is no requirement for GA at this time, I'd keep it as it is.

Optionally, you can use the AMA feature flag to use another `metrics-extension` systemd unit file with the desired features and enable them only in canary so that PMs can test this easily in canary.

ngowran: Thanks very much for your input Sajith! I agree with you, lets go with option 2. Closing this pr for now.

GreyCat (approved):

PR #2197, opened 2026-07-15
AzureMonitorAgent: support Azure Linux 4

thperapp:
Adds Azure Linux 4 (x86_64 + aarch64) support:
- supported_distros.py: add '4' to the 'azurelinux' key in supported_dists_x86_64 and supported_dists_aarch64 so the extension no longer rejects Azure Linux 4 with 'Unsupported operating system'.
- agent.py install(): Azure Linux 4 ships journald-only with no rsyslog pre-installed, so AMA's Syslog collection produces no data until rsyslog is present. Mirror the existing Debian 12/13 and Amazon Linux 2023 handling by installing rsyslog via dnf when missing.

Validated end-to-end on a real Azure Linux 4.0 VM (AMA 1.42.0)

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -473,6 +473,15 @@ def install():
             if rsyslog_exit_code != 0:
                 return rsyslog_exit_code, rsyslog_output
     
+    # Check if Azure Linux 4+ VMs have rsyslog package (Azure Linux 4 ships journald-only, no rsyslog by default)
+    if (vm_dist.startswith('azurelinux')) and int(vm_ver.split('.')[0]) >= 4:
+        check_rsyslog, _ = run_command_and_log("rpm -q rsyslog")
+        if check_rsyslog != 0:
+            hutil_log_info("'rsyslog' package missing from Azure Linux {0} machine, installing to allow AMA to run.".format(vm_ver))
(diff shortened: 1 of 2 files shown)


AzureMonitorAgent/agent.py
```diff
@@ -473,6 +473,15 @@ def install():
             if rsyslog_exit_code != 0:
                 return rsyslog_exit_code, rsyslog_output
     
+    # Check if Azure Linux 4+ VMs have rsyslog package (Azure Linux 4 ships journald-only, no rsyslog by default)
+    if (vm_dist.startswith('azurelinux')) and int(vm_ver.split('.')[0]) >= 4:
+        check_rsyslog, _ = run_command_and_log("rpm -q rsyslog")
+        if check_rsyslog != 0:
+            hutil_log_info("'rsyslog' package missing from Azure Linux {0} machine, installing to allow AMA to run.".format(vm_ver))
+            rsyslog_exit_code, rsyslog_output = run_command_and_log("dnf install -y rsyslog")
```
hestolz, line 481: any need for `dnf update` / `dnf makecache` first

hestolz (approved):

AzureMonitorAgent/agent.py
thperapp: dnf should auto-refresh repo metadata as part of install when the cache is missing or expired

PR #2198, opened 2026-07-23
Fixing protected settings key leak

jaytripath:
Fixes a security incident where the decrypted protectedSettings value was leaked into extension logs.

**Root cause:**
On VMs where OpenSSL is misconfigured (missing /usr/local/ssl/openssl.cnf), openssl cms/smime prints a non-fatal warning to stderr. Because stderr was merged into the same stream as the decrypted stdout, the warning text got prepended to the decrypted JSON, breaking json.loads(). The resulting exception handler in HandlerUtil.py then logged the raw decrypted payload through redact_protected_settings() — but that helper only strips the outer, still-encrypted envelope fields (protectedSettings, protectedSettingsCertThumbprint); it doesn't recognize secrets inside the already-decrypted payload (e.g. workspaceKey), so the redaction was a silent no-op and the raw key was logged in plaintext.

Separately, agent.py's Install operation could crash outright (exit code 1) if logrotate.d doesn't exist on the VM, since the logrotate-config copy wasn't guarded against that failure.

**Changes:**

- HandlerUtil.py — _parse_config():

On JSON-decode failure of the decrypted protectedSettings, stop logging the (ineffectively "redacted") raw payload. Log only the exception message instead, which never contains the original content.

- agent.py — get_settings():

Changed the OpenSSL decrypt subprocess to capture stderr separately (subprocess.PIPE) instead of merging it into stdout (subprocess.STDOUT). Prevents non-fatal OpenSSL warnings from corrupting the JSON payload and causing decode failures in the first place — hardening the equivalent decrypt path used by Enable/metrics/syslog/transform-config operations.

- agent.py — parse_context():

Wrapped the logrotate config copyfile() in a try/except (mirroring the existing pattern for the sibling os.remove() call), so a missing logrotate.d directory logs a warning instead of failing the entire Install operation with exit code 1.

**Work-item:**
https://dev.azure.com/msazure/One/_workitems/edit/38590195

### DIFF
--- a/AzureMonitorAgent/agent.py
+++ b/AzureMonitorAgent/agent.py
@@ -2447,9 +2447,13 @@ def parse_context(operation):
                         output = 'Logrotate removal failed with error: {0}\nStacktrace: {1}'.format(ex, traceback.format_exc())
                         hutil_log_info(output)
             else:
-                if not os.path.exists(AMAExtensionLogRotateFilePath):      
-                    logrotateFilePath = os.path.join(os.getcwd(), 'azuremonitoragentextension.logrotate')
-                    copyfile(logrotateFilePath,AMAExtensionLogRotateFilePath)
+                if not os.path.exists(AMAExtensionLogRotateFilePath):
+                    try:
+                        logrotateFilePath = os.path.join(os.getcwd(), 'azuremonitoragentextension.logrotate')
+                        copyfile(logrotateFilePath,AMAExtensionLogRotateFilePath)
+                    except Exception as ex:
+                        output = 'Logrotate config copy failed with error: {0}\nStacktrace: {1}'.format(ex, traceback.format_exc())
+                        hutil_log_info(output)
             
         # parse_context may throw KeyError if necessary JSON key is not
         # present in settings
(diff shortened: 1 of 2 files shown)


jaytripath:
@microsoft-github-policy-service agree company="Microsoft"

AzureMonitorAgent/agent.py
```diff
@@ -2447,9 +2447,13 @@ def parse_context(operation):
                         output = 'Logrotate removal failed with error: {0}\nStacktrace: {1}'.format(ex, traceback.format_exc())
                         hutil_log_info(output)
             else:
-                if not os.path.exists(AMAExtensionLogRotateFilePath):      
-                    logrotateFilePath = os.path.join(os.getcwd(), 'azuremonitoragentextension.logrotate')
-                    copyfile(logrotateFilePath,AMAExtensionLogRotateFilePath)
+                if not os.path.exists(AMAExtensionLogRotateFilePath):
+                    try:
+                        logrotateFilePath = os.path.join(os.getcwd(), 'azuremonitoragentextension.logrotate')
```
simathih, line 2452: is this related to the security bug?

simathih (approved):

NarineM (approved):

AzureMonitorAgent/agent.py
jaytripath: No, this is not related to the security bug. This code change addresses a separate issue involving an install crash.

PR #2206, opened 2026-08-28
Fix metrics-sourcer crash loop by cleaning up orphaned telegraf unit

NarineM:


AzureMonitorAgent/agent.py
```diff
@@ -1671,9 +1671,16 @@ def uninstall_azureotelcollector():
...
+    # Whether the telegraf (metrics-sourcer) unit file exists.
+    try:
+        return os.path.isfile(telhandler.get_telegraf_service_path(is_lad=False))
+    except Exception:
+        return False
+
 def stop_metrics_process():
 
-    if telhandler.is_running(is_lad=False):
+    if telhandler.is_running(is_lad=False) or is_telegraf_service_installed():
```
simathih, line 1683: maybe check if its enabled as well right?

simathih (changes requested):

AzureMonitorAgent/agent.py
simathih: + @sajithmuh @ngowran @GreyCat as well to review

NarineM: metrics-sourcer is never [systemctl enabled (only restarted), so an enabled-check wouldn't gate anything — the crash loop comes from the leftover unit file + Restart=on-failure + the health-check restart, independent of enablement.

simathih (approved):

AzureMonitorAgent/agent.py
simathih: oops realized this was metrics-sourcer and not metrics-extension; sorry for the spam ME team :)

PR #2210, opened 2026-09-01
Fix MSRC Oracle Issue

ms-ritesh:
[MSRC Issue Link](https://msazure.visualstudio.com/One/_workitems/edit/39292584?src=WorkItemMention&src-action=artifact_link)
$ORACLE_HOME is attacker controller so attacker can put malicious su binary in that path and we could end up calling that fake su with root priviledges.

Find below logs of testing
[Logs.txt](https://github.com/user-attachments/files/31679477/Logs.txt)


VMBackup/main/workloadPatch/WorkloadPatch.py
```diff
@@ -765,6 +765,16 @@ def thread_for_sql(self,args):
     
     def getRole(self):
         return "master"
+
+    def getSuPath(self):
+        # Resolve su to an absolute path so a directory injected into PATH (e.g. via ORACLE_HOME) cannot supply a fake su binary run as root.
+        su_paths = ["/usr/bin/su", "/bin/su"]
```
ppusuluri-ms, line 772: do we need /bin/su as well?

VMBackup/main/workloadPatch/WorkloadPatch.py
```diff
@@ -765,6 +765,16 @@ def thread_for_sql(self,args):
     
     def getRole(self):
         return "master"
+
+    def getSuPath(self):
+        # Resolve su to an absolute path so a directory injected into PATH (e.g. via ORACLE_HOME) cannot supply a fake su binary run as root.
```
ppusuluri-ms, line 770: please add try catch to fallback to su in case of exception and add a comment saying fallback to be removed in future

VMBackup/main/workloadPatch/WorkloadPatch.py
ms-ritesh: There are some old linux distributions where su could be present at /bin/su thats why added this.

VMBackup/main/workloadPatch/WorkloadPatch.py
ms-ritesh: Exception catch is unnecessary here.
we are checking if su exists at those paths using os.path.exists(su_path) -- this method doesnt throw if su is absent, it simply returns false, thats why we have fallback on false value rather than exception.

ppusuluri-ms: not here, but while executing self.getSuPath(), are we sure that also does not throw any kind of exception?

ms-ritesh: as discussed offline, we don't have any breaking scenario for this usage as we have already verified presence of su in absolute path.

ppusuluri-ms (approved):

PR #2211, opened 2026-09-07
Fix bundled waagent path resolution

linugeorgeofficial:
**Summary**

• Resolve the bundled  WaagentLib.py  relative to  WAAgentUtil.py  instead of the process current working directory.
• Resolve  handle.sh  relative to  handle.py .
• Treat an unset  PYTHONPATH  as an empty search path instead of raising  KeyError .
• Add focused  unittest  coverage and document the VMBackup test workflow.

**Problem**

 WAAgentUtil.py  currently locates the bundled  main/WaagentLib.py  through  os.getcwd() . When the extension process starts from a different working directory, that lookup fails and VMBackup can fall back to the guest-installed waagent. Guest waagent versions are independently updated and may not be loadable as standalone modules; affected restore-point operations have failed with  name 'shellutil' is not defined .

The fallback searches also index  os.environ['PYTHONPATH']  directly. If that variable is unset, module initialization can instead fail with  KeyError: 'PYTHONPATH' .

**Scope**

This refreshes the production changes from #2160 on the current  master  branch. The system-waagent and  PYTHONPATH  fallback behavior is intentionally retained to keep this PR scoped to correcting path resolution and the unset-environment failure. Removing those fallbacks changes recovery behavior when the bundled file is unavailable and should be evaluated separately.

README.txt  is replaced with  README.md  to document the extension layout and dependency-free unit-test command. Packaging continues to exclude the existing  VMBackup/test  directory.


VMBackup/main/Utils/WAAgentUtil.py
```diff
@@ -29,10 +29,10 @@
 # it as a submodule of current module
 #
 def searchWAAgent():
-    agentPath = os.path.join(os.getcwd(), "main/WaagentLib.py")
+    agentPath = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir, "WaagentLib.py")
```
ms-ritesh, line 32: Do we know exactly in which scenario extension process's working directory was different?

VMBackup/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Run unit tests
+
+The unit tests use Python's standard `unittest` framework and require no additional packages. Use Python 3.6 or later for the test suite; this does not change the extension's runtime compatibility requirements.
+
+From the `VMBackup` directory, run:
+
+```sh
+python -m unittest discover -s test -p "test_*.py" -v
```
ms-ritesh, line 34: I think we should execute this in our build pipeline. 
Can you pls add this work item and we can plan it in upcoming sprints?

VMBackup/main/Utils/WAAgentUtil.py
linugeorgeofficial: I have not identified the exact production invocation that supplied a different working directory. The normal handler path appears to run from the versioned extension directory:  HandlerManifest.json  invokes  main/handle.sh , and  handle.sh  itself assumes the extension root as its working directory.
The available failure telemetry does not record the process CWD or selected waagent path. In a controlled VM test, running the existing lookup from  /tmp  consistently skipped the bundled  WaagentLib.py  and entered the guest-waagent fallback. The exact fallback exception depended on the installed WALinuxAgent version, so that test demonstrates the CWD-dependent failure mechanism but does not reproduce or identify the precise production launcher transition.
Since the bundled library has a fixed relationship to  WAAgentUtil.py , resolving it through  __file__  removes that dependency regardless of how the process was launched.

VMBackup/README.md
linugeorgeofficial: Created WI 
https://dev.azure.com/msazure/One/_workitems/edit/39645366

VMBackup/README.md
```diff
@@ -0,0 +1,37 @@
...
+|-- main/          Extension source and bundled runtime files
+|-- test/          Unit tests, excluded from the extension package
+|-- references     Shared repository dependencies used during packaging
+|-- setup.py       Extension packaging script
+`-- README.md
+```
+
+## Run unit tests
+
+The unit tests use Python's standard `unittest` framework and require no additional packages. Use Python 3.6 or later for the test suite; this does not change the extension's runtime compatibility requirements.
```
ppusuluri-ms: can we run unittests against all python versions that we support to verify compatibility?

VMBackup/README.md
ppusuluri-ms: The newly added unit tests are not being run in build pipeline? if yes, we should add them as part of this PR

VMBackup/main/Utils/WAAgentUtil.py
ppusuluri-ms, line 32: do we need to wrap these inside try catch and fallback to existing code in catch to not regress?

linugeorgeofficial: dirname,  join,  abspath, and  os.pardir  are standard-library operations available across the target python range and require no external commands.

VMBackup/README.md
linugeorgeofficial: Removed python 3 dependency and ran UT's with older python version

ms-ritesh (approved):

.github/workflows/vmbackup-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
```
ppusuluri-ms, line 24: should be tested with python 2.7 as well

.github/workflows/vmbackup-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Run VMBackup tests
+        working-directory: VMBackup
+        run: python -m unittest discover -s test -p "test_*.py" -v
```
ppusuluri-ms, line 48: where are we failing the build if tests are failed?

.github/workflows/vmbackup-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Run VMBackup tests
+        working-directory: VMBackup
```
ppusuluri-ms, line 35: Can you also please add code coverage checks? currently if it is at 10%, lets add 10% as gating and we will improve coverage with every PR

ppusuluri-ms (changes requested):

.github/workflows/vmbackup-tests.yml
linugeorgeofficial: The unit tests have already been validated locally using Python 2.7 and all three tests passed. The automated workflow currently covers Python 3.8 through Python 3.14.
Python 2.7 is not available through the current  ubuntu-latest  and  actions/setup-python  setup. Adding it to CI would therefore require a separate approved provisioning mechanism.

This follow-up is tracked in work item 39645366: https://dev.azure.com/msazure/One/_workitems/edit/39645366

.github/workflows/vmbackup-tests.yml
linugeorgeofficial: If a unit test fails,  python -m unittest  returns a non-zero exit code, causing the VMBackup Tests step, matrix job, and workflow check to fail.

D1v38om83r (commented):
Besides test coverage for Python 2.7 changes look good to me.

Shall approve when Python 2.7 is covered in the tests.

linugeorgeofficial:
@D1v38om83r , 
Python 2.7 cannot be added as another entry in the existing matrix because that matrix uses  ubuntu-latest  with  actions/setup-python . Python 2.7 has been removed from the action’s available Python builds and is not present on the Ubuntu 24.04 runner, so the job would fail during Python setup before running the tests.
The tests have been validated locally with Python 2.7.18. Adding ongoing automated coverage requires a separate approved Python 2.7 container or internal runner, which is tracked in work item 39714776: https://dev.azure.com/msazure/One/_workitems/edit/39714776

.github/workflows/vmbackup-tests.yml
linugeorgeofficial: Added automated Python 2.7.18 validation using a separate digest-pinned container job. It cannot use the existing  actions/setup-python  matrix because Python 2.7 is no longer available through that action.

ppusuluri-ms (approved):
changes look good to me

D1v38om83r (approved):

